Issue #1, opened 2015-10-08
`make lint` blows up on Python 3.5


Ichimonji10:
The `lint` make target calls `pylint`, which makes use of astroid. From my reading of the Travis log files, astroid does not yet work well on Python 3.5. We should skip this particular test for Python 3.5. For an example of how to conditionally execute a `script:` element in `.travis.yml`, see Tornado's `.travis.yml`: https://github.com/tornadoweb/tornado/blob/master/.travis.yml

Ichimonji10:
Pylint depends on a library called astroid. According to [this issue](https://bitbucket.org/logilab/pylint/issues/660/python35-problem-importing-module), astroid 1.3.8 is incompatible with Python 3.5, but astroid 1.4.0 is compatible. PyPI currently offers astroid 1.3.8. I suggest working around this issue and reverting the workaround when 1.4.0 is uploaded.

Ichimonji10:
I'm currently pushing branches to master, waiting for Travis to validate them, and then merging them unilaterally. I'm doing this because Pulp Smash is an extremely new library with numerous weaknesses, and as the primary author, I feel comfortable taking this license. However, this is *not* a good long term development strategy. When it seems appropriate, I will switch to forking the library, submitting pull requests and waiting for others to provide at least two ACKs and a merge.

Ichimonji10:
See: https://bitbucket.org/logilab/pylint/issues/660/python35-problem-importing-module#comment-22430431

PR #6, opened 2015-10-12
Fixed example of config for verify setting.

dkliban:
The capitalized False in the example is not proper JSON. I was getting an error until I changed it to a lower case.


Ichimonji10:
Ahh. That's my fault. Thanks for catching this.

ACK

Can one more person review this?

Ichimonji10:
Ehh. I'm actually going to merge this right away, as per https://github.com/PulpQE/pulp-smash/issues/1#issuecomment-146677147

dkliban:
Thank!

PR #8, opened 2015-10-12
Make Travis notify chat.freenode.net#pulp

Ichimonji10:
Given that the purpose of Pulp Smash is to test Pulp, it makes more sense for
Pulp contributors to see notifications than Robottelo contributors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ script:
 after_success:
     coveralls
 notifications:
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Travis checks pass. Merging as per https://github.com/PulpQE/pulp-smash/issues/1#issuecomment-146677147

PR #9, opened 2015-10-12
Update login test

Ichimonji10:
Update method `pulp_smash.tests.test_login.LoginFailureTestCase.test_body`. Formerly, the test asserted that the response body did not contain "key" and "certificate" keys, as documented [here](https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/authentication.html). Now, the test asserts that the response body has the keys documented [here](https://pulp.readthedocs.org/en/latest/dev-guide/conventions/exceptions.html#exception-handling), plus the "href" key as documented [here](PulpQE#9 (comment)).

Test results:

    $ python -m unittest2 pulp_smash.tests.test_login
    F...
    FAIL: test_body (pulp_smash.tests.test_login.LoginFailureTestCase)
    Assert that the response is valid JSON and does not have "key" and
      File "pulp_smash/tests/test_login.py", line 70, in test_body
        self.assertEqual(set(self.response.json().keys()), ERROR_KEYS)
    u'http_request_method'
    u'auth_error_code'
    u'href'

    Ran 4 tests in 1.319s

    FAILED (failures=1)

### DIFF
--- a/pulp_smash/tests/test_login.py
+++ b/pulp_smash/tests/test_login.py
@@ -14,6 +14,15 @@ from unittest2 import TestCase
 
 LOGIN_KEYS = {'certificate', 'key'}
 LOGIN_PATH = '/pulp/api/v2/actions/login/'
+ERROR_KEYS = {
+    '_href',
+    'error',
+    'error_message',
+    'exception',
+    'href',  # Present prior to Pulp 3.0 for backward compatibility.
+    'http_status',
+    'traceback',
+}
 
 
 class LoginSuccessTestCase(TestCase):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@dkliban

dkliban:
@Ichimonji10 This looks good, but we will not be able to remove 'href' until Pulp 3.0 (for backwards compatibility). However, we will be able to add '_href' to the response.  So I think the test should check that both 'href' and '_href' is present for now. I'll submit a PR to fix this in Pulp.

Ichimonji10:
Updated. Can you re-review, @dkliban?

dkliban:
This looks right now.

Ichimonji10:
Thank you!

PR #10, opened 2015-10-12
Ignore SSL verification warnings

### LINKED ISSUE #4: Ignore SSL verification warnings
Let's suppress SSL verification warnings. If a user sets `"verify": False` in their configuration file, then the responsibility for the fall-out of that decision lies with them. And if they do not suppress SSL verification, then a different warning is raised.

Here's what Pulp testing looks like right now:

```
(pulp-smash3)[ichimonji10@beech:pulp-smash]$ python -m unittest2 discover pulp_smash.tests
/home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
../home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
../home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
../home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
..
Ran 8 tests in 4.695s

OK
```

NailGun's solution may be adaptable to our situation:

```python
from warnings import simplefilter

# The urllib3 module (which requests uses) refuses to make insecure HTTPS
# connections. You can override this behaviour by passing `verify=False` to any
# of its methods. For example:
#
#     requests.get('https://insecure.example.com', verify=False)
#
# However, urllib3 will still raise an InsecureRequestWarning if this is done.
# This is problematic. If a user has explicitly stated that they want HTTPS
# connections to be made and we are pestering the user with warnings about that
# fact, then we are effectively training the user to ignore warnings. The
# statement below filters out that warning.
#
# The urllib3 module will still refuse to make HTTPS connections if
# `verify=True`. In other words, the use of a warning filter does not affect
# whether urllib3 enforces the security of HTTPS connections.
#
# Unfortunately, Python's warnings system is process-wide. Imagine an
# application which uses both NailGun and some other library, and imagine that
# the other library generates `InsecureRequestWarning`s. NailGun will filter
# out the warnings produced by that library, due to the process-wide nature of
# Python's warnings and filter system. The `warnings.catch_warnings` context
# manager is not a good solution to this problem, as it is thread-unsafe.
simplefilter(
    'ignore',
    requests.packages.urllib3.exceptions.InsecureRequestWarning,
)
```

Ichimonji10:
Fix #4:

> Let's suppress SSL verification warnings. If a user sets "verify": False in
> their configuration file, then the responsibility for the fall-out of that
> decision lies with them. And if they do not suppress SSL verification, then a
> different warning is raised.

Do this by using `warnings.simplefilter`. Example results:

    $ python -m unittest2 discover pulp_smash.tests
    ....F...
    FAIL: test_body (pulp_smash.tests.test_login.LoginFailureTestCase)
    Assert that the response is valid JSON and does not have "key" and
      File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/test_login.py", line 70, in test_body
        self.assertEqual(set(self.response.json().keys()), ERROR_KEYS)
    'auth_error_code'
    'http_request_method'
    'href'

    Ran 8 tests in 4.857s

    FAILED (failures=1)

The test failure is known about and should be present. See #9.

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -11,12 +11,10 @@ Why does Pulp Smash exist? What are its goals, and what does it *not* do?
 Why Pulp Smash?
 ---------------
 
-Pulp Smash exists to make testing Pulp easy.
+Pulp Smash exists to make testing Pulp easy. [1]_
 
 Scope and Limitations
 ---------------------
-
-This is a list of goals that 
 
 Portability
 ~~~~~~~~~~~
@@ -29,7 +27,7 @@ Pulp Smash should be usable in any environment that supports:
 * `GNU Make`_ (or a compatible clone),
 * and the `XDG Base Directory Specification`_.
 
-This level of portability [1]_ allows Pulp Smash to be accessible [2]_.
+This level of portability [2]_ allows Pulp Smash to be accessible [3]_.
 
 Provisioning
 ~~~~~~~~~~~~
@@ -42,7 +40,7 @@ Destructiveness
 
 Should Pulp Smash record all changes it makes to a remote system and revert them
 when testing is complete, or should systems be treated as throw-away? In other
-words, should the systems under test be treated like pets or cattle? [3]_ This
+words, should the systems under test be treated like pets or cattle? [4]_ This
 has yet to be decided.
 
 Contributing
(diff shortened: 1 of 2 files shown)


Ichimonji10:
It'd be good to have some feedback on this pull request and its associated issue, #4. I favor suppressing warnings about insecure connection requests for the reasons listed in the diff, and because littering stderr with warnings makes it hard to see how tests are progressing.

That said, these warnings are a good reminder that using invalid SSL configurations is a *bad practice*, and we would be better off if we addressed the underlying cause, not just this symptom.

Ichimonji10:
I've received no feedback for three days. I'm going to make a judgment call and merge this. If it's the wrong choice, we can easily revert it.

elyezer:
I commented on the issue. And I ACK the PR.

Will try to commit more on reviewing pulp-smash PRs. Sorry about that.

Ichimonji10:
@elyezer Ahhhh, sorry about that. You're right, you did comment on the issue, and I'm not doing you justice. Thanks for the feedback you gave!

elyezer:
@Ichimonji10 ping me at any time to review PRs here if needed.

PR #20, opened 2015-10-23
Adding api for repositories and tasks and few repository tests

peterlacko:


.gitignore
```diff
@@ -14,3 +14,7 @@ __pycache__/
 # Coverage information
 /.coverage
 /htmlcov/
+
+# Vi(m)'s swap files
+*.swp
+*.swo
```
Ichimonji10: These swapfiles are produced because an individual developer chose to use vim, not because the project necessitated the use of vim. Consequently, they are best ignored in the individual developer's Git configuration, not in the project-wide Git configuration.

You can do this like so:

```sh
cat >>~/.gitconfig <<EOF
[core]
    excludesfile = /home/peterlacko/.gitignore_global
EOF
echo '*.swp' > ~/.gitignore_global
```

pulp_smash/resources/__init__.py
```diff
@@ -0,0 +1,4 @@
+# coding=utf-8
+"""Modules for easy manipulation with pulp resources like users, repositories
+etc."""
```
Ichimonji10: We can word-smith this and make it fit on one line. How about:

    """Modules for manipulating pulp resources like users, repositories, etc."""

pulp_smash/resources/platform/__init__.py
```diff
@@ -0,0 +1,4 @@
+# coding=utf-8
+"""Modules for easy manipulation with pulp resources like users, repositories
+etc."""
```
Ichimonji10: See: https://github.com/PulpQE/pulp-smash/pull/20/files#r43006121

pulp_smash/resources/platform/api_v2/__init__.py
```diff
@@ -0,0 +1,4 @@
+# coding=utf-8
+"""Modules for easy manipulation with pulp resources like users, repositories
+etc."""
```
Ichimonji10: See: https://github.com/PulpQE/pulp-smash/pull/20/files#r43006121

pulp_smash/resources/__init__.py
Ichimonji10: The `pulp_smash.resources.platform` and `pulp_smash.resources.platform.api_v2` modules have identical docstrings. As a result, the purpose of each module is not clear. Other developers will be asking themselves questions like "where does my new helper code go in this hierarchy of modules," and I'd like it if they can answer such questions by reading the docstrings for these modules.

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+            json=my_delta,
+            **self.cfg.get_requests_kwargs()
+        )
+
+
+class Task(object):
+    """Handles tasks related operations. So far only waiting for given tasks
+    to immediate finish is implemented.
+    """
+    cfg = get_config()
```
Ichimonji10: In my experience, class attributes are deceptive. They seem simple when you start using them, but become a real pain down the road when more use cases for your class arise. Here's two specific issues:

* Let's say a user installs Pulp Smash for the very first time and executes `make docs-html`. As part of the documentation generation process, this module is imported and `get_config` is called. Since the user has just installed Pulp Smash, a config will not be found, and a `ConfigFileNotFoundError` is raised. :cry:
* Let's say we have two Pulp servers, and we want to monitor a task on each of them. It would be nice if we could instantiate two `Task` objects, give each a unique `cfg` object and call `wait_for_task()` on each. Using a class attribute makes this impossible.

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+            # task finished with error
+            if response.json()["state"] in TASK_ERROR_STATES:
+                raise Exception("Error occured while polling task: ",
+                                response.text)
+            # task finished properly
+            if response.json()["state"] in TASK_FINISHED_STATES:
+                break
+        # task probably timed out
+        else:
+            raise Exception("Timeout occured while waiting for task: ",
```
Ichimonji10: Never raise `Exception`. It's a Bad Thing™ to do. If a client wants to catch this exception, they will write something like this:

```python
try:
    Task.wait_for_task()
except Exception:
    pass # do something
```

This is problematic because Exception is the parent class of most exceptions. This code might catch completely unrelated exceptions.

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+        self.data_keys.update(kwargs)
+        self.last_response = requests.post(
+            self.cfg.base_url + CREATE_REPOSITORY_PATH,
+            json=self.data_keys,
+            **self.cfg.get_requests_kwargs()
+        )
+
+    def delete_repo(self):
+        """Delete repository from pulp server.
+        After calling this method, <repo>.last_response.raise_for_status()
```
Ichimonji10: This is syntactically valid, but its semantics are probably not what you intend. reStructuredText denotes a new paragraph with a blank line, and there is no blank line here, so this is interpreted as one giant paragraph. See [_A reStructuredText Primer_](http://docutils.sourceforge.net/docs/user/rst/quickstart.html).

Ichimonji10:
I see that this pull request has been submitted from the master branch. You may wish to learn about using Git branching and merging. Using branches allows you to work on multiple tasks at once. You can add some repository tests in one branch, task tests in another branch, experiment with some helpers in a third branch, and so on. Super useful. See: http://www.git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging

I'm also going to request that you squash the commits in this pull request before it is merged. The end result is that this pull request should have very few commits - probably one. This makes it easier to review the commit log and revert changes if necessary. If you want to learn how to do a squash, try [researching](https://duckduckgo.com/?q=git+squash+commit) the topic, and ask me if you need further help.

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+            time.sleep(frequency)
+            response = requests.get(
+                Task.cfg.base_url +
+                POLL_TASK_PATH.format(task["task_id"]),
+                **Task.cfg.get_requests_kwargs()
+            )
+            try:
+                response.raise_for_status()
+            except HTTPError:
+                break
```
Ichimonji10: This logic is below:

```python
            if response.json()["state"] in TASK_FINISHED_STATES:
                break
```

As a result, an HTTP error code looks the same as task success to a caller. How about just letting this exception propagate up instead of catching it?

pulp_smash/tests/platform/api_v2/simple_repo_test.py
```diff
@@ -0,0 +1,174 @@
+# coding=utf-8
+"""Test for basic repo creating functionality.
+"""
```
Ichimonji10: If a docstring fits on one line, do this:

    """Ask a foo for a bar."""

If a docstring has a summary line and a body, do this:

    """Ask a foo for a bar.

    Ensure that the bar has a biz and a baz.

    """

pulp_smash/tests/platform/api_v2/simple_repo_test.py
```diff
@@ -0,0 +1,174 @@
...
+
+class RepoCreateSuccessTestCase(TestCase):
+    """Tests for successfull repo creating functionality.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create repo on pulp server
+        """
+        cls.repo = Repository(id=cls.__name__)
```
Ichimonji10: What happens if I cancel this test before `tearDownClass` runs, or if the repository deletion fails? In that case, I am unable to run this test again, because a repository with an identical ID already exists. Using a random string for the repository ID lets me avoid this issue.

pulp_smash/tests/platform/api_v2/simple_repo_test.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_status_code(self):
+        """Test if Create repo returned 201."""
+        self.assertEqual(self.repo.last_response.status_code, 201)
+
+    def test_correct_id(self):
+        """Test if response contain correct repo id.
+        """
+        self.assertEqual(
+            self.repo.last_response.json()['id'],
+            self.__class__.__name__,
```
Ichimonji10: Please avoid using names like this. You can access an object's class with `type(self)` instead of `self.__class__`. And the `__name__` reference can be dropped in favor of randomly generating an ID, which has benefits as described here https://github.com/PulpQE/pulp-smash/pull/20/files#r43018157

pulp_smash/tests/platform/api_v2/simple_repo_test.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        self.assertEqual(self.repo.last_response.status_code, 200,
+                         self.repo.last_response.json())
+
+    def test_body(self):
+        """Test if repo has all set attributes: id, description and display_name.
+        """
+        self.assertTrue(
+            all(self.repo.last_response.json()[key] == self.__class__.__name__
+                for key in ['id', 'display_name', 'description']))
```
Ichimonji10: This works. But you'll get more useful assertion error messages if you first check that the correct keys are present before checking the value of those keys.

Imagine for a moment that this assertion fails with a `KeyError`, and exception states that the no "display_name" attribute exists. In this case, you know that the server didn't return a "display_name" key, but you're left wondering what other attributes the server returned. Did the server return a "description" key? An assertion like this tells you for sure:

python
self.assertLessEqual(
    {'id', 'display_name', 'description'},
    keys(self.repo.last_response.json())
)
```

You can then go ahead and check the value of each key.

File: pulp_smash/tests/platform/api_v2/simple_repo_test.py [hunk 13]
```diff
@@ -0,0 +1,174 @@
...
+        """Test if Create repo returned 201."""
+        self.assertEqual(self.repo.last_response.status_code, 201)
+
+    def test_correct_id(self):
+        """Test if response contain correct repo id.
+        """
+        self.assertEqual(
+            self.repo.last_response.json()['id'],
+            self.__class__.__name__,
+            set(self.repo.last_response.json()))
```
Ichimonji10: pep8 allows for quite a bit of flexibility in formatting, and there are no really great code formatters for Python as in other languages, like Go and C. As a result, many different styles have emerged.

Please use the following layout for all method and function calls unless the situation strongly calls for something different. I know it's nit-picky of me to ask for this. But this layout is formulaic enough to work nearly everywhere, and using a consistent style makes it possible for people to easily scan through code.

```python
# Yes.
self.assertEqual(
    self.repo.last_response.json()['id'],
    self.__class__.__name__,
    set(self.repo.last_response.json())
)
self.assertEqual(self.repo.last_response.json()['id'], self.__class__.__name__)

# No: closing parenthesis is not at same indentation as start of statement.
self.assertEqual(
    self.repo.last_response.json()['id'],
    self.__class__.__name__,
    set(self.repo.last_response.json()))
# No: hanging arguments are used.
self.assertEqual(self.repo.last_response.json()['id'],
                 self.__class__.__name__,
                 set(self.repo.last_response.json()))
# No: multiple arguments on same line
self.assertEqual(
    self.repo.last_response.json()['id'], self.__class__.__name__
)
```

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+
+import requests
+import time
+from requests.exceptions import HTTPError
+from pulp_smash.config import get_config
+
+# List of all paths here
+CREATE_REPOSITORY_PATH = "/pulp/api/v2/repositories/"
+REPOSITORY_PATH = "/pulp/api/v2/repositories/{}/"  # .format(<repo_id>)
+REPOSITORY_PATH = "/pulp/api/v2/repositories/{}/"  # .format(<repo_id>)
```
Ichimonji10: Duplicate code.

pulp_smash/resources/platform/api_v2/api.py
```diff
@@ -0,0 +1,183 @@
...
+
+from __future__ import unicode_literals
+
+import requests
+import time
+from requests.exceptions import HTTPError
+from pulp_smash.config import get_config
+
+# List of all paths here
+CREATE_REPOSITORY_PATH = "/pulp/api/v2/repositories/"
```
Ichimonji10: This path is also used for searching. How about making a single constant like so:

```python
REPOSITORY_PATH = '/pulp/api/v2/repositories/'
```

This path can then be re-used throughout. If you'd really like to have separate constants for repositories in general and specific repositories, how about this?

```python
REPOSITORY_PATH = '/pulp/api/v2/repositories/'
REPOSITORY_ID_PATH = REPOSITORY_PATH + '{}/'
```

Ichimonji10:
I appreciate what you've done with creating the `Repository` and `Task` helper classes, @peterlacko. However, I'd really like to hold off on adding that functionality to Pulp Smash. Adding more tests and then writing up helper classes will give us more use cases, which helps to ensure that whatever we *do* implement will be more likely to stick. (Also, my own experiences writing API bindings for Satellite 6 inform my judgment call here.)

In the meantime, feel free to add something simple like `pulp_smash/constants.py` and `pulp_smash/helpers.py`, where the two modules have objects like `REPOSITORY_PATH` and `poll_task()`, respectively. I expect that these two modules will change or be dropped entirely as time goes on. But they're simple things, and should be fairly easy to manipulate when the time comes.

pulp_smash/resources/__init__.py
peterlacko: Since it's very likely that big portion of code and it's organization will change I didn't pay much attention to docstrings in here, partly also because I wasn't sure about purpose of each module. I agree about docstrings, one-liner looks better.

pulp_smash/resources/platform/api_v2/api.py
peterlacko: I see, will change it.

pulp_smash/resources/platform/api_v2/api.py
peterlacko: I already rewrote that, I also didn't find it as a good solution, relates also to [previous comment about raising Exception](https://github.com/PulpQE/pulp-smash/pull/20#discussion-diff-43011416)

pulp_smash/tests/platform/api_v2/simple_repo_test.py
peterlacko: Got it.

pulp_smash/tests/platform/api_v2/simple_repo_test.py
peterlacko: I was also thinking about this case and thought about trying to remove repo in setUpClass(), but similar problems could appear, so random ID makes sense.

pulp_smash/tests/platform/api_v2/simple_repo_test.py
peterlacko: OK, I will try to follow it this way.

peterlacko:
@Ichimonji10: Thank you for thoughtful and detailed responses, I'm refining my code now.

Ad: Helper classes: I didn't write them with intention to stay there for long time/forever, but only as simple temporary solution which can be easily extended or replaced later when we will have more use cases for them. I think that for now they could help us to write tests, ie. create and delete repository by on simple method call.

Ad: Branching and committing: I will follow your recommendations ;-)

pulp_smash/tests/platform/api_v2/simple_repo_test.py
Ichimonji10: Thank you.

pulp_smash/resources/platform/api_v2/api.py
Ichimonji10: Sorry, I don't understand the wording here. What happened?

pulp_smash/tests/platform/api_v2/simple_repo_test.py
Ichimonji10: :+1:

pulp_smash/tests/platform/api_v2/simple_repo_test.py
Ichimonji10: Thank you.

Ichimonji10:
> Ad: Helper classes: I didn't write them with intention to stay there for long time/forever, but only as simple temporary solution which can be easily extended or replaced later when we will have more use cases for them. I think that for now they could help us to write tests, ie. create and delete repository by on simple method call.

Aye, I understand the intent behind writing them. But in my opinion, they don't quite qualify as "simple". To use or understand code that makes use of `Repository`, a user must answer questions like:

* When a `Repository` object is instantiated, what happens to the arguments passed in? Are they saved as instance attributes? (No. They're saved in the `data_keys` attribute.)
* Do all arguments correspond to values accepted by the server, or are some keywords reserved by the object itself? (All args are passed to the server.)
* Do the methods like `create_repo` and `delete_repo` mutate the object they're called on, or do they not? (They do mutate the object.)
* Do methods like `create_repo` and `delete_repo` return anything? (No.)
* Why is it that `__init__` and `create` accepts an arbitrary set of kwargs, but `update_repo` explicitly constrains what values can be passed in via its signature? (I don't know.)
* The methods include "repo" in their name. Does this mean that I have to learn a new set of method names for every single object? (Yes.) Does this mean each object is unique in other ways? (Yes.)

So please, do not add helper objects right now. A simple set of constants (and *maybe* functions) is enough.

Mind you, I'm expecting unit tests for any globally accessible code. Code has a nasty habit of sticking around longer than expected.

pulp_smash/resources/platform/api_v2/api.py
peterlacko: I meant that I changed it like this:
    
    while time.time() <= task_timeout:
        time.sleep(frequency)
        response = requests.get(
             self.cfg.base_url + POLL_TASK_PATH.format(task["task_id"]),
             **self.cfg.get_requests_kwargs()
         )
         response.raise_for_status()
         # task finished (with success or failure)
         if (response.json()["state"]
                 in TASK_ERROR_STATES | TASK_SUCCESS_STATES):
             return response

so we can we can check response manually for error later (maybe error was expected?) and also Exception raising was removed.

peterlacko:
> I'm expecting unit tests for any globally accessible code

What do you mean by it?

pulp_smash/resources/platform/api_v2/api.py
Ichimonji10: Ahhh, now I understand. Thanks for explaining. :smile:

Ichimonji10:
> What do you mean by it?

If a private function is added to a test module, no unit tests need to be added. For example, if you add a function called `_handle_response()` to `pulp_smash.tests.platform.api_v2.test_repository.py`, no unit test needs to be added for that function.

On the other hand, if a module like `pulp_smash.resources.platform.api_v2` is added, then unit tests should be added for all new code.

pulp_smash/tests/platform/api_v2/simple_repo_test.py
elyezer: I've reread pep8 a few days ago, and figured out that the recommended way for multiline docstring is like example below ( see https://www.python.org/dev/peps/pep-0008/#documentation-strings for more information):

```python
"""Return a foobang

Optional plotz says to frobnicate the bizbaz first.
"""
```

In other words, the closing `"""` should stay on its own line and don't need a blank line before it.

I'm commenting because I am using the PEP8 recommend format. But as the PEP8 document states, it is a code style guide not a rule.

Ichimonji10: @elyezer Thanks for the heads up. See #24

omaciel:
Talked to @peterlacko and he agreed that we should close this PR. He will submit another soon.

Ichimonji10:
Thank you.

FYI, #28 was just merged, and https://github.com/Ichimonji10/pulp-smash/tree/zoo_repo_sync is in the works.

PR #26, opened 2015-10-30
Add tests for Pulp's searching facilities

### LINKED ISSUE #19: Expand GET/POST search tests
Pulp supports searching via both GET and POST requests. When searching with a GET request, parameters should be passed in via the URL, and when searching with a POST request, parameters should be passed in the body of the request. (This is appropriate behaviour, by the way. Network intermediaries are at liberty to discard the body of any GET request.) Unfortunately, this makes Pulp's search facilities more complicated and, therefore, more prone to bugs.

Tests should be written that execute identical GET and POST searches and verify that the results are identical. Much of the searching code is very generic and applies no matter what kind of entity is being searched for. As a result, even adding a few tests should help immensely with preventing regressions.

Further documentation on the GET and POST search queries is available [here](https://pulp.readthedocs.org/en/latest/dev-guide/conventions/criteria.html#search-api).

Ichimonji10:
Add a new module, `pulp_smash.tests.platform.api_v2.test_search`:

> """Test Pulp's `Searching`_ facilities.
>
> The tests in this module make use of the `User APIs`_. However, few
> user-specific references are made. These tests could be rewritten to use
> repositories or something else with only minimal changes. Thus, the name of
> this module.
>
> Each test case executes one or more pair of semantically identical POST and GET
> requests. Each pair of search results should match exactly.

Beware that the tests are slightly doctored to achieve these results. When
searching for users, search results do not contain an "id" attribute. The
affected tests work around this issue by referencing `['_id']['$oid']` instead.

Test results:

    $ python -m unittest2 pulp_smash.tests.platform.api_v2.test_search
    ssss............
    Ran 16 tests in 16.521s

    OK (skipped=4)

### DIFF
--- a/pulp_smash/tests/platform/api_v2/test_search.py
+++ b/pulp_smash/tests/platform/api_v2/test_search.py
@@ -1,0 +1,338 @@
+# coding=utf-8
+"""Test Pulp's `Searching`_ facilities.
+
+The tests in this module make use of the `User APIs`_. However, few
+user-specific references are made. These tests could be rewritten to use
+repositories or something else with only minimal changes. Thus, the name of
+this module.
+
+Each test case executes one or more pairs of semantically identical POST and
+GET requests. Each pair of search results should match exactly.
+
+Most test cases assume that the assertions in some other test case hold true.
+The assumptions explored in this module have the following dependencies::
+
+    It is possible to ask for all resources of a kind.
+    ├── It is possible to sort search results.
+    ├── It is possible to ask for a single field in search results.
+    ├── It is possible to ask for several fields in search results.
+    └── It is possible to ask for a resource with a specific ID.
+        └── It is possible to ask for a resource with one of several IDs.
+            ├── It is possible to skip some search results.
+            └── It is possible to limit how many search results are returned.
+
+.. _Searching:
+    https://pulp.readthedocs.org/en/latest/dev-guide/conventions/criteria.html
+.. _User APIs:
+    https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/user/index.html
+
+"""
+from __future__ import unicode_literals
+
+import requests
+from pulp_smash.config import get_config
+from pulp_smash.constants import USER_PATH
+from random import randint
+from unittest2 import TestCase, skip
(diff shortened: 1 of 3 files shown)


Ichimonji10:
@asmacdo

Ichimonji10:
@asmacdo Can you re-review? Have I commented out GET searches too aggressively?

pulp_smash/tests/platform/api_v2/test_search.py
```diff
@@ -0,0 +1,336 @@
...
+
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create one user. Execute searches."""
+        super(FieldsTestCase, cls).setUpClass()
+        cls.attrs_iter = (_create_user(cls.cfg),)
+        cls.responses = {
+            'get': _search_get(cls.cfg, '?field=login&field=roles'),
```
asmacdo: fyi this will fail on master.  https://pulp.plan.io/issues/1332

asmacdo:
LGTM. I have no problem with  searches for now since they only work in master. With what you have here, it will be very easy to add them when you are testing against pulp >= 2.7.1.

pulp_smash/tests/platform/api_v2/test_search.py
```diff
@@ -0,0 +1,336 @@
...
+    def test_field(self):
+        """Only the requested key should be in each response."""
+        for action, response in self.responses.items():
+            with self.subTest(action=action):
+                response.raise_for_status()
+                for attrs in response.json():
+                    self.assertEqual(set(attrs.keys()), {'name'})
+
+
+class FieldsTestCase(_BaseTestCase):
```
asmacdo: fyi these will fail on master. https://pulp.plan.io/issues/1332

Ichimonji10: Thanks for the link to the issue at hand. That's exactly what I need!

Ichimonji10: See: https://github.com/PulpQE/pulp-smash/issues/30#issuecomment-155201406

PR #27, opened 2015-11-03
[DO NOT MERGE] Adds test of syncing zoo repository.

dkliban:


pulp_smash/tests/rpm/api_v2/test_repo_create.py
```diff
@@ -0,0 +1,81 @@
...
+
+_RPM_REPO_URL = 'http://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/'
+
+
+def _rand_str():
+    """Return a randomized string."""
+    return type('')(randint(-999999, 999999))
+
+
+class CreateFeedTestCase(TestCase):
```
omaciel: Would we be adding similar tests for different importers? In other words, should he name of this module OR the name of the class state this this test is valid for a yum importer only? This way, one could then add another module/class for puppet, docker, etc.

dkliban: We definitely want to do similar tests for both yum_importer and puppet_importer. 

In both cases we want to have tests that are using a feed. The sync task needs to be monitored via API and assertions need to be made about the task status and error field. These tests need to assert that a publish task was also spawned after the sync finishes. The publish task should also be monitored and the same assertions should be made about its status and errors. Then a few pieces of the published content should be retrieved via HTTP request from the published repository to make sure that it is actually published. 

We also need tests that check how uploading functionality works for RPMs and Puppet modules. This workflow is documented here [0]. All upload tests should include an API call to publish the newly uploaded content. Then the content should be retrieved via HTTP request to assert that it got published. 

[0] http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/content/upload.html

Ichimonji10:
See #39.

PR #28, opened 2015-11-05
Add basic tests for repository CRUD.

jeremycline:
This commit also introduces a ``utils`` module which currently only
provides a function to generate random strings using ``uuid.uuid4``.

### DIFF
--- a/pulp_smash/tests/platform/api_v2/test_repository.py
+++ b/pulp_smash/tests/platform/api_v2/test_repository.py
@@ -1,0 +1,263 @@
+# coding=utf-8
+"""Test the `repository`_ API endpoints.
+
+The assumptions explored in this module have the following dependencies::
+
+        It is possible to create a repository.
+        ├── It is impossible to create a repository with a duplicate ID
+        │   or other invalid attributes.
+        ├── It is possible to read a repository.
+        ├── It is possible to update a repository.
+        └── It is possible to delete a repository.
+
+.. _repository:
+    https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/index.html
+
+"""
+from __future__ import unicode_literals
+
+import requests
+
+from pulp_smash.config import get_config
+from pulp_smash.constants import REPOSITORY_PATH, ERROR_KEYS
+from pulp_smash.utils import rand_str
+from unittest2 import TestCase
+
+# pylint:disable=duplicate-code
+# Once https://github.com/PulpQE/pulp-smash/pull/28#discussion_r44172668
+# is resolved, pylint can be re-enabled.
+
+
+class CreateSuccessTestCase(TestCase):
+    """Establish that we can create repositories."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create several repositories.
+
+        Create one repository with the minimum required attributes, and a
+        second with all available attributes except importers and distributors.
+
+        """
+        cls.cfg = get_config()
+        cls.url = cls.cfg.base_url + REPOSITORY_PATH
+        cls.bodies = (
+            {'id': rand_str()},
+            {
+                'id': rand_str(),
+                'display_name': rand_str(),
+                'description': rand_str(),
+                'notes': {rand_str(): rand_str()},
+            },
+
+        )
+        cls.responses = tuple((
+            requests.post(
+                cls.url,
+                json=body,
+                **cls.cfg.get_requests_kwargs()
+            )
+            for body in cls.bodies
+        ))
+
+    def test_status_code(self):
+        """Assert that each response has a HTTP 201 status code."""
+        for i, response in enumerate(self.responses):
+            with self.subTest(self.bodies[i]):
+                self.assertEqual(response.status_code, 201)
+
+    def test_location_header(self):
+        """Assert that the Location header is correctly set in the response."""
+        for i, response in enumerate(self.responses):
+            with self.subTest(self.bodies[i]):
+                self.assertEqual(
+                    self.url + self.bodies[i]['id'] + '/',
+                    response.headers['Location']
+                )
+
+    def test_attributes(self):
+        """Assert that each repository has the requested attributes."""
+        for i, body in enumerate(self.bodies):
+            with self.subTest(body):
+                attributes = self.responses[i].json()
+                self.assertLessEqual(set(body.keys()), set(attributes.keys()))
+                attributes = {key: attributes[key] for key in body.keys()}
+                self.assertEqual(body, attributes)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created repositories."""
+        for response in cls.responses:
+            requests.delete(
+                cls.cfg.base_url + response.json()['_href'],
+                **cls.cfg.get_requests_kwargs()
+            ).raise_for_status()
+
+
+class CreateFailureTestCase(TestCase):
+    """Establish that repositories are not created in documented scenarios."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create several repositories.
+
+        Each repository is created to test a different failure scenario. The
+        first repository is created in order to test duplicate ids.
+
+        """
+        cls.cfg = get_config()
+        cls.url = cls.cfg.base_url + REPOSITORY_PATH
+        identical_id = rand_str()
+        cls.bodies = (
+            (201, {'id': identical_id}),
+            (400, {'id': None}),
+            (400, ['Incorrect data type']),
+            (400, {'missing_required_keys': 'id'}),
+            (409, {'id': identical_id}),
+        )
+        cls.responses = tuple((
+            requests.post(
+                cls.url,
+                json=body[1],
+                **cls.cfg.get_requests_kwargs()
+            )
+            for body in cls.bodies
+        ))
+
+    def test_status_code(self):
+        """Assert that each response has the expected HTTP status code."""
+        for i, response in enumerate(self.responses):
+            with self.subTest(self.bodies[i]):
+                self.assertEqual(response.status_code, self.bodies[i][0])
+
+    def test_location_header(self):
+        """Assert that the Location header is correctly set in the response."""
+        for i, response in enumerate(self.responses):
+            with self.subTest(self.bodies[i]):
+                if self.bodies[i][0] == 201:
+                    self.assertEqual(
+                        self.url + self.bodies[i][1]['id'] + '/',
+                        response.headers['Location']
+                    )
+                else:
+                    self.assertNotIn('Location', response.headers)
+
+    def test_exception_keys_json(self):
+        """Assert the JSON body returned contains the correct keys."""
+        for i, response in enumerate(self.responses):
+            if self.bodies[i][0] >= 400:
+                response_body = response.json()
+                with self.subTest(self.bodies[i]):
+                    for error_key in ERROR_KEYS:
+                        with self.subTest(error_key):
+                            self.assertIn(error_key, response_body)
+
+    def test_exception_json_http_status(self):
+        """Assert the JSON body returned contains the correct HTTP code."""
+        for i, response in enumerate(self.responses):
+            if self.bodies[i][0] >= 400:
+                with self.subTest(self.bodies[i]):
+                    json_status = response.json()['http_status']
+                    self.assertEqual(json_status, self.bodies[i][0])
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created repositories."""
+        for response in cls.responses:
+            if response.status_code == 201:
+                requests.delete(
+                    cls.cfg.base_url + response.json()['_href'],
+                    **cls.cfg.get_requests_kwargs()
+                ).raise_for_status()
+
+
+class ReadUpdateDeleteSuccessTestCase(TestCase):
+    """Establish that we can read, update, and delete repositories.
+
+    This test assumes that the assertions in :class:`CreateSuccessTestCase` are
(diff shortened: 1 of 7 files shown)


pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,226 @@
+# coding=utf-8
+"""Test the `repository`_ API endpoints."""
```
Ichimonji10: No target is given for the "repository" [reference](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#reference-names). As a result, it should be broken. You can provide a target like so:

    .. repository: https://example.com

Of course, you don't actually want to link the user to example.com.

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,226 @@
...
+        ))
+
+    def test_status_code(self):
+        """Assert that each response has a HTTP 201 status code."""
+        for i, response in enumerate(self.responses):
+            with self.subTest(self.bodies[i]):
+                self.assertEqual(response.status_code, 201)
+
+    def test_location_header(self):
+        """Assert that the Location header is correctly set in the response."""
```
Ichimonji10: I didn't know that this is possible. :+1:

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,226 @@
...
+
+class CreateFailureTestCase(TestCase):
+    """Establish that repositories are not created in documented scenarios."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create several repositories.
+
+        Create one repository with the minimum required attributes, and a
+        second with all available attributes except importers and distributors.
```
Ichimonji10: This docstring appears incorrect.

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,226 @@
...
+        attributes = self.read_response.json()
+        self.assertLessEqual(
+            set(self.bodies[0].keys()),
+            set(attributes.keys())
+        )
+        attributes = {key: attributes[key] for key in self.bodies[0].keys()}
+        self.assertEqual(self.bodies[0], attributes)
+
+    def test_update_attributes(self):
+        """Assert that the repository has updated attributes."""
```
Ichimonji10: Try splitting this test method into multiple methods. For example:

```python
def test_update_attributes_v1(self):
    """Assert that the update request did not spawn any tasks."""
    # Assert that `spawned_tasks` is present and sane.

def test_update_attributes_v2(self):
    """Assert that the updated repository has correct attributes."""
    # Assert that `result` is present and sane.
```

It's generally undesirable to have more than one assertion in a test case, as an assertion failure prevents you from running following assertions. It only really makes sense to place multiple assertions in a test case when one assertion would always trigger a failure in a second test case (in which case it's worth trying to make multiple test cases).

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,226 @@
...
+                    self.assertEqual(
+                        self.url + self.bodies[i][1]['id'] + '/',
+                        response.headers['Location']
+                    )
+                else:
+                    self.assertNotIn('Location', response.headers)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created repositories."""
```
Ichimonji10: Each test case has identical `tearDownClass` methods, except for this test case. I'd suggest defining a parent class or mixin class like so:

```python
def _BaseTestCase(TestCase):
    @classmethod
    def tearDownClass(cls):
        for response in cls.responses:
            requests.delete(...).raise_for_status()
```

And you can make use of it here like so:

```python
@classmethod
def tearDownClass(cls):
    cls.responses = (cls.responses[0],)
```

All that said, working with response objects directly like this is a bit of a kludge, as is the use of numeric indices. So you may wish to leave things as is and later refactor things as per the example in #26.

pulp_smash/utils.py
```diff
@@ -0,0 +1,9 @@
+# coding=utf-8
+"""Utility functions for Pulp tests."""
+from __future__ import unicode_literals
+from uuid import uuid4
```
Ichimonji10: Nitpick: add a blank line above this.

pulp_smash/utils.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def rand_str():
```
Ichimonji10: Despite the definition of this new function, the coverage statistics don't change. That's because the `make test-coverage` make target needs to be tweaked. That said, this module is enormously simple, so I'm not terribly worried about fixing this right away. We can leave this as is, but if we do so, a GitHub issue should be created to fix this issue.

Ichimonji10:
Can you run the new tests and paste the output here when done?

pulp_smash/tests/platform/api_v2/test_repository.py
jeremycline: I'll leave it to you to refactor things, if that's alright.

jeremycline:
Output is:
```
(pulp-smash)[vagrant@dev pulp-smash]$ python -m unittest2 discover pulp_smash
....F.........................
FAIL: test_body (tests.platform.api_v2.test_login.LoginFailureTestCase)
Assert that the response is valid JSON and has correct keys.
  File "/home/vagrant/devel/pulp-smash/pulp_smash/tests/platform/api_v2/test_login.py", line 69, in test_body
    _LOGIN_ERROR_KEYS,
u'http_request_method'
u'auth_error_code'
u'href'

FAIL: test_status_code (tests.platform.api_v2.test_repository.CreateFailureTestCase) [(400, [u'Incorrect data type'])]
Assert that each response has the expected HTTP status code.
  File "/home/vagrant/devel/pulp-smash/pulp_smash/tests/platform/api_v2/test_repository.py", line 118, in test_status_code
    self.assertEqual(response.status_code, self.bodies[i][0])
AssertionError: 500 != 400

FAIL: test_use_deleted_user (tests.platform.api_v2.test_user.ReadUpdateDeleteTestCase) [put]
Assert that one cannot read, update or delete a deleted user.
  File "/home/vagrant/devel/pulp-smash/pulp_smash/tests/platform/api_v2/test_user.py", line 169, in test_use_deleted_user
    self.assertEqual(response.status_code, 404)
AssertionError: 400 != 404

Ran 32 tests in 2.966s

FAILED (failures=3)
```
I'm not sure what the problem is with the user tests, but the repository test should be failing (https://pulp.plan.io/issues/1356).

pulp_smash/tests/platform/api_v2/test_repository.py
Ichimonji10: Works for me.

asmacdo:
Line 169 is this failure: https://github.com/PulpQE/pulp-smash/issues/34

Ichimonji10:
I believe the failing user test is an expected failure. There's a known bug tested by that API call, and... I don't have a link to it.

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        cls.cfg = get_config()
+        cls.url = cls.cfg.base_url + REPOSITORY_PATH
+        identical_id = rand_str()
+        cls.bodies = (
+            (201, {'id': identical_id}),
+            (400, {'id': None}),
+            (400, ['Incorrect data type']),
+            (400, {'missing_required_keys': 'id'}),
+            (409, {'id': identical_id}),
```
Ichimonji10: As per [Exception Handling](https://pulp.readthedocs.org/en/latest/dev-guide/conventions/exceptions.html), these four API calls should return a call report with [`ERROR_KEYS`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.constants.html#pulp_smash.constants.ERROR_KEYS). Can we add in a test method for that?

Ichimonji10:
The two new `.rst` files which have been added to the documentation are not accessible. Can you link to them from [`docs/api.rst`](https://raw.githubusercontent.com/PulpQE/pulp-smash/master/docs/api.rst)?

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,235 @@
...
+        """Assert that the read repository has the correct attributes."""
+        attributes = self.read_response.json()
+        self.assertLessEqual(
+            set(self.bodies[0].keys()),
+            set(attributes.keys())
+        )
+        attributes = {key: attributes[key] for key in self.bodies[0].keys()}
+        self.assertEqual(self.bodies[0], attributes)
+
+    def test_update_attributes_spawned_tasks(self):
```
Ichimonji10: Can you add `  # pylint:disable=invalid-name` to the end of this line?

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,235 @@
+# coding=utf-8
+"""Test the `repository`_ API endpoints.
+
+.. _repository:
+    https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/index.html
+
+"""
```
Ichimonji10: Can we update this docstring to state the assumptions explored by this module? For example:

```
"""Test the `repository`_ API endpoints.

The assumptions explored in this module have the following dependencies::

    It is possible to create a repository.
    ├── It is impossible to create a repository with a duplicate ID or other
    │   invalid attributes.
    ├── It is possible to read a repository.
    ├── It is possible to update a repository.
    └── It is possible to delete a repository.

.. _repository:
    https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/index.html

"""
```

Ichimonji10:
As per https://github.com/PulpQE/pulp-smash/pull/28#discussion_r44172668, I'm going to refactor this module at a later point. Can you add `# pylint:disable=duplicate-code` to the top of the module of tests, with a comment explaining what's going on?

Ichimonji10:
I think that's about it. If you can address the points above, I think this is great.

pulp_smash/tests/platform/api_v2/test_repository.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def test_update_attributes_result(self):
+        """Assert that `result` is present and has the correct attributes."""
+        response = self.update_response.json()
+        self.assertIn('result', response)
+        for key in self.update_body['delta']:
+            with self.subTest(key):
+                self.assertIn(key, response['result'])
+                self.assertEqual(
+                    self.update_body[key],
```
asmacdo:
``` python
self.update_body['delta'][key]
```

Ichimonji10: Nice catch. This could also be written as:

```python
for key, value in self.update_body['delta'].items():
    with self.subTest(key=key):
        self.assertIn(key, response['result'])
        self.assertEqual(response['result'][key], value)
```

Ichimonji10:
@jeremycline Please rebase and apply this patch:

```diff
diff --git a/Makefile b/Makefile
index 22b7274..1dada96 100644
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,9 @@ docs-clean:
 
 lint:
 	flake8 .
-	pylint --reports=n --disable=I docs/conf.py pulp_smash tests setup.py
+	pylint --reports=n --disable=I docs/conf.py tests setup.py \
+		pulp_smash/{__init__,__main__,config,constants,utils}.py
+	pylint --reports=n --disable=I,duplicate-code pulp_smash/tests/
 
 test:
 	python $(TEST_OPTIONS)
```

The above patch causes pylint to ignore any duplicate code in the `pulp_smash.tests` module. I dislike just ignoring this warning - it's there for a reason, and the proper solution is to use composition or some other mechanism to pull the commonalities out. But I'm willing to take on that technical debt for now.

jeremycline:
@Ichimonji10 I have made it so. Note that I had to add `SHELL=/bin/bash` in the Makefile to support the bash path expansion in that patch.

Ichimonji10:
@jeremycline Will that cause the makefile to break on systems that may not have bash installed, such as Mac OS (or Cygwin, or other more exotic *nix-like distros)?

jeremycline:
@Ichimonji10 Yes. Does ``/usr/bin/env sh`` work in OSX/*nix?

Ichimonji10:
I don't know. We can just expand those braces:

```diff
diff --git a/Makefile b/Makefile
index 22b7274..1dada96 100644
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,9 @@ docs-clean:
 
 lint:
 	flake8 .
-	pylint --reports=n --disable=I docs/conf.py pulp_smash tests setup.py
+	pylint --reports=n --disable=I docs/conf.py tests setup.py \
+		pulp_smash/__init__.py \
+		pulp_smash/__main__.py \
+		pulp_smash/config.py \
+		pulp_smash/constants.py \
+		pulp_smash/utils.py
+	pylint --reports=n --disable=I,duplicate-code pulp_smash/tests/
 
 test:
 	python $(TEST_OPTIONS)
```

jeremycline:
I guess we *could* just do that.

Ichimonji10:
Ugly, but highly compatible.

jeremycline:
Very true. It is done.

Ichimonji10:
lady-gaga

PR #39, opened 2015-11-13
Add RPM Repository tests

Ichimonji10:
Add RPM repository tests. Module `pulp_smash.tests.rpm.api_v2.test_repo_create`
includes the following test cases:

* `CreateTestCase` ensures it is possible to create repositories with and
  without feeds.
* `SyncValidFeedTestCase` and `SyncInvalidFeedTestCase` ensure that Pulp
  correctly handles sync requests, regardless of whether the feed is valid.
* `PublishTestCase` ensures it is possible to upload an RPM file to a
  repository, copy the repository's contents to a second repository, add a
  distributor to the first repository, publish the first repository, and
  download the original RPM.

Test results:

    $ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_sync_publish
    .............s........
    Ran 22 tests in 27.118s

    OK (skipped=1)

Many thanks to @dkliban for help with designing these tests.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -1,0 +1,650 @@
+# coding=utf-8
+"""Test the API endpoints for RPM `repositories`_.
+
+This module assumes that the tests in
+:mod:`pulp_smash.tests.platform.api_v2.test_repository` hold true. The
+following trees of assumptions are explored in this module::
+
+    It is possible to create an RPM repo with a feed. (CreateTestCase)
+    ├── If a valid feed is given, the sync completes and no errors are
+    │   reported. (SyncValidFeedTestCase)
+    └── If an invalid feed is given, the sync completes and errors are
+        reported. (SyncInvalidFeedTestCase)
+
+    It is possible to create an RPM repository without a feed. (CreateTestCase)
+    └── It is possible to upload an RPM file to a repository, copy the
+        repository's contents to a second repository, add a distributor to the
+        first repository, publish the first repository, and download the
+        original RPM. (PublishTestCase)
+
+Assertions not explored in this module include:
+
+* Given an RPM repository without a feed, sync requests fail.
+* It is impossible to create two RPM repositories with the same relative URL.
+* It is possible to upload a directory of RPM files to an RPM repository.
+* It is possible to upload an ISO of RPM files to an RPM repository.
+* It is possible to upload content and copy it into multiple repositories.
+* It is possible to get content into a repository via a sync and publish it.
+
+.. _repositories:
+   http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/cud.html
+
+"""
+from __future__ import unicode_literals
+
+import os
+import requests
+from itertools import chain
+from pulp_smash.config import get_config
+from pulp_smash.constants import REPOSITORY_PATH
+from pulp_smash.utils import rand_str
+from time import sleep
+from unittest2 import TestCase
+
+
+from sys import version_info
+if version_info.major == 2:
+    from urlparse import urlparse  # noqa pylint:disable=import-error,no-name-in-module
+else:
+    from urllib.parse import urlparse  # noqa pylint:disable=import-error,no-name-in-module
+
+
+_VALID_FEED = 'http://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/'
+_INVALID_FEED = 'https://example.com'
+_TASK_END_STATES = ('canceled', 'error', 'finished', 'skipped', 'timed out')
+_RPM_URL = (
+    'https://repos.fedorapeople.org'
+    '/pulp/pulp/demo_repos/zoo/bear-4.1-1.noarch.rpm'
+)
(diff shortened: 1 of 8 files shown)


Ichimonji10:
This PR isn't quite ready for a merge yet. But I'd like to get some early review from @dkliban, or anyone else with expertise with RPM repositories. I've reworked the exact form of these tests pretty heavily from when they were first prototyped, and it'd be good to know if I broke anything. Also, the `test_error_details` test case is failing, and I think there was a bug for that. Can anyone point me to the right bug in our bug tracker?

pulp_smash/tests/rpm/api_v2/test_repo_create.py
```diff
@@ -0,0 +1,662 @@
...
+                self.assertEqual(body['importer_' + key], importers[0][key])
+
+    @classmethod
+    def tearDownClass(cls):
+        """Destroy the created RPM repositories."""
+        for attrs in cls.attrs_iter:
+            _delete(cls.cfg, attrs['_href'])
+
+
+class SyncFeedTestCase(TestCase):
```
dkliban: Should this be SyncInvalidFeedTestCase?

pulp_smash/tests/rpm/api_v2/test_repo_create.py
```diff
@@ -0,0 +1,662 @@
...
+        self.assertIn('spawned_tasks', self.response.json())
+
+    def test_spawned_tasks_v2(self):
+        """The server should return exactly one spawned task."""
+        task_bodies = tuple(chain.from_iterable(
+            _poll_tasks(self.cfg, spawned_task['_href'])
+            for spawned_task in self.response.json()['spawned_tasks']
+        ))
+        self.assertEqual(len(task_bodies), 1, task_bodies)
+
```
dkliban: There should probably be a test to check that an error was reported.

pulp_smash/tests/rpm/api_v2/test_repo_create.py
Ichimonji10: The `SyncFeedTestCase` and `SyncInvalidFeedTestCase` have basically identical set-up procedures. I might as well just combine the two classes. :+1:

Ichimonji10: In fact, I'm not sure I really even need any methods from this class except `test_status_code`. Calling `poll_tasks()` implicitly ensures that tasks have been spawned.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+        'importer_type_id': 'yum_importer',
+        'notes': {'_repo-type': 'rpm-repo'},
+    }
+
+
+class _TaskTimedOutException(Exception):
+    """Indicates that polling a task timed out."""
+
+
+def _poll_tasks(server_config, href):
```
dkliban: Would it be better to put this in some place that is shared by other tests? We need to poll tasks for other plugins also.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+
+    """
+    return _handle_response(requests.post(
+        server_config.base_url + href + 'actions/associate/',
+        json={'source_repo_id': source_repo_id},
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _add_yum_distributor(server_config, href, responses=None):
```
dkliban: Could this be more generic? We need to have ability to add any type of supported distributor. Other tests will add different types of distributors.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: Yes. But I'd like to do that in a separate commit.

Ichimonji10: Several of the other helper functions can also be extracted from this module too.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: I don't know how easy or hard that is to accomplish. What differs between the payload for creating a yum distributor and the payload for creating the other types of distributors? Just the `distributor_type_id`?

dkliban:
I started putting comments on everything that I thought could live outside the test. I think it would be perfectly OK to move those methods in a separate commit. 

This looks goo to merge.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
dkliban: I believe that the only difference is the distributor_type_id. I will confirm this and this can be updated in a separate PR.

Ichimonji10: Cool. If it's just the `distributor_type_id`, then something like this could be written:

```python
def add_distributor(server_config, href, distributor_type_id, responses=None):
    pass
```

Ichimonji10:
Thank you @dkliban. See #40.

Issue #40, opened 2015-11-17
Move utility functions to common module


Ichimonji10:
Move some common utility functions out of individual test modules and into `pulp_smash.utils`. In particular, I'm thinking of methods like `_poll_tasks()` and `_add_yum_distributor()` from module `pulp_smash.tests.rpm.api_v2.test_sync_publish`. See the discussion on #35 for more specifics.

This done, review the existing tests and see if any of the test modules can drop their own private utilities in favor of the global ones.

Depends on #35.

Ichimonji10:
Unassigning self from this issue for now, as I'm focused more on adding Ansible support. Ansible support is high-priority, as it'll unlock an entire category of new tests, and it should solve #32 and #31.

Moving helper functions from the various individual test modules to the `pulp_smash.utils` is a positive step forward, but I'd like to do a bit better than that. Right now, the various helper functions are fairly consistent and follow a similar implementation pattern. The pattern looks like this:

```python
def create_user(server_config, body, responses=None):
    """Create a user. Return the response body.

    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
    :param body: An object to encode as JSON and pass as the request body.
    :param responses: Same as :meth:`handle_response`.
    :returns: Same as :meth:`handle_response`.
    :raises: Same as :meth:`handle_response`.

    """
    return handle_response(requests.post(
        server_config.base_url + USER_PATH,
        json=body,
        **server_config.get_requests_kwargs()
    ), responses)
```

This works, but there's two annoyances here. First, the helper functions never worry about doing things like checking for an HTTP 202 (ACCEPTED) status code and polling the server for the action to complete. In some cases, we definitely do want to immediately accept the server's response and manually poll. But that won't be desirable everywhere, and by leaving the HTTP 202 check up to the end user, we're creating a recipe for intermittent bugs in our test scripts.

A second issue is that helper functions often differ in modest-but-meaningful ways. Here's a second helper function:

```python
def sync_repository(server_config, href, responses=None):
    """Sync a repository.

    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
    :param href: A string. The path to a repository.
    :param responses: Same as :meth:`handle_response`.
    :returns: Same as :meth:`handle_response`.
    :raises: Same as :meth:`handle_response`.

    """
    return handle_response(requests.post(
        server_config.base_url + href + 'actions/sync/',
        json={'override_config': {}},
        **server_config.get_requests_kwargs()
[...]

PR #48, opened 2015-12-01
Add puppet tests for syncing and publishing with puppet_distributor

peterlacko:
PR contains tests for following functionality:
* It is possible to create two puppet repos, with and without feed URLs respectively (CreateTestCase),
* Create repo with valid feed and run sync ends without errors for both valid and invalid query (SyncValidFeedTestCase),
* Create repo with invalid feed and run sync ends with error (SyncInvalidFeedTestCase),
* Following steps finish without errors (PublishTestCase)
  * Create two repositories: `foo` and `bar`.
  * upload custom puppet module to `foo`,
  * copy `foo` to `bar` and publish both,
  * query pulp server for modules using all query formats supported by given pulp version
  * download modules from their expected locations on pulp server file system,
  * verify integrity of downloaded modules.

### DIFF
--- a/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
@@ -1,0 +1,706 @@
+# coding=utf-8
+"""Test the API endpoints for puppet `repositories`_.
+
+This module assumes that the tests in
+:mod:`pulp_smash.tests.platform.api_v2.test_repository` hold true. The
+following trees of assumptions are explored in this module::
+
+    It is possible to create an puppet repo with a feed. (CreateTestCase)
+    ├── If a valid feed is given, the sync completes and no errors are
+    │   reported for both valid and ivalid query. (SyncValidFeedTestCase)
+    └── If an invalid feed is given, the sync completes and errors are
+        reported. (SyncInvalidFeedTestCase)
+
+    It is possible to create an puppet repo without a feed. (CreateTestCase)
+    └── It is possible to upload an puppet module to a repository, copy the
+        repository's content to a second repository, add distributor to
+        both repositories, publish them and query modules back. Data integrity
+        of modules queried from pulp server is preserved. (PublishTestCase)
+
+Assertions not explored in this module include:
+
+* Given an puppet repository without a feed, sync requests fail.
+* It is not possible to create two puppet repos with the same relative URL.
+* It is possible to upload a directory of puppet modules to an repository.
+* It is possible to upload content and copy it into multiple repositories.
+* It is possible to get content into a repository via a sync and publish it.
+* Functionality of puppet_install_distributor which install puppet module on
+  pulp server.
+* Functionality of puppet_file_distributor which makes puppet modules available
+  in single directory on pulp server over HTTPS.
+
+.. _repositories:
+   http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/cud.html
+
+"""
+from __future__ import unicode_literals
+
+from itertools import chain
+from packaging.version import Version
+from pulp_smash.config import get_config
+from pulp_smash.constants import CALL_REPORT_KEYS
+from pulp_smash.utils import (
+    create_repository,
+    delete,
+    get_importers,
+    handle_response,
+    poll_spawned_tasks,
+    publish_repository,
+    sync_repository,
+    uuid4,
+)
+from unittest2 import TestCase
+import hashlib
+import requests
+
+_VALID_PUPPET_FEED = 'http://forge.puppetlabs.com'
+_VALID_PUPPET_QUERY = 'pulp-pulp'
+_INVALID_PUPPET_QUERY = 'invalid_puppet_query'
+_PUPPET_QUERY_URL = (
+    'http://forge.puppetlabs.com'
+    '/v3/files/pulp-pulp-1.0.0.tar.gz'
+)
+_PUPPET_QUERY_PACKAGE = {
+    'name': 'pulp',
+    'author': 'pulp'
+}
+_PUPPET_DIRECT_URL = (
+    'https://repos.fedorapeople.org/'
+    'pulp/pulp/demo_repos/puppet_symlink/examplecorp-mymodule-0.1.0.tar.gz'
+)
+
+_TASK_END_STATES = ('canceled', 'error', 'finished', 'skipped', 'timed out')
+_REPO_PUBLISH_PATH = '/pulp/puppet/'  # + relative_url + puppet-module
+_CONTENT_UPLOADS_PATH = '/pulp/api/v2/content/uploads/'
+
+
+def _get_units(server_config, href, responses=None):
+    """Search for a repository's units."""
+    return handle_response(requests.post(
+        server_config.base_url + href + 'search/units/',
+        json={'criteria': {}},
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _start_content_upload(server_config, responses=None):
+    """Start a content upload.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.post(
+        server_config.base_url + _CONTENT_UPLOADS_PATH,
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _upload_file(server_config, href, content, responses=None):
+    """Upload a file and return the server's response.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param href: A string. The path to an open upload request.
+    :param content: The content of the file to upload.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.put(
+        server_config.base_url + href + '0/',
+        data=content,
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _copy_repo(server_config, source_repo_id, href, responses=None):
+    """Copy content from one repository to another.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param source_repo_id: A string. The ID of the source repository.
+    :param href: A string. The path to the repository on which the association
+        action is being performed. Content is copied to this repo.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.post(
+        server_config.base_url + href + 'actions/associate/',
+        json={'source_repo_id': source_repo_id},
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _query_repo_old(server_config, query_url, repo_id, responses=None):
+    """Query pulp puppet repo for presence of given unit (puppet ver. <3.3).
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param query_url: Url in Puppet Forge formatted URL. URL is formatted as
+                        '/api/v1/releases.json?module=modulename/author'
+    :param repo_id: id of queried repository
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    kwargs = server_config.get_requests_kwargs()
+    kwargs['auth'] = ('.', repo_id)
+    return handle_response(requests.get(
+        server_config.base_url + query_url,
+        **kwargs
+    ), responses)
+
+
+def _query_repo_middle(server_config, query_url, responses=None):
+    """Query pulp puppet repo for presence of given unit (puppet ver. 3.4-3.5).
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param query_url: Url in Puppet Forge formatted URL. URL is formatted as
+    '/pulp_puppet/forge/repository/repo_id/api/v1/releases.json?module=modulename/author'
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.get(
+        server_config.base_url + query_url,
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _query_repo_new(server_config, query_url, repo_id, responses=None):
+    """Query pulp puppet repo for presence of given unit (puppet ver. >=3.6).
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param query_url: Url in Puppet Forge formatted URL. URL is formatted as
+        /v3/releases?module=modulename/author
+    :param repo_id: id of queried repository
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    kwargs = server_config.get_requests_kwargs()
+    kwargs['auth'] = ('repository', repo_id)
+    return handle_response(requests.get(
+        server_config.base_url + query_url,
+        **kwargs
+    ), responses)
+
+
+def _import_puppet_module_into_repo(
+        server_config, upload_id, href, responses=None):
+    """Import an puppet from an upload into a repository.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param upload_id: A string. The ID of an upload request.
+    :param href: A string. The path to an puppet repository.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.post(
+        server_config.base_url + href + 'actions/import_upload/',
+        json={
+            'unit_key': {},
+            'unit_type_id': 'puppet_module',
+            'upload_id': upload_id
+        },
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _add_puppet_distributor(server_config, href, responses=None):
+    """Add a puppet distributor to a puppet repository.
+
+    The puppet distributor will not be published automatically and is available
+    over HTTP and HTTPS. The distributor ID and relative URL are random.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param href: A string. The path to the repository to which a distributor
+        shall be added.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.post(
+        server_config.base_url + href + 'distributors/',
+        json={
+            'auto_publish': False,
+            'distributor_id': uuid4(),
+            'distributor_type_id': 'puppet_distributor',
+            'distributor_config': {
+                'serve_http': True,
+                'serve_https': True,
+                'relative_url': '/' + uuid4(),
+            },
+        },
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _gen_puppet_repo_body():
+    """Return a semi-random dict that used for creating a puppet repo."""
+    return {
+        'id': uuid4(),
+        'importer_config': {},
+        'importer_type_id': 'puppet_importer',
+        'notes': {'_repo-type': 'puppet-repo'},
+    }
+
+
+class _BaseTestCase(TestCase):
+    """Provide a server config, and tear down created resources."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Provide a server config and an iterable of resources to delete."""
+        cls.cfg = get_config()
+        cls.attrs_iter = tuple()
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete created resources."""
+        for attrs in cls.attrs_iter:
+            delete(cls.cfg, attrs['_href'])
+
+
+class CreateTestCase(_BaseTestCase):
+    """Create two puppet repos, with and without feed URLs respectively."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create two puppet repositories, with and without feeds."""
+        super(CreateTestCase, cls).setUpClass()
+        cls.bodies = tuple((_gen_puppet_repo_body() for _ in range(2)))
+        cls.bodies[1]['importer_config'] = {
+            'feed': 'http://' + uuid4(),
+            'queries': [_VALID_PUPPET_QUERY],
+        }
+        cls.attrs_iter = tuple((
+            create_repository(cls.cfg, body) for body in cls.bodies
+        ))
+        cls.importers_iter = tuple((
+            get_importers(cls.cfg, attrs['_href']) for attrs in cls.attrs_iter
+        ))
+
+    def test_id_notes(self):
+        """Validate the ``id`` and ``notes`` attributes for each repo."""
+        for key in ('id', 'notes'):
+            for body, attrs in zip(self.bodies, self.attrs_iter):
+                with self.subTest((key, body, attrs)):
+                    self.assertIn(key, attrs)
+                    self.assertEqual(body[key], attrs[key])
+
+    def test_number_importers(self):
+        """Each repository should have only one importer."""
+        for i, importers in enumerate(self.importers_iter):
+            with self.subTest(i=i):
+                self.assertEqual(len(importers), 1, importers)
+
+    def test_importer_type_id(self):
+        """Validate the ``importer_type_id`` attribute of each importer."""
+        key = 'importer_type_id'
+        for body, importers in zip(self.bodies, self.importers_iter):
+            with self.subTest((body, importers)):
+                self.assertIn(key, importers[0])
+                self.assertEqual(body[key], importers[0][key])
+
+    def test_importer_config(self):
+        """Validate the ``config`` attribute of each importer."""
+        key = 'config'
+        for body, importers in zip(self.bodies, self.importers_iter):
+            with self.subTest((body, importers)):
+                self.assertIn(key, importers[0])
+                self.assertEqual(body['importer_' + key], importers[0][key])
+
+
+class SyncValidFeedTestCase(_BaseTestCase):
+    """Test creating repository with valid feed.
+
+    Create two repositories wih valid feeds, one with valid and one with
+    invalid query. Check that sync finish succesfully in both of them.
+
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create an puppet repositories with a valid feed and sync it."""
+        super(SyncValidFeedTestCase, cls).setUpClass()
+        bodies = tuple((_gen_puppet_repo_body() for i in range(2)))
+        bodies[0]['importer_config'] = {
+            'feed': _VALID_PUPPET_FEED,
+            'queries': [_VALID_PUPPET_QUERY],
+        }
+        bodies[1]['importer_config'] = {
+            'feed': _VALID_PUPPET_FEED,
+            'queries': [_INVALID_PUPPET_QUERY],
+        }
+        cls.attrs_iter = tuple((
+            create_repository(cls.cfg, body) for body in bodies
+        ))
+        cls.sync_repo = []  # raw responses
+        cls.task_bodies = tuple((chain.from_iterable(  # response bodies
+            poll_spawned_tasks(cls.cfg, call_report)
+            for call_report in (sync_repository(
+                cls.cfg,
+                attr_iter['_href'],
+                cls.sync_repo
+            ) for attr_iter in cls.attrs_iter)
+        )))
+
+    def test_start_sync_code(self):
+        """Assert the call to sync each repository returns an HTTP 202."""
+        for i, sync_response in enumerate(self.sync_repo):
+            with self.subTest(i=i):
+                self.assertEqual(sync_response.status_code, 202)
+
+    def test_task_error(self):
+        """Assert each task's "error" field is null."""
+        for i, task_body in enumerate(self.task_bodies):
+            with self.subTest(i=i):
+                self.assertEqual(task_body['error'], None)
+
+    def test_task_traceback(self):
+        """Assert each task's "traceback" field is null."""
+        for i, task_body in enumerate(self.task_bodies):
+            with self.subTest(i=i):
+                self.assertEqual(task_body['traceback'], None)
+
+    def test_task_progress_report(self):
+        """Assert no task's progress report contains error details."""
+        for i, task_body in enumerate(self.task_bodies):
+            with self.subTest(i=i):
+                self.assertEqual(
+                    task_body['progress_report']['puppet_importer']['metadata']['error_message'],  # noqa pylint:disable=line-too-long
+                    None
+                )
+
+
+class SyncInvalidFeedTestCase(_BaseTestCase):
+    """If an invalid feed is given, the sync completes with reported errors."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create puppet repository with an invalid feed and sync it."""
+        super(SyncInvalidFeedTestCase, cls).setUpClass()
+        body = _gen_puppet_repo_body()
+        body['importer_config'] = {'feed': 'http://' + uuid4()}
+        cls.attrs_iter = (create_repository(cls.cfg, body),)
+        cls.sync_repo = []  # raw responses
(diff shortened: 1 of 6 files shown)


Ichimonji10:
Please rebase onto master.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+    ├── If a valid feed is given, the sync completes and no errors are
+    │   reported for both valid and ivalid query. (SyncValidFeedTestCase)
+    └── If an invalid feed is given, the sync completes and errors are
+        reported. (SyncInvalidFeedTestCase)
+
+    It is possible to create an puppet repo without a feed. (CreateTestCase)
+    └── It is possible to upload an puppet module to a repository, copy the
+        repository's contents to a second repository, add distributor to both
+        both repositories, publish them. Data integrity of queried modules
+        from pul serverver is also verified (PublishTestCase)
```
Ichimonji10: Can we re-word this?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+
+Assertions not explored in this module include:
+
+* Given an puppet repository without a feed, sync requests fail.
+* It is impossible to create two puppet repositories with the same relative URL.  # noqa pylint:disable=line-too-long
+* It is possible to upload a directory of puppet modules to an repository.
+* It is possible to upload content and copy it into multiple repositories.
+* It is possible to get content into a repository via a sync and publish it.
+* Functionality of puppet_install_distributor which install puppet module on pulp server  # noqa pylint:disable=line-too-long
```
Ichimonji10: Please wrap these lines of text. There's no compelling reason to exceed the line length limit here.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+* Functionality of puppet_file_distributor which makes puppet modules available
+in single directory on pulp server over HTTPS.
```
Ichimonji10: Please take advantage of whitespace when wrapping lines, like so:

```
* first line
  second line
  third line
* fourth line
  fifth line
  sixth line
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+import requests
+import hashlib
+from itertools import chain
+from pulp_smash.config import get_config
+from pulp_smash.utils import (
+    uuid4,
+    create_repository,
+    delete,
+    handle_response,
+    poll_spawned_tasks,
```
Ichimonji10: Please sort these lines alphabetically.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+_CONTENT_UPLOADS_PATH = '/pulp/api/v2/content/uploads/'
+_PUPPET_QUERY_URL = (
+    'http://forge.puppetlabs.com'
+    '/v3/files/pulp-pulp-1.0.0.tar.gz'
+)
+_PUPPET_QUERY_PACKAGE = {
+    'name': 'pulp',
+    'author': 'pulp'
+}
+_PUPPET_DIRECT_URL = 'https://repos.fedorapeople.org/pulp/pulp/demo_repos/puppet_symlink/examplecorp-mymodule-0.1.0.tar.gz'  # noqa pylint:disable=line-too-long
```
Ichimonji10: Please take advantage of automatic line joins.

```python
_my_string = (
    'The first segment of the string does not end with a comma, so '
    'it is automatically joind with the second segment.'
)
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+
+
+def _sync_repo(server_config, href, responses=None):
```
Ichimonji10: There are quite a few identical methods in module `pulp_smash.tests.rpm.api_v2.test_sync_publish`. How about either splitting this PR into two commits and moving them to `pulp_smash.utils` in the first, or making an issue describing a refactor.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+    :param upload_id: A string. The ID of an upload request.
+    :param href: A string. The path to an puppet repository.
+    :param responses: A list, or some other object that supports the ``append``
+        method. If given, all server responses are appended to this object.
+    :returns: The server's JSON-decoded response.
+
+    """
+    return handle_response(requests.post(
+        server_config.base_url + href + 'actions/import_upload/',
+        json={'unit_key': {}, 'unit_type_id': 'puppet_module', 'upload_id': upload_id},  # noqa pylint:disable=line-too-long
```
Ichimonji10: Please wrap these lines. There's no compelling reason to exceed a line length limit here.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+class CreateTestCase(_BaseTestCase):
+    """Create two puppet repos, with and without feed URLs respectively."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create two puppet repositories, with and without feeds."""
+        super(CreateTestCase, cls).setUpClass()
+        cls.bodies = tuple((_gen_puppet_repo_body() for _ in range(2)))
+        cls.bodies[1]['importer_config'] = {'feed':
+                                            ''.join(['http://', uuid4()])}
```
Ichimonji10: Is there any reason to use an `http://` prefix? This doesn't result in a valid base URL as is.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+        cls.bodies[1]['importer_config']['queries'] = [_VALID_PUPPET_QUERY]
```
Ichimonji10: Can this be included in the line above?

```python
cls.bodies[1]['importer_config'] = {
    'feed': ...,
    'queries': ...,
}
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Create an puppet repositories with a valid feed and sync it."""
+        super(SyncValidFeedTestCase, cls).setUpClass()
+        bodies = tuple((_gen_puppet_repo_body() for i in range(2)))
+        bodies[0]['importer_config']['feed'] = _VALID_PUPPET_FEED
+        bodies[0]['importer_config']['queries'] = [_VALID_PUPPET_QUERY]
+        bodies[1]['importer_config']['feed'] = _VALID_PUPPET_FEED
+        bodies[1]['importer_config']['queries'] = [_INVALID_PUPPET_QUERY]
```
Ichimonji10: What's the reasoning behind creating two identical puppet repositories and syncing them? It seems one would suffice.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+
+class SyncInvalidFeedTestCase(_BaseTestCase):
+    """If an invalid feed is given, the sync completes with reported errors."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create puppet repository with an invalid feed and sync it."""
+        super(SyncInvalidFeedTestCase, cls).setUpClass()
+        body = _gen_puppet_repo_body()
+        body['importer_config']['feed'] = ''.join(['http://', uuid4()])
```
Ichimonji10: How about dropping `''.join()`, like so:

```python
body['importer_config']['feed'] = 'http://' + uuid4()
```

That said, is there any reason to use `http://` at all?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+        json={'source_repo_id': source_repo_id},
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _add_puppet_distributor(server_config, href, responses=None):
+    """Add a puppet distributor to an puppet repository.
+
+    The  puppet distributor will not automatically be published, and is available  # noqa pylint:disable=line-too-long
+    over HTTP and HTTPS. The distributor ID and relative URL are random.
```
Ichimonji10: Drop the linter disables and wrap these lines of text instead.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+            cls.attrs_iter[0]['id'],
+            cls.attrs_iter[1]['_href'],
+            cls.responses['copy'],
+        )
+        for call_report in cls.bodies['copy']:
+            poll_spawned_tasks(cls.cfg, call_report)
+
+        # Add a distributor to the repositories, then publish repositories
+        # to the distributors. Poll the publish request.
+        # First repository
```
Ichimonji10: I don't understand the meaningo of the "first repository" comment.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+            _PUPPET_QUERY_PACKAGE['name'],
+            '/',
+            _PUPPET_QUERY_PACKAGE['author']
+        )))
+        hrefs.append(''.join((
+            '/v3/releases?module=',
+            _PUPPET_QUERY_PACKAGE['name'],
+            '/',
+            _PUPPET_QUERY_PACKAGE['author']
+        )))
```
Ichimonji10: The two elements of `hrefs` are identical, so I see no reason to create an iterable. Just create a single variable. In addition, using `format` is more readable.

```python
href = '/v3/releases?module={}/{}'.format(
    _PUPPET_QUERY_PACKAGE['name'],
    _PUPPET_QUERY_PACKAGE['author'],
)
```

That done, you can drop the `i` variable from the loop below.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+    """Upload custom puppet module to a repository, publish and download it."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Upload puppet module to a repo, copy it to another, publish and download.
+
+        Create two puppet repositories, both without feeds. Upload an module to
+        the first repository. Copy this content to the second repository. Add a
+        distributor to the first repository, publish it, and download
+        the module uploaded earlier.
```
Ichimonji10: The contents of this test differ from what is described here. Puppet modules are not downloaded to one repository and then copied to the second. Is there a reason for the difference?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+    def test_call_reports_keys(self):
+        """Verify each call report has the correct keys."""
+        for step in {'import', 'copy'}:
+            with self.subTest(step=step):
+                self.assertEqual(
+                    set(self.bodies[step].keys()),
+                    {'error', 'result', 'spawned_tasks'},
+                )
+        # same for publish
+        for i in range(2):
```
Ichimonji10: Do not use hard-coded loop boundaries. Instead, use the length of `self.bodies['publish']`, or whatever it is you're walking through.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+                )
+
+    def test_call_reports_values(self):
+        """Verify no call report contains any errors."""
+        for step in {'import', 'copy'}:
+            for attr in ('result', 'error'):
+                with self.subTest((step, attr)):
+                    self.assertIsNone(self.bodies[step][attr])
+        # same for publish
+        for i in range(2):
```
Ichimonji10: Do not use hard-coded loop boundaries.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+    def test_search_units(self):
+        """Verify the two repositories have the same units."""
+        self.assertEqual(
+            set(unit['unit_id'] for unit in self.bodies['search units'][0]),
+            set(unit['unit_id'] for unit in self.bodies['search units'][1]),
+        )
+
+    def test_units_integrity(self):
+        """Verify integrity of modules uploaded to pulp server."""
+        for i in range(2):
```
Ichimonji10: Do not use hard-coded loop boundaries.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,650 @@
...
+                for response in self.responses[step]:
+                    self.assertEqual(response.status_code, code)
+
+    def test_call_reports_keys(self):
+        """Verify each call report has the correct keys."""
+        for step in {'import', 'copy'}:
+            with self.subTest(step=step):
+                self.assertEqual(
+                    set(self.bodies[step].keys()),
+                    {'error', 'result', 'spawned_tasks'},
```
Ichimonji10: Is this what every call report returns? If so, can we either split this PR into two commits, where the first commit adds the appropriate constant, or can we make a GitHub issue for this change?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Ok, thanks.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Sure, don't know why I left it like this.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Without `http://` create repo is failing on url validation, see [1]. Ok, I'll use `+` instead.

[1] https://pulp.plan.io/issues/1382

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: First one is querying for existing package, second for non existent. We want to make sure that syncing will success in both cases -- first repo will contain one module, second will be empty. Test for checking content of repos would be probably also useful here.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Same as [comment above](https://github.com/PulpQE/pulp-smash/pull/48#discussion_r46463952).

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, deleted, leftover from before rewriting.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks, done.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Fixed, docstring wasn't updated after test modification.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Got it, fixed.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Fixed.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Fixed.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Looks like [1] it's fixed. So I would make another PR for it.

[1] http://pulp.readthedocs.org/en/latest/dev-guide/conventions/sync-v-async.html#call-report

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: I think Issue [#40](https://github.com/PulpQE/pulp-smash/issues/40) discuss exactly this problem. We could move functions that are common for all repository types to `utils.py`. These are so far:
- `_sync_repo()`
- `_get_importers()`
- `_get_units()`
- `_start_content_upload()`
- `_upload_file()`
- `_copy_repo()`
- `_publish_repo()`

We can also move constants `_TASK_END_STATES`, `_CONTENT_UPLOADS_PATH` and list of call report keys [mentioned below](https://github.com/PulpQE/pulp-smash/pull/48#discussion_r46467945) to `constants.py`.

Ichimonji10:
@placko, if the tests in this new module are in good shape, please run them and capture some test output, and put that test output in the commit message.

While we're at it, please update the commit message. I notice that the PR message here on GitHub is significantly more fleshed out than the commit itself.

peterlacko:
@Ichimonji10 Ok, I will also split commits into two as [is written above](https://github.com/PulpQE/pulp-smash/pull/48#discussion_r46553801)

Ichimonji10:
Thank you.

Ichimonji10:
Version handling logic has been added on the master branch. The big new difference is that  `pulp_smash.config.ServerConfig` objects all now have a `version` attribute. This attribute is a `packaging.version.Version` object. These `Version` objects have quite intuitive usage:

```python
>>> Version('1') < Version('10')
True
>>> Version('2.7.3') > Version('2.7')
True
>>> Version('2.7.0') == Version('2.7')                                                                    
True
```

Version handling objects are useful for a number of reasons. Among others, they are superior to (codepoint-based) string comparisons:

```python
>>> '1' < '10'  # OK
True
>>> '6' < '10'  # uh oh
False
```

They're also better than tuple comparisons in that they properly implement [PEP 0440](https://www.python.org/dev/peps/pep-0440/). While that PEP has come from the Python community, it's useful in the general sense, and compatible with semantic versioning. Should we want less strict version handling, we can also use `packaging.version.parse`. See more info [here](https://packaging.pypa.io/en/latest/version/).

How do you use these new versioning objects in Pulp Smash? As mentioned, they're available on `ServerConfig` objects. Check it out:

```python
>>> from pulp_smash.config import ServerConfig
>>> cfg = ServerConfig()
>>> type(cfg.version)
<class 'packaging.version.Version'>
>>> str(cfg.version)
'1!0'
```

What in the world is `'1!0'`? It's a valid version string - no, really! It reads "epoch 1, version 0". Any version with "epoch 1" or above will always be greater than a version in "epoch 0", and by default, every version is implicitly in epoch 0:

```python
>>> Version('1!0') > Version('2.6')                                                                       
True
>>> Version('1!0') > Version('3.1')                                                                       
True
>>> Version('0!3.1') == Version('3.1')
True
```

By the way, version string epochs aren't some wacky weird thing. They're widely recognized and used. I use them in some of the software packages I maintain, due to the fact that the developers have messed up and used incorrect version numbers, or due to developers changing numbering schemes.

Back to Pulp Smash.

How can we use these versions in our code? Let's take an example from this pull request. This bit of code is only valid on Pulp 2.7 and above:

```python
        # Query modules on server. Data integrity is tested in tests below.
        query_href = '/v3/releases?module={}/{}'.format(
            _PUPPET_QUERY_PACKAGE['name'],
            _PUPPET_QUERY_PACKAGE['author']
        )
        cls.bodies['query units'] = tuple((
            _query_repo(
                cls.cfg,
                query_href,
                body['repo_id'],
                cls.responses['query units']
            ) for body in cls.bodies['distribute']
        ))
```

Now, I don't know the details of Pulp's behaviour here. But I'm going to assume that Pulp 2.6 and older also let you query for puppet modules, but with a different query. You can add in some conditional logic like so:

```python
        # Query modules on server. Data integrity is tested in tests below.
        if cls.cfg.version < Version('2.7'):
            query_href = …
        else:
            query_href = '/v3/releases?module={}/{}'.format(
                _PUPPET_QUERY_PACKAGE['name'],
                _PUPPET_QUERY_PACKAGE['author']
            )
        cls.bodies['query units'] = tuple((
            _query_repo(
                cls.cfg,
                query_href,
                body['repo_id'],
                cls.responses['query units']
            ) for body in cls.bodies['distribute']
        ))
```

Just to be clear, this works as-is with absolutely no changes to the end user. An end user can now add a "version" attribute to their configuration file to explicitly state which version of Pulp they're testing against, like so:

```json
{
    "default": {
        "version": "2.6.3",
        "base_url": "https://192.168.121.222",
        "verify": false,
        "auth": ["admin", "admin"]
    }
}
```

But this is entirely optional. Again, by default, Pulp Smash will assume that you're testing the most up-to-date development version of Pulp. It does this by automatically setting a vesrion of `'1!0'` if none is specified.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,655 @@
...
+                cls.bodies['distribute'][i]['id'],
+                cls.responses['publish'],
+            ) for i in range(2)
+        ))
+        for i in range(2):
+            for call_report in cls.bodies['publish'][i]:
+                poll_spawned_tasks(cls.cfg, call_report)
+
+        # Query modules on server. Data integrity is tested in tests below.
+        query_href = '/v3/releases?module={}/{}'.format(
```
dkliban: We need to test that you can use the older versions of the puppet forge API (puppet clients prior to version 3.3).

On my dev environment I am able to run:

http --verify=no GET http://.:repo1@hostname/api/v1/releases.json?module=domcleal/katellovirt

In response I get json that tells me where that module can be downloaded. The basic auth in the above request is explained here [0].

We should also attempt to query the forge API used for clients between versions 3.3 and 3.6. The way the request is formulated is described here [1]
These two additional scenarios will work on Pulp 2.6+, however the test that is already implemented will only work for Pulp version 2.7+.

[0] http://pulp-puppet.readthedocs.org/en/latest/tech-reference/forge_api.html#basic-auth
[1] http://pulp-puppet.readthedocs.org/en/latest/user-guide/recipes.html#installing-with-puppet-client-3-3

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,655 @@
...
+            poll_spawned_tasks(cls.cfg, call_report)
+
+        # Add a distributor to the repositories, then publish repositories
+        # to the distributors. Poll the publish request.
+        cls.bodies['distribute'] = tuple((
+            _add_puppet_distributor(
+                cls.cfg,
+                cls.attrs_iter[i]['_href'],
+                cls.responses['distribute']
+            ) for i in range(2)
```
Ichimonji10: Hard-coded boundaries are fragile. We can avoid that fragility by basing our index off the thing being iterated through:

```python
_add_puppet_distributor(
    cls.cfg,
    cls.attrs_iter[i]['_href'],
    cls.responses['distribute'],
) for i in range(len(cls.attrs_iter))
```

It's even nicer to iterate through the objects themselves intead of using indices. An index is annoying because you have to go hunting for every place the index variable is used. We can make life easier for the reader by just iterating directly through the target set of objects, which means there's one less piece of information to hunt for!

```python
_add_puppet_distributor(
    cls.cfg,
    attrs['_href'],
    cls.responses['distribute'],
) for attrs in cls.attrs_iter
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,655 @@
...
+            _publish_repo(
+                cls.cfg,
+                cls.attrs_iter[i]['_href'],
+                cls.bodies['distribute'][i]['id'],
+                cls.responses['publish'],
+            ) for i in range(2)
+        ))
+        for i in range(2):
+            for call_report in cls.bodies['publish'][i]:
+                poll_spawned_tasks(cls.cfg, call_report)
```
Ichimonji10: I think the index variable can be dropped:

```python
        for call_report in cls.bodies['publish']:
            poll_spawned_tasks(cls.cfg, call_report)
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,655 @@
...
+                attr_iter['_href'],
+                cls.sync_repo
+            ) for attr_iter in cls.attrs_iter)
+        )))
+
+    def test_start_sync_code(self):
+        """Assert the call to sync each repository returns an HTTP 202."""
+        for i in range(2):
+            with self.subTest(i=i):
+                self.assertEqual(self.sync_repo[i].status_code, 202)
```
Ichimonji10: Please avoid this hard-coded range and, instead, iterate directly through the objects being tested. The `i` variable is only generated for ease of troubleshooting should the test fail.

```python
        for i, response in enumerate(self.sync_repo):
            with self.subTest(i=i):
                self.assertEqual(resopnse.status_code, 202)
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
Ichimonji10: @dkliban I don't understand the bit about "we should attempt to query the forge API". Are you saying that the portion of the test case below this line is simply not applicable on Pulp 2.6 and older?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
Ichimonji10: I've gone ahead and done this for you. You can use the new `pulp_smash.constants.CALL_REPORT_KEYS` constant by rebasing onto the master branch.

Ichimonji10:
FYI, I've moved added functions to module `pulp_smash.utils`:

* `get_importers`
* `publish_repository`
* `sync_repository`

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: `cls.bodies['publish']` is tuple so we have to iterate over it like
```python
for publish_report in cls.bodies['publish']:
    for call_report in publish_report:
        poll_spawned_tasks(cls.cfg, call_report)
```
but got the idea, avoid hard coded indexes whenever possible.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: OK.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
Ichimonji10: Thanks for the correction. Yeah, the idea is to just avoid hard coded indices.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: After discussion with @dkliban querying the module will look as follows:
* for pulp >= 2.7, 3 puppet queries formats will be tested, namely for puppet <=3.3, 3.3 - 3.5 and >=3.6
* for pulp <= 2.6 only queries from puppet <3.6 will be tested, as pulp in these versions doesn't support queries to new forge API [1]

I hope it makes sense.
[1] https://forgeapi.puppetlabs.com/

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,705 @@
...
+    kwargs = server_config.get_requests_kwargs()
+    kwargs['auth'] = ('repository', repo_id)
+    return handle_response(requests.get(
+        server_config.base_url + query_url,
+        **kwargs
+    ), responses)
+
+
+def _import_puppet_to_repo(server_config, upload_id, href, responses=None):
+    """Import an puppet from an upload into a repository.
```
dkliban: Import a puppet module from an upload request into a repository.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,705 @@
...
+    """
+    kwargs = server_config.get_requests_kwargs()
+    kwargs['auth'] = ('repository', repo_id)
+    return handle_response(requests.get(
+        server_config.base_url + query_url,
+        **kwargs
+    ), responses)
+
+
+def _import_puppet_to_repo(server_config, upload_id, href, responses=None):
```
dkliban: _import_puppet_module_into_repo

or do you think that's too long of a name?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,705 @@
...
+                'serve_https': True,
+                'relative_url': '/' + uuid4(),
+            },
+        },
+        **server_config.get_requests_kwargs()
+    ), responses)
+
+
+def _gen_puppet_repo_body():
+    """Return a semi-random dict that used for creating an puppet repo."""
```
dkliban: s/an puppet/a puppet/

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -0,0 +1,705 @@
...
+
+        module_uris = []  # Save modules' uri on server
+        for query_body in cls.bodies['query units']:
+            try:
+                module_uris.append(
+                    query_body['{}/{}'.format(
+                        _PUPPET_QUERY_PACKAGE['name'],
+                        _PUPPET_QUERY_PACKAGE['author'])][0]['file']
+                )
+            except KeyError:
```
dkliban: When does this happen?

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: `_import_puppet_module_into_repo` is more accurate, I will change it.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Thanks.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
peterlacko: Here I'm not sure what is best. The reason for try except block is: New forge API returns dictionary in difrerent format, so when trying to get module's uri on server, we got `KeyError` and thus second dictionary format is tried. I could avoid this by adding another key to `cls.bodies` or access them by index but I found first option (try except) least painful.

Ichimonji10:
@peterlacko, I've walked through this pull request and given it a thorough once-over. Please take a look at https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:puppet, and let me know if I've messed anything up.

peterlacko:
@Ichimonji10: looks OK.

Ichimonji10:
Merged as c0d8d08311644181ce3181b7f12133602df968e7.

Issue #31, opened 2015-11-06
Add ability to fiddle with services Pulp depends on


Ichimonji10:
Let's add some mechanism(s) that let us fiddle with the services that Pulp depends on. Pulp is designed to be resilient against failure of certain services, and it would be nice if we could start or stop services at run-time to test that resilience.

At its simplest, the mechanism would look like this:

```python
def twiddle_service(state, services):
    # determine if remote system has `systemctl` available, or some other service manager
    # ssh some_user@remote_system systemctl state ' '.join(services)
```

One design consideration is what to do if the service you'd like to test might go by multiple names. For example, the message broker might be either RabbitMQ or Qpid. Do we make the user explicitly name the service to be twiddled, or do we provide abstraction? Another question is what to do if a non-zero return code is received. Do we raise an exception? (No, really. systemctl has a range of valid non-zero return codes.)

Ichimonji10:
PR #67 solves this issue. Within that PR is a function, `pulp_smash.utils.get_broker`. It has the following signature:

def get_broker(server_config):
    """Build an object for managing the target system's AMQP broker."""
    return cli.Service(server_config, ...)  # either qpidd or rabbitmq

Another option for solving this problem is to create a subclass of `Service`, like so:

```python
class Broker(Service):
    def __init__(self, server_config):
        service = …  # 'qpidd', 'rabbitmq', etc.
        super(Broker, self).__init__(server_config, service)
```

However, this approach has two drawbacks:

* It favors inheritance over composition.
* It violates the Liskov substitution principle, due to its `__init__` method requiring a more specific set of arguments.

That said, the ultimate test of whether some code is good lies not in its theoretical prettiness, but whether it clearly and correctly accomplishes its job, and whether it is flexible enough to accomodate inevitable changes in requirements and usage patterns. And my opinion is not universally shared: @bmbouter favors the class-based approach.

On a related note, I'm trying to keep some level of separation between generic CLI-handling logic and Pulp-specific logic. The API and CLI handling code is, thus far, quite generic and could in theory by used by Robottelo or anyone else who wants some Python code for executing commands.

PR #49, opened 2015-12-04
Skip tests affected by bug 695

Ichimonji10:
Add a decorator for skipping tests, `pulp_smash.utils.require`. Use it to skip a test that is known to fail under Pulp 2.6 and older. See individual commit messages for more details.

### DIFF
--- a/pulp_smash/tests/platform/api_v2/test_repository.py
+++ b/pulp_smash/tests/platform/api_v2/test_repository.py
@@ -20,7 +20,7 @@ import requests
 
 from pulp_smash.config import get_config
 from pulp_smash.constants import REPOSITORY_PATH, ERROR_KEYS
-from pulp_smash.utils import create_repository, delete, uuid4
+from pulp_smash.utils import create_repository, delete, require, uuid4
 from requests.exceptions import HTTPError
 from unittest2 import TestCase
 
@@ -54,6 +54,7 @@ class CreateSuccessTestCase(TestCase):
             with self.subTest(self.bodies[i]):
                 self.assertEqual(response.status_code, 201)
 
+    @require('2.7')  # https://pulp.plan.io/issues/695
     def test_location_header(self):
         """Assert that the Location header is correctly set in the response."""
         for response, attrs in zip(self.responses, self.attrs_iter):
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 
 import requests
 import uuid
+from functools import wraps
+from packaging.version import Version
 from pulp_smash.constants import REPOSITORY_PATH, USER_PATH
 from time import sleep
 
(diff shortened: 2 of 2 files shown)


Ichimonji10:
At this point, you have to manually configure Pulp Smash and tell it which version of Pulp you're targeting, and there's also no integration with the bug tracker at pulp.plan.io. But it's a start, and it demonstrates how to use the new "version" attribute on the `pulp_smash.config.ServerConfig` object.

bmbouter:
This looks good. I do have one suggestion. Can a decorator be introduced? It would be like this:

```
     @requires('2.7')
     # 'https://pulp.plan.io/issues/695'
     def test_location_header(self):
```

Or maybe this:

```
     @requires('2.7', 'https://pulp.plan.io/issues/695')
     def test_location_header(self):
```

Ichimonji10:
Yes, we can do that. The problem with using decorators here is that they are evaluated when defined, but we don't want to be instantiating `ServerConfig` objects and inspecting their "version" attribute or possibly calling `@skip` until the tests are executed. As a consequence, the decorators are going to be complicated and bug-prone.

It's definitely doable. It's done in Robottelo, and @elyezer in particular has expertise with writing them. But I'm nervous about adopting something like that right away. This approach is not quite as elegant, but it serves the lowest common denominator. My spidey senses tell me to do other things (like integrate with pulp.plan.io and expand the test suite) first.

Ichimonji10:
By the way, I do like the two-argument form you've given. I like the idea of requiring that users provide a bug URL in order to skip a test.

bmbouter:
The complexity of introducing a decorator versus integrating with a web server are significantly different. Regarding making a decorator I could write one up and post it here if that is helpful.

I really think this is an idiom we want to include from the start. I'm not concerned if `ServerConfigs` get instantiated a whole lot. We could easily add a caching there so it's only read from disk the first time.

What do you think?

Also if we want to be sure it's a bug (which could be good to require) we could use it like:

```
@requires('2.7', 'https://pulp.plan.io/issues/695')
```

Ichimonji10:
> We could easily add a caching there so it's only read from disk the first time.

`pulp_smash.config.get_config()` already has caching, so instantiating lots of objects isn't an issue.

Ichimonji10:
What I'm much more worried about is that `ServerConfig` objects might be instantiated when each decorator is evaluated, and that they would persist until when each test is run. I'd strongly prefer if `ServerConfig` objects were instantiated as lazily as possible. There's no problem with the agressive instantiation approach at this time, but it invites unexpected behaviour should we ever need to do something like clear the `pulp_smash.config._CONFIG` cache at run-time.

Ichimonji10:
But hey, if we can get a good decorator now - that'd be great. Please do put something together if you have a few minutes. :smile:

Ichimonji10:
@bmbouter Actually, is there even any need to state the version at which a bug is fixed? We could just do something like this:

```python
@skip_if_open(695)
def test_foo(self):
    pass
```

And at run-time, we can grab a `ServerConfig` object and talk to the server, and compare the version on the former with the "target release" attribute on the latter.

dkliban:
This would require integration with Redmine at this time. It would be advantageous to hold off on doing that integration till later.

Ichimonji10:
Why so? Integration is extremely easy.

dkliban:
Even though it is easy, it is still another step that we have to take before we can start skipping tests.

Ichimonji10:
We can skip tests right now!

Ichimonji10:
Basically, if we're going to go above and beyond serving the lowest common denominator, let's make sure what we come up with is orthogonal to what we already have.

bmbouter:
Here is a simple decorator for it. Personally I think introducing the decorator is good to do now. More than redmine integration, I think all the core tools bugs need to be done. I like the integration, but I don't think it's blocking test writing. I'm still blocked on #31

```
def require(func, version_string):
    def new_func(self, *args, **kwargs):
        if self.cfg.version < Version(version_string):
            self.skipTest('Test requires version %s' % version_string)
        else:
            return func(self, *args, **kwargs)
    return new_func
```

Ichimonji10:
@bmbouter Concise and orthogonal. I like it.

Ichimonji10:
I'll add that as a second commit in this PR and ask for a re-review then.

This may come in handy later:

```python
#!/usr/bin/env python
"""Fetch the target release of a bug.

This can fail in at least two obvious cases:

* There is an error while communicating with the server. (e.g. cannot be
  reached)
* There is no "target release" set on the bug.

In either case, an appropriately cautious response is to assume that the bug is
applicable and open.

"""
from __future__ import print_function

import requests

response = requests.get('https://pulp.plan.io/issues/695.json')
response.raise_for_status()
target_releases = [
    custom_field for custom_field in response.json()['issue']['custom_fields']
    if custom_field['id'] == 4  # 'name': 'Target Release'
]
assert len(target_releases) == 1
print(target_releases[0]['value'])
```

Ichimonji10:
Decorator usage boils down to this:

```python
@require('2.7')  # https://pulp.plan.io/issues/695
def test_foo(self):
    pass
```

I implemented the decorator so that it will fail if an invalid version string is passed. This will throw an exception:

```python
@require('2.7.foobar')  # https://pulp.plan.io/issues/695
def test_foo(self):
    pass
```

bmbouter:
@Ichimonji10 nice! I really like it.

Ichimonji10:
I've manually verified this change one more time. When the test suite is run
against Pulp 2.6, we've correctly removed two failures and added one skipped
test.

    branch         results
    -------------  ---------------------------------------
    master         FAILED (failures=9, errors=1, skipped=5)
    skip-failures  FAILED (failures=7, errors=1, skipped=6)

When the test suite is run against Pulp 2.7, nothing changes:

    branch         results
    -------------  ---------------------------------------
    master         FAILED (failures=8, errors=1, skipped=5)
    skip-failures  FAILED (failures=8, errors=1, skipped=5)

Merging changes now.

PR #54, opened 2015-12-16
Make Pulp Smash test MODIFIED issues

Ichimonji10:
The `bug_is_testable` and `bug_is_untestable` functions talk to the pulp.plan.io bug tracker, inspect the state of a named bug and then tell the caller whether the named bug is testable or not. Make these functions treat MODIFIED issues as test-worthy. Both before and after this patch:

    2.6           OK (skipped=13)
    2.7           OK (skipped=12)
    dev           OK (skipped=12)

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -23,9 +23,9 @@ _UNTESTABLE_BUGS = frozenset((
     'NEW',  # bug just entered into tracker
     'ASSIGNED',  # bug has been assigned to an engineer
     'POST',  # bug fix is being reviewed by dev ("posted for review")
-    'MODIFIED',  # bug fix has been accepted by dev
 ))
 _TESTABLE_BUGS = frozenset((
+    'MODIFIED',  # bug fix has been accepted by dev
     'ON_QA',  # bug fix is being reviewed by qe
     'VERIFIED',  # bug fix has been accepted by qe
     'CLOSED - CURRENTRELEASE',


Ichimonji10:
@bmbouter @dkliban @bowlofeggs and others: is this change appropriate? One or two people on IRC asked for this change.

rbarlow:
@Ichimonji10 Modified means that the change is in our git repos, but that it is not in an RPM build. I think this make the questions maybe a little complicated. Does pulp-smash know if it is testing a source install (like the dev environment) vs. an RPM install? Because a MODIFIED bug will be fixed in a source install and be broken in the RPM install. I know that sounds complicated, but it's our way of tracking WHERE a bug fix can be expected to be found.\

Would it be crazy to make pulp-smash aware of whether it is testing a source install vs. an RPM install, and let it dynamically choose whether MODIFIED or ON_QA is appropriate?

If the goal is to only test source installs then I think this change is good.

dkliban:
@rbarlow I think that it is safe to say that pulp-smash will be running on an RPM install that contains the fix. We currently build RPMs nightly, deploy Pulp using those RPMs, run pulp-smash. So if something is in MODIFIED state then it should be in the newly build RPM.

Ichimonji10:
> Does pulp-smash know if it is testing a source install (like the dev environment) vs. an RPM install?

Nope. You can tell Pulp Smash which version of Pulp it is testing, but you cannot at this time tell it whether it's testing a source or RPM install.

Ichimonji10:
> Because a MODIFIED bug will be fixed in a source install and be broken in the RPM install.

Gotcha. And then an ON_QA bug will be fixed in both a source and RPM install.

bmbouter:
If we wait until the issue is at ON_QA there will be a multi-week, perhaps multi-month delay in tests becoming enabled. We really need this @ MODIFIED.

I think [dkliban's comment](https://github.com/PulpQE/pulp-smash/pull/54#issuecomment-165265623) suggests that it will work in both RPM based installs and source installs since RPM based installs are rebuilt from source nightly.

Ichimonji10:
Fixed in 8ce9edae00eb65834fc47e81fcd2f7c22c2dd296

PR #57, opened 2015-12-17
rpm sync test verifies unit counts on repo

mhrivnak:


pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -291,6 +295,20 @@ def test_task_progress_report(self):
                     0
                 )
 
+    def test_unit_count_on_repo(self):
+        """
+        Verify that the sync added the correct number of units to the repo.
```
Ichimonji10: Flake8 is asking for a line to be inserted between the summary line ("Verify that the sync added...") and the remainder of the docstring. This will pass:

```
"""
Verify that the sync added the correct number of units to the repo.

Looks at the content counts on the repo.
"""
```

This will pass too (it's the docstring form used elsewhere in the project):

```
"""Verify that the sync added the correct number of units to the repo.

Looks at the content counts on the repo.

"""
```

pulp_smash/utils.py
```diff
@@ -124,6 +125,22 @@ def get_importers(server_config, href, responses=None):
     ), responses)
 
 
+def get_repo(server_config, href, responses=None):
```
Ichimonji10: FYI, this method can be used to get any sort of resource from the server. Can we rename it and update its docstring to reflect its generic nature?

pulp_smash/utils.py
```diff
@@ -9,6 +9,7 @@
 from packaging.version import Version
 from pulp_smash.constants import REPOSITORY_PATH, USER_PATH
 from time import sleep
+from urlparse import urljoin
```
Ichimonji10: Is there a benefit to using urljoin? If so, can we switch over to using urljoin in all of the utility functions in a separate commit? (You can file a GitHub issue for this if you'd like to kick this can down the road.)

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -291,6 +295,20 @@ def test_task_progress_report(self):
...
+        Looks at the content counts on the repo.
+
+        This also verifies that the counts themselves are getting set on the
+        repo.
+        """
+        counts = self.repos_iter[0].get('content_unit_counts', {})
+        self.assertEqual(counts.get('rpm'), 32)
+        self.assertEqual(counts.get('erratum'), 4)
+        self.assertEqual(counts.get('package_group'), 2)
+        self.assertEqual(counts.get('package_category'), 1)
```
Ichimonji10: Where do these numbers come from?

pulp_smash/utils.py
mhrivnak: good call. will do.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
mhrivnak: I looked at the upstream repo metadata. Should I make a note of that?

Ichimonji10: Yes, please do. Will these counts change as time goes on?

pulp_smash/utils.py
mhrivnak: Pure addition of path parts is dangerous, because standard "urljoin" logic, which follows rules in various RFCs, behaves differently. For example:

```
>>> '/a/b/c' + '/d/e'
'/a/b/c/d/e'
>>> urljoin('/a/b/c', '/d/e')
'/d/e'
```

Pulp should produce paths that behave correctly with urljoin, and we've had problems in the past with developers making "relative paths" that have a leading slash.

I expect most clients of our REST API will use a urljoin implementation to combine paths.

Do you prefer that I undo this use of urljoin and file an issue to change them all at once? Or can we leave this as an incremental improvement, and file an issue to change the rest?

pulp_smash/utils.py
Ichimonji10: This function seems to follow the same common pattern as other functions in this module. Can you add a new test case in module `tests.test_utils`? You should be able to create a class like `GetTestCase(CommonAssertionsMixin, TestCase)`.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
mhrivnak: Haven't changed since March of 2012.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -265,6 +266,9 @@ def setUpClass(cls):
             cls.sync_repo,
         )
         cls.task_bodies = tuple(poll_spawned_tasks(cls.cfg, report))
+        cls.repos_iter = tuple((
+            get_repo(cls.cfg, attrs['_href']) for attrs in cls.attrs_iter
+        ))
```
Ichimonji10: `attrs_iter` is an iterable of dicts for two reasons:

* The base class, `_BaseTestCase`, assumes that an `attrs_iter` class attribute is present.
* Many other classes do the same, and adopting the same convention here allows people familiar with patterns in Pulp Smash more easily understand this code.

In this particular test case, `attrs_iter` only contains one dict. That's something you can take advantage of, and we already do that a few lines up by writing `cls.attrs_iter[0]['_href']`. With that in mind, try this out:

```python
    @classmethod
    def setUpClass(cls):
        """Create an RPM repository with a valid feed and sync it."""
        super(SyncValidFeedTestCase, cls).setUpClass()
        body = _gen_rpm_repo_body()
        body['importer_config']['feed'] = _VALID_FEED
        repo_href = create_repository(cls.cfg, body)['_href']
        cls.sync_repo = []  # raw responses
        report = sync_repository(cls.cfg, repo_href, cls.sync_repo)
        cls.task_bodies = tuple(poll_spawned_tasks(cls.cfg, report))
        cls.attrs_iter = (get(cls.cfg, repo_href),)  # see parent class
```

pulp_smash/utils.py
Ichimonji10: > Pulp should produce paths that behave correctly with urljoin, and we've had problems in the past with developers making "relative paths" that have a leading slash.

So the benefit here is that, by using urljoin, we ensure that the relative paths given by Pulp do not have a leading slash?

Ichimonji10: > Do you prefer that I undo this use of urljoin and file an issue to change them all at once?

I'd like to do this all at once through the entire `pulp_smash.utils` module. Having more targeted commits makes it easier to do things like use `git bisect`, revert commits and walk through the commit log. It's getting toward the end of the day - I'd be delighted if we get a GitHub issue up by the end of today.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: Gotcha. In that case, I guess hard-coded numbers should be OK. Let's just reference where they're coming from.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: Basically, `cls.attrs_iter` and `cls.repos_iter` are duplicate variables. The change I've suggested merges them.

mhrivnak: I'm definitely not crazy about the readability here, and I only did it this way to be consistent with ``self.attrs_iter``. I'm very happy to do it differently.

That said, now I really don't understand what ``self.attrs_iter`` is supposed to be. The name is not descriptive, and your example seems to have assigned completely different data to it than before.

Can we just not use that attribute name? It's confusing, and doesn't seem to be helping. What if the last line is:

``self.repo = get(cls.cfg, repo_href)``

pulp_smash/utils.py
mhrivnak: I will make it so.

Relative paths with leading slashes are just one demonstration of how the behavior is different. Here is another where string addition would produce an incorrect result:

```
>>> urljoin('http://foo.com', 'a/b/c')
'http://foo.com/a/b/c'
```

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: > That said, now I really don't understand what self.attrs_iter is supposed to be. The name is not descriptive, and your example seems to have assigned completely different data to it than before.

`attrs_iter` is an iterable of attributes. It can be an iterable of repository attributes, or an iterable of user attributes, or anything else really.

Ichimonji10: > your example seems to have assigned completely different data to it than before.

When you create a repository, you should get back attributes about the repositories created. Same for reading and updating. I'm assuming that there's minor differences, if any, between the response to a POST and GET request.

Ichimonji10: > Can we just not use that attribute name? It's confusing, and doesn't seem to be helping. What if the last line is:
>
> `self.repo = get(cls.cfg, repo_href)`

We could. But at that point, we do have duplication between the `attrs_iter` and `repo` variables. They both describe the same thing.

Another option is to read the repository in the test case itself. Reading doesn't mutate anything server-side, so this shouldn't affect any tests:

```python
    def test_unit_count_on_repo(self):
        """Verify that the sync added the correct number of units to the repo.

        Looks at the content counts on the repo.

        This also verifies that the counts themselves are getting set on the
        repo.

        """
        attrs = get(self.cfg, self.attrs_iter[0]['_href'])
        counts = attrs.get('content_unit_counts', {})
        self.assertEqual(counts.get('rpm'), 32)
        self.assertEqual(counts.get('erratum'), 4)
        self.assertEqual(counts.get('package_group'), 2)
        self.assertEqual(counts.get('package_category'), 1)
```

Ichimonji10: > I'm definitely not crazy about the readability here,

Definitely not. I'm intentionally being basic with our API abstractions. It's on my mind, though. See #40 as an example of some of the thought I've put into this. (The comments on that issue are already a bit obsolete. I've been wondering if we can do anything nice with named tuples to cut down on those function signatures.)

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -257,14 +258,15 @@ def setUpClass(cls):
         super(SyncValidFeedTestCase, cls).setUpClass()
         body = _gen_rpm_repo_body()
         body['importer_config']['feed'] = _VALID_FEED
-        cls.attrs_iter = (create_repository(cls.cfg, body),)  # see parent cls
+        cls.repo = create_repository(cls.cfg, body)
```
Ichimonji10: In an earlier comment, you asked about the meaning of the `attrs_iter` ("an iterable of attributes") variable. It's used by the parent class, `_BaseTestCase`, in its `tearDownClass` method:

```python
    @classmethod
    def tearDownClass(cls):
        """Delete created resources."""
        for attrs in cls.attrs_iter:
            delete(cls.cfg, attrs['_href'])
```

The tear down procedure provided by the parent `_BaseTestCase`

mhrivnak: Ok, gotcha. I'll change the test back to using that attribute name then.

Michael

Ichimonji10: Thank you. It's not ideal, for sure.

Ichimonji10:
Looks good to me!

Before change (4 tests):

    2.7           OK
    dev (2.8)     OK

After change (5 tests):

    2.7           OK
    dev (2.8)     OK

Command used:

    PULP_SMASH_CONFIG_FILE=$PS_CONFIG python -m unittest2 \
    pulp_smash.tests.rpm.api_v2.test_sync_publish.SyncValidFeedTestCase

Ichimonji10:
(I've rebased locally and am waiting for Travis to finish.)

Ichimonji10:
Merged as 543532f844b1c5962b691dfd5fb3115f150176d9.

PR #59, opened 2015-12-18
Use urljoin in `pulp_smash.utils`

### LINKED ISSUE #58: Do URL and path joining with urljoin
Many places in the tests combine URLs and paths with simple string addition. This has the potential to produce incorrect results.

Pure addition of path parts is dangerous, because standard "urljoin" logic, which follows rules in various RFCs, behaves differently. For example:

```
>>> '/a/b/c' + '/d/e'
'/a/b/c/d/e'
>>> urljoin('/a/b/c', '/d/e')
'/d/e'
```

Pulp should produce paths that behave correctly with urljoin, and we've had problems in the past with developers making "relative paths" that have a leading slash.

Relative paths with leading slashes are just one demonstration of how the behavior is different. Here is another where string addition would produce an incorrect result:

```
>>> urljoin('http://foo.com', 'a/b/c')
'http://foo.com/a/b/c'
```

Example of the undesirable use of string addition: https://github.com/PulpQE/pulp-smash/blob/668f4b7fd5e88c5f9290e4d3b251be587e026f43/pulp_smash/utils.py#L106

I expect most clients of our REST API will use a urljoin implementation to combine paths, and pulp-smash should do the same.

Ichimonji10:
Fix #58:

> Many places in the tests combine URLs and paths with simple string addition.
> This has the potential to produce incorrect results. […] Pulp should produce
> paths that behave correctly with urljoin, and we've had problems in the past
> with developers making "relative paths" that have a leading slash.

Using urljoin instead of plain string concatenation lets Pulp Smash test the
paths returned by Pulp. If Pulp returns a relative path with a leading slash,
plain string concatenation will produce a path containing a (typically harmless
but incorrect) `//`, but urljoin will produce a just-plain-wrong URL.

Test results both before and after change:

    ============  =========================================
    ============  =========================================
    2.7           OK (skipped=13)
    dev (2.8)     FAILED (failures=1, errors=2, skipped=12)
    ============  =========================================

Command used:

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -11,6 +11,12 @@ from pulp_smash.constants import REPOSITORY_PATH, USER_PATH
 from time import sleep
 
 
+try:  # try Python 3 import first
+    from urllib.parse import urljoin  # pylint:disable=no-name-in-module
+except ImportError:
+    from urlparse import urljoin  # pylint:disable=import-error
+
+
 _TASK_END_STATES = ('canceled', 'error', 'finished', 'skipped', 'timed out')
 
 # These are all possible values for a bug's "status" field.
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@mhrivnak

elyezer:
:+1:

PR #66, opened 2016-01-14
Add a CLI helper

### LINKED ISSUE #32: Add ability to inspect remote system's state
It'd be nice if we could inspect the state of remote systems under test. For example, it may be useful to log in to a remote system and determine whether certain files are lying around in certain places in the file system.

@bmbouter and others may be able to provide more info about this issue.

Ichimonji10:
Add, document and test module `pulp_smash.cli`. This module provides "tools for
working with Pulp's CLI." This fixes #32:

> It'd be nice if we could inspect the state of remote systems under test. For
> example, it may be useful to log in to a remote system and determine whether
> certain files are lying around in certain places in the file system.

Here's a simple example of how to use the new module:

```python
>>> from pulp_smash import cli, config
>>> server_config = config.ServerConfig('localhost')
>>> client = cli.Client(server_config)
>>> response = client.run(('echo', '-n', 'foo'))
>>> response.returncode == 0
True
>>> response.stdout == 'foo'
True
>>> response.stderr == ''
True
```

The `Client` class is flexible enough that both request and response handling
can be customized on a per-instance basis. See the docstrings in module
`pulp_smash.cli` for details.

Add "plumbum" as a dependency. It is used by the CLI helper.

Many thanks to @elyezer for help in brainstorming a solution to this issue.

This commit does not affect test results (107 tests):

    ============  =============================
    ============  =============================
    2.7           OK
    dev (2.8)     FAILED (errors=2, skipped=13)
    ============  =============================

### DIFF
--- a/docs/api/pulp_smash.cli.rst
+++ b/docs/api/pulp_smash.cli.rst
@@ -1,0 +1,6 @@
+`pulp_smash.cli`
+================
+
+Location: :doc:`/index` → :doc:`/api` → :doc:`/api/pulp_smash.cli`
+
+.. automodule:: pulp_smash.cli
--- a/pulp_smash/__main__.py
+++ b/pulp_smash/__main__.py
@@ -20,15 +20,19 @@ MESSAGE = tuple((
         "base_url": "https://pulp.example.com",
         "auth": ["username", "password"],
         "verify": true,
-        "version": "2.7.5"
+        "version": "2.7.5",
+        "cli_transport": "local"
     }}''',
     '''\
-    The `verify` and `version` keys are completely optional. By default, Pulp
-    Smash respects SSL verification procedures, but the `verify` option can be
-    used to explicitly enable or disable SSL verification. By default, Pulp
-    Smash assumes that the Pulp server under test is the absolute latest
-    development version, but the `version` option can be used to explicitly run
-    tests suitable for an older version of Pulp.
+    The `verify`, `version` and `cli_transport` keys are optional. The `verify`
+    option can be used to explicitly enable or disable SSL verification. The
+    `version` option can be used to explicitly run tests suitable for an older
+    version of Pulp. By default, Pulp Smash assumes that the Pulp server under
+    test is the latest development version. The `cli_transport` key can be used
+    to explicitly choose how to contact the Pulp server when executing shell
+    commands. This can be set to "local" or "ssh". If omitted, Pulp Smash
+    guesses which transport to use by comparing the hostname in the `base_url`
+    against the current system's hostname.
     ''',
     '''\
     Notes:
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -1,0 +1,189 @@
+# coding=utf-8
+"""Tools for working with Pulp's CLI."""
+from __future__ import unicode_literals
+
+import socket
+import subprocess
+from sys import version_info
+try:  # try Python 3 import first
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse  # pylint:disable=C0411,E0401
+
+import plumbum
+
+
+def _get_hostname(urlstring):
+    """Get the hostname from a URL string.
+
+    ``urlparse`` follows RFC 1808 and requires that network locations be
+    prefixed with "//". This function is a thin wrapper. It treats the leading
+    "//" as optional::
+
+    >>> urls = ('ftp://localhost', '//localhost', 'localhost', 'localhost:123')
+    >>> for url in urls:
+    ...     print((_get_hostname(url), urlparse(url).hostname))
+    ('localhost', 'localhost')
+    ('localhost', 'localhost')
+    ('localhost', None)
+    ('localhost', None)
+
+    Usage::
+
+        if server_config is None:
+            server_config = get_config()
+        hostname = _get_hostname(server_config.base_url)
+
+    :param urlstring: A string such as "localhost:3000", "pulp.example.com" or
+        "https://pulp.example.com".
+    :returns: A hostname, such as "localhost" or "pulp.example.com".
+
+    """
+    parts = urlparse(urlstring)
+    if parts.hostname is None:
+        return _get_hostname('//' + parts.path)
+    else:
+        return parts.hostname
+
+
+def echo_handler(completed_proc):
+    """Immediately return ``completed_proc``."""
+    return completed_proc
+
+
+def code_handler(completed_proc):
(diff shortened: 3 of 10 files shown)


Ichimonji10:
@bmbouter

docs/api/pulp_smash.cli.rst
```diff
@@ -0,0 +1,6 @@
+`pulp_smash.cli`
+================
+
+Location: :doc:`/index` → :doc:`/api` → :doc:`/api/pulp_smash.cli`
```
elyezer: Breadcrumbs++

pulp_smash/__main__.py
```diff
@@ -20,15 +20,20 @@
...
+        "cli_transport": "local"
     }}''',
     '''\
-    The `verify` and `version` keys are completely optional. By default, Pulp
-    Smash respects SSL verification procedures, but the `verify` option can be
-    used to explicitly enable or disable SSL verification. By default, Pulp
-    Smash assumes that the Pulp server under test is the absolute latest
-    development version, but the `version` option can be used to explicitly run
-    tests suitable for an older version of Pulp.
+    The `verify`, `version` and `cli_transport` keys are completely optional.
```
bmbouter: s/completely //

pulp_smash/__main__.py
```diff
@@ -20,15 +20,20 @@
...
+    The `verify` option can be used to explicitly enable or disable SSL
+    verification. By default, it is enabled. The `version` option can be used
+    to explicitly run tests suitable for an older version of Pulp. By default,
+    Pulp Smash assumes that the Pulp server under test is the absolute latest
```
bmbouter: s/absolute //

pulp_smash/__main__.py
```diff
@@ -20,15 +20,20 @@
...
     '''\
-    The `verify` and `version` keys are completely optional. By default, Pulp
-    Smash respects SSL verification procedures, but the `verify` option can be
-    used to explicitly enable or disable SSL verification. By default, Pulp
-    Smash assumes that the Pulp server under test is the absolute latest
-    development version, but the `version` option can be used to explicitly run
-    tests suitable for an older version of Pulp.
+    The `verify`, `version` and `cli_transport` keys are completely optional.
+    The `verify` option can be used to explicitly enable or disable SSL
+    verification. By default, it is enabled. The `version` option can be used
```
bmbouter: s/By default, it is enabled. //

pulp_smash/__main__.py
```diff
@@ -20,15 +20,20 @@
...
+    to explicitly run tests suitable for an older version of Pulp. By default,
+    Pulp Smash assumes that the Pulp server under test is the absolute latest
+    development version. The `cli_transport` key can be used to explicitly
+    choose how to contact the Pulp server when executing shell commands. This
+    can be set to "local" or "ssh", and by default, Pulp Smash guesses which
```
bmbouter: s/, and by default,/ and defaults to "local"/

bmbouter: /Pulp Smash/If omitted, Pulp Smash/

pulp_smash/cli.py
```diff
@@ -0,0 +1,199 @@
...
+        str_kwargs = ', '.join([
+            'args={!r}'.format(self.args),
+            'returncode={!r}'.format(self.returncode),
+            'stdout={!r}'.format(self.stdout),
+            'stderr={!r}'.format(self.stderr),
+        ])
+        return '{}({})'.format(type(self).__name__, str_kwargs)
+
+    def check_returncode(self):
+        """Raise ``subprocess.CalledProcessError`` if exit code is non-zero."""
```
bmbouter: I like this behavior :+1:

pulp_smash/cli.py
```diff
@@ -0,0 +1,199 @@
...
+        # How do we make requests?
+        hostname = _get_hostname(server_config.base_url)
+        if server_config.cli_transport is None:
+            transport = 'local' if hostname == socket.gethostname() else 'ssh'
+        else:
+            transport = server_config.cli_transport
+        if transport == 'local':
+            self.machine = plumbum.machines.local
+        else:  # transport == 'ssh'
+            # Get SSH configuration options from an ssh_config file. See
```
bmbouter: I propose removing L#163 - L#173 altogether since this already does effectively use `ssh_config`. @elyezer and I were talking about this and he confirmed that if the value of hostname has an entry in `ssh_config` all of its ssh options will be used. We don't need to parse `ssh_config`. That is my understanding. @Ichimonji10 is that your understanding?

bmbouter:
Does this need any representation in the docs or do the docstrings create the docs? If not, I think the docstring examples can be moved into the docs or have the docs build from them so we don't duplicate them.

Ichimonji10:
The documentation is built directly from the docstrings.

pulp_smash/__main__.py
Ichimonji10: It doesn't actually default to "local". If `cli_transport` isn't set in the configuration file, this logic executes:

```python
transport = 'local' if hostname == socket.gethostname() else 'ssh'
```

pulp_smash/cli.py
Ichimonji10: Yes, it's correct that `SshMachine` will use `~/.ssh/config` if it is present. Consequently, I do need to edit this section. (My understanding of how Plumbum's `SshMachine` handled SSH config files was murky when I wrote this code.)

My train of thought here is that it'd be cool to configure the `SshMachine` a little bit. For example, it could be nice if we could tell the `SshMachine` to never interact with the user. This would prevent sad things like test sessions on a CI server hanging forever due to SSH key unlock prompts just hanging. It could also be nice to make the `SshMachine` look for an alternate config file. This would prevent Pulp Smash from knowing things it shouldn't (it shouldn't know the key file for logging into my home server or VMs at work), it would let me configure Pulp Smash differently from my regular SSH client (maybe I want Pulp Smash to use a different username, key and port).

But -- I don't see an easy way to do this, and we may be meeting users expectations best by just falling back to standard OpenSSH behaviours.

elyezer:
The travis failure is `./pulp_smash/cli.py:7:1: F401 'warnings' imported but unused`.

Ichimonji10:
Updated.

elyezer:
ACK

pulp_smash/__main__.py
bmbouter: @Ichimonji10 Oh yes, I see that now.

bmbouter:
This looks good to me. +1 to merging it.

Issue #55, opened 2015-12-16
Broker Initial Connect and Reconnect Test Plan


bmbouter:
Pulp offers a collection of behaviors known as "reconnect support" for the Pulp Broker. Here are the expected behaviors:

* If you start a Pulp service that connects to the broker and the broker is not running or is not network accessible for some reason, the Pulp services will wait-and-retry. It has a backoff behavior, but the important part is that it doesn't exit if it can't connect due to availability and when the availability problem is resolved it does connect.
* If you have a Pulp service connected to the broker and the broker shuts down, the pulp services need wait-and-retry as described above. Once the broker becomes available again the pulp services should reconnect.

The way I usually test this is as follows.

Test that "Pulp Wait and Retry Broker Initial Connect":
1. Stop the broker, `pulp_workers`, `pulp_celerybeat`, `pulp_resource_manager`, and `httpd`
2. Start `pulp_workers`, `pulp_celerybeat`, `pulp_resource_manager`, and `httpd`.
3. Wait 30 seconds
4. Start the broker
5. Smoke test using a basic sync/publish of rpm with Pulp.

Test "Pulp broker reconnect".
1. Start the broker, `pulp_workers`, `pulp_celerybeat`, `pulp_resource_manager`, and `httpd`
2. Stop the broker
3. Wait 30 seconds
4. Start the broker
5. Smoke test using a basic sync/publish of rpm with Pulp.

I delineate these as separate because it would be useful to know independently if either is working. We've had cases in the past where one was broken, or the other, or both. Note, "the broker" here is Qpid or rabbitMQ whichever the system under test is using.

PR #67, opened 2016-01-20
Add broker (re)connect tests

Ichimonji10:
This PR is comprised of two commits:

1. Add utilities for managing system services. These utilities execute shell commands like "systemctl start httpd" or "sudo service httpd stop," and they do so either locally or over SSH.
2. Add broker (re)connect tests.

I *just* ran the tests one final time as a sanity check. Each of the following commands run as expected on my system:

```sh
python -m unittest2 pulp_smash.tests.rpm.api_v2.test_broker
```

Related to #31 and #55.

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -13,6 +13,12 @@ except ImportError:
 import plumbum
 
 
+# A dict mapping hostnames to *nix service managers.
+#
+# For example: {'old.example.com': 'sysv', 'new.example.com', 'systemd'}
+_SERVICE_MANAGERS = {}
+
+
 def _get_hostname(urlstring):
     """Get the hostname from a URL string.
 
@@ -44,6 +50,10 @@ def _get_hostname(urlstring):
         return _get_hostname('//' + parts.path)
     else:
         return parts.hostname
+
+
+class NoKnownServiceManagerError(Exception):
+    """Indicates we cannot determine the service manager used by a system."""
 
 
 def echo_handler(completed_proc):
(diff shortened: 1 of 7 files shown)


Ichimonji10:
This branch is not yet ready to be merged. I'm submitting this pull request anyway to make work more visible.

Ichimonji10:
Relevant material:

* #31
* #55
* https://github.com/Ichimonji10/pulp-smash/commit/a056d1ffd7f8a4c5af996021940d128804ef5467

pulp_smash/utils.py
```diff
@@ -80,3 +82,38 @@ def poll_task(server_config, href):
...
+    * It violates the Liskov substitution principle, due to its ``__init__``
+      method requiring a more specific set of arguments.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        system on which an AMQP broker exists.
+    :rtype: pulp_smash.cli.Service
+    """
+    client = cli.Client(server_config, cli.echo_handler)
+    for executable in ('qpidd', 'rabbitmq'):  # ordering is important
+        file_check = ('[', '-f', '/usr/sbin/' + executable, ']')
```
Ichimonji10: `/etc/profile` is only executed for login shells. Fedora 23 adds `/usr/sbin` and `/usr/local/sbin` to `$PATH` in `/etc/profile`. As a result, `ssh fedora-23-host.example.com which qpidd` returns a non-zero exit code.

pulp_smash/cli.py
```diff
@@ -184,6 +190,97 @@ def run(self, args, **kwargs):
...
+                (('which', 'systemctl'), 'systemd'),
+                (('which', 'service'), 'sysv'),
+        ):
+            try:
+                client.run(command)
+                _SERVICE_MANAGERS[hostname] = svc_manager
+                return svc_manager
+            except subprocess.CalledProcessError:
+                pass
+        raise Exception  # FIXME: Raise a custom exception.
```
bmbouter: Out of curiosity, where are the coded exceptions kept for pulp smash?

Ichimonji10: There is no single `pulp_smash.exceptions` module or anything like it. They're sprinkled throughout the codebase. :/

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+
+Tests for `#55 <https://github.com/PulpQE/pulp-smash/issues/55>`_:
+
+> Pulp offers a collection of behaviors known as "reconnect support" for the
+> Pulp Broker. Here are the expected behaviors:
+>
+> * If you start a Pulp service that connects to the broker and the broker is
+>   not running or is not network accessible for some reason, the Pulp services
+>   will wait-and-retry. It has a backoff behavior, but the important part is
+>   that it doesn't exit if it can't connect due to availability and when the
```
bmbouter: I think I wrote this originally...
s/that it doesn't/that Pulp services don't/
s/if it/if they/
s/ and when/, and when/

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+>   availability problem is resolved it does connect.
```
bmbouter: s/resolved it does connect/resolved, the Pulp services reconnect/

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+> * If you have a Pulp service connected to the broker and the broker shuts
+>   down, the pulp services need wait-and-retry as described above. Once the
```
bmbouter: s/pulp/Pulp/
s/need /need the /

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+>   broker becomes available again the pulp services should reconnect.
```
bmbouter: s/pulp/Pulp/

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+
+There are two scenarios to test here: support for initially connecting to a
+broker, and support for reconnecting to a broker that goes missing. These two
```
bmbouter: s/broker,/broker/
Consider making these two cases bullet points.

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+           distributor, publish it, and download an RPM.
+        """
+        # Step 1 and 2.
+        for service in self.services + (self.broker,):
+            service.stop()
+        for service in self.services:
+            service.start()
+        time.sleep(15)  # Let services try to connect to the dead broker.
+        self.broker.start()
+        self.health_check()
```
bmbouter: This should have a comment saying step 3

pulp_smash/tests/rpm/api_v2/broker_reconnect.py
```diff
@@ -0,0 +1,147 @@
...
+        2. Stop the broker, wait, and start it again.
+        3. Test Pulp's health. Create an RPM repository, sync it, add a
+           distributor, publish it, and download an RPM.
+        """
+        # We assume that the broker and other services are already running. As
+        # a result, we skip step 1 and go straight to step 2.
+        self.broker.stop()
+        time.sleep(30)
+        self.broker.start()
+        self.health_check()
```
bmbouter: step 3 comment here also

pulp_smash/utils.py
```diff
@@ -80,3 +82,38 @@ def poll_task(server_config, href):
...
+
+    Talk to the host named by ``server_config`` and use simple heuristics to
+    determine which AMQP broker is installed. If Qpid or RabbitMQ appear to be
+    installed, return a :class:`pulp_smash.cli.Service` object for managing
+    those services respectively. Otherwise, raise an exception.
+
+    Another option for solving this problem is to create a subclass of
+    ``Service``, like so:
+
+    >>> class Broker(Service):
```
bmbouter: +1 to doing it like this anyway :smile_cat:

bmbouter: Consider removing these alternate design docs, they lower the actual docs information density considerable. These would be good on either the issue itself or in the PR of review. Keeping them as a permanent part of the code's docstrings is probably not the best place for it.

bmbouter: I'm ok in not switching this to a class, but I do want to identify that I prefer that approach. I also recognize that my preference does not dictate the design.

pulp_smash/cli.py
```diff
@@ -184,6 +194,99 @@ def run(self, args, **kwargs):
...
+    'systemctl start httpd'
+    """
+
+    def __init__(self, server_config, service):
+        """Initialize a new object."""
+        self._client = Client(server_config)
+        self._command_builder = None
+
+        # Set `self._command_builder`.
+        # FIXME: Conditionally use sudo. Explain when sudo is used.
```
bmbouter: +1 to implementing the sudo check machinery as internal methods on this class.

pulp_smash/utils.py
```diff
@@ -19,6 +21,10 @@ class TaskTimedOutException(Exception):
     """Indicates that polling a task timed out."""
 
 
+class NoKnownBrokerError(Exception):
```
bmbouter: Consider having a designated module for all custom exceptions.

pulp_smash/utils.py
Ichimonji10: > Consider removing these alternate design docs, they lower the actual docs information density considerable. These would be good on either the issue itself or in the PR of review. Keeping them as a permanent part of the code's docstrings is probably not the best place for it.

Definitely. I'll remove this info from the codebase, add these thoughts to the relevant Pulp Smash issues, and add some one-line comments pointing users to the relevant issues.

pulp_smash/utils.py
Ichimonji10: Sure. I've seen it done in other well-respected projects. Doing so should also make life easier for end users should a breaking change need to be made; the number of references that need to change will be minimized if all exceptions live in a single location.

pulp_smash/utils.py
```diff
@@ -80,3 +86,48 @@ def poll_task(server_config, href):
...
+    # On Fedora 23, /usr/sbin and /usr/local/sbin are only added to the $PATH
+    # for login shells. (See /etc/profile.) As a result, logging into a system
+    # and executing `which qpidd` and remotely executing `ssh pulp.example.com
+    # which qpidd` may return different results. For that reason, we try both
+    # checks here.
+    client = cli.Client(server_config, cli.echo_handler)
+    executables = ('qpidd', 'rabbitmq')  # ordering indicates preference
+    for executable in executables:
+        file_check = ('[', '-f', '/usr/sbin/' + executable, ']')
+        if (client.run(('which', executable)).returncode == 0 or
```
bmbouter: There is very little chance that the `qpidd` and `rabbitmq` binaries will move. Consider using:
`test -e /usr/sbin/qpidd` as a simple way to do this. I'm not sure `which` is providing much benefit here and it's introducing complexity.

pulp_smash/utils.py
bmbouter: That sounds good. We used to put links in our code, then one day we switched issue trackers. I'm not sure to recommend to guard against that. What you propose sounds good.

pulp_smash/utils.py
Ichimonji10: I do like using `which` to locate executables. It's easy to use, widely available, and provides compatibility with distributions that don't place executables in `/usr/sbin` and `/usr/local/sbin`. That said, Pulp is compatible with a small number of *nix distributions, so we know approximately what our target systems will be like. (This is in contrast to highly software that might be packaged for use on who-knows-how-many different distributions. And even a distro like Arch Linux, as forward-looking as it is, symlinks `/bin`, `/lib`, `/lib64`, `/sbin`, etc to the appropriate destinations.) And code paths that are not traversed or are never expected to succeed are a liability. So - sure, this makes sense. I'll drop this, and we can add back in a `which "$executable"` check if and when it actually is needed.

pulp_smash/utils.py
Ichimonji10: See #68.

Ichimonji10:
This PR is in a pretty good spot. If anyone feels like reviewing the changes and giving feedback, now's a good time.

bmbouter:
I just reviewed this, and it looks good to merge! Great job @Ichimonji10 and @elyezer 

Testing this over and over has been a challenge for Pulp's stability. Usually it breaks due to a bug in a dependency, so it can break at any time with Pulp not even changing. So great!

PR #69, opened 2016-01-21
Skip tests affected by Pulp 1440

Ichimonji10:
Skip known-failing test code affected by Pulp issue 1440. [1] This issue affects
`pulp_smash.tests.puppet.api_v2.test_sync_publish.PublishTestCase`. Test results
before this commit (21 and 11 tests, respectively):

    ============  =============================
    ============  =============================
    2.7           OK
    dev (2.8)     FAILED (errors=1)
    ============  =============================

Test results after this commit (21 tests):

    2.7           OK
    dev (2.8)     OK

Test results produced locally with the following command:

    python -m unittest2 pulp_smash.tests.puppet.api_v2.test_sync_publish

[1] https://pulp.plan.io/issues/1440

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -32,6 +32,7 @@ developers, not a gospel.
     api/pulp_smash.tests.puppet.api_v2.test_sync_publish
     api/pulp_smash.tests.rpm
     api/pulp_smash.tests.rpm.api_v2
+    api/pulp_smash.tests.rpm.api_v2.test_broker
     api/pulp_smash.tests.rpm.api_v2.test_iso_crud
     api/pulp_smash.tests.rpm.api_v2.test_sync_publish
     api/pulp_smash.utils
--- a/docs/api/pulp_smash.tests.rpm.api_v2.test_broker.rst
+++ b/docs/api/pulp_smash.tests.rpm.api_v2.test_broker.rst
@@ -1,0 +1,7 @@
+`pulp_smash.tests.rpm.api_v2.test_broker`
+=========================================
+
+Location: :doc:`/index` → :doc:`/api` →
+:doc:`/api/pulp_smash.tests.rpm.api_v2.test_broker`
+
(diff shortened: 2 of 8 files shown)


Ichimonji10:
This should fix ten test failures currently appearing in Jenkins.

asmacdo:
Cool! I know this PR isn't adding this feature, but this is the first time I've seen it. Does this mean that pulp-smash is making a REST request for every bug that has this decorator every time pulp-smash is run? Does this mean that pulp-smash cannot be run without the internet for instance?

Ichimonji10:
Yes, Pulp Smash does talk to pulp.plan.io each time `pulp_smash.selectors.bug_is_testable` or `pulp_smash.selectors.bug_is_untestable` is called. These functions make use of a cache and force the cache to have homogenous keys:

```python
    # It's rarely a good idea to do type checking in a duck-typed language.
    # However, efficiency dictates we do so here. Without this type check, the
    # following will cause us to talk to the bug tracker twice and store two
    # values in the cache:
    #
    #     _get_bug_status(1356)
    #     _get_bug_status('1356')
    #
    if not isinstance(bug_id, int):
        raise TypeError(
            'Bug IDs should be integers. The given ID, {} is a {}.'
            .format(bug_id, type(bug_id))
        )
    try:
        return _BUG_STATUS_CACHE[bug_id]
    except KeyError:
        pass
```

Pulp Smash can be run without internet access. If no bug tracker is available, the `bug_is_testable` and `bug_is_untestable` functions will assume that the bug at hand is testable:

```python
    try:
        status = _get_bug_status(bug_id)
    except requests.exceptions.ConnectionError as err:
        message = (
            'Cannot contact the bug tracker. Pulp Smash will assume that the '
            'bug referenced is testable. Error: {}'.format(err)
        )
        warnings.warn(message, RuntimeWarning)
        return True
```

PR #77, opened 2016-02-01
RPM tests bugfix

peterlacko:
SyncValidFeedTestCase and SyncInvalidFeedTestCase didn't delete
    repositories after completition.
    Fixed by adding repositories' paths to resources for deletion.

    This commit does not affect functionality of pulp-smash.


Ichimonji10:
I can confirm that test results do not change as a result of this commit.

Ichimonji10:
Thank you.

PR #80, opened 2016-02-03
Adds a _CleanUpTestCaseMixin

dkliban:
This mixin provides two methods. One for resetting the database. The second
to clean up /var/lib/pulp/content and /var/lib/pulp/published directories.

### DIFF
--- a/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
@@ -43,7 +43,7 @@ except ImportError:
 import unittest2
 from packaging.version import Version
 
-from pulp_smash import api, config, selectors, utils
+from pulp_smash import api, cli, config, selectors, utils
 from pulp_smash.constants import (
     CALL_REPORT_KEYS,
     CONTENT_UPLOAD_PATH,
@@ -66,6 +66,12 @@ _PUPPET_MODULE_URL = (
 )
 _PUPPET_QUERY = _PUPPET_MODULE['author'] + '-' + _PUPPET_MODULE['name']
 
+_SERVICES = {
+    'httpd',
+    'pulp_celerybeat',
+    'pulp_resource_manager',
+    'pulp_workers',
+}
 
 def _gen_repo():
     """Return a semi-random dict that used for creating a puppet repo."""
@@ -75,6 +81,40 @@ def _gen_repo():
         'importer_type_id': 'puppet_importer',
         'notes': {'_repo-type': 'puppet-repo'},
     }
+
+class _CleanUpTestCaseMixin(object):
+    """Provide functions to clean up database and filesystem of Pulp"""
+
+    @classmethod
+    def reset_database(cls, server_config):
+        """Drop database and re-run migrations."""
+        cls.services = tuple((
+            cli.Service(cls.cfg, service) for service in _SERVICES
+        ))
+
+        # Stop all Pulp services
+        for service in cls.services:
+            service.stop()
+
+        # Drop database and re-run migrations
+        client = cli.Client(server_config, cli.echo_handler)
+        drop_db_command = ('mongo', 'pulp_database', '--eval', 'db.dropDatabase()')
+        client.run(drop_db_command)
+        migrations_command = ('sudo', '-u', 'apache', 'pulp-manage-db')
+        client.run(migrations_command)
+
+        # Restart all Pulp services
+        for service in cls.services:
+            service.start()
+
+    @classmethod
+    def clean_filesystem(cls, server_config):
+        """Remove all content from /var/lib/pulp/[content, published]"""
+        client = cli.Client(server_config, cli.echo_handler)
+        content_command = ('rm', '-rf', '/var/lib/pulp/content/*')
+        published_command = ('rm', '-rf', '/var/lib/pulp/published/*')
+        client.run(content_command)
+        client.run(published_command)
 
 
 class _BaseTestCase(unittest2.TestCase):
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -76,6 +82,40 @@ def _gen_repo():
         'notes': {'_repo-type': 'puppet-repo'},
     }
 
+class _CleanUpTestCaseMixin(object):
+    """Provide functions to clean up database and filesystem of Pulp"""
+
+    @classmethod
+    def reset_database(cls, server_config):
+        """Drop database and re-run migrations."""
+        cls.services = tuple((
```
Ichimonji10: Why make this a class attribute? I don't see any advantage in filling client classes' namespaces further.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -76,6 +82,40 @@ def _gen_repo():
...
+        # Stop all Pulp services
+        for service in cls.services:
+            service.stop()
+
+        # Drop database and re-run migrations
+        client = cli.Client(server_config, cli.echo_handler)
+        drop_db_command = ('mongo', 'pulp_database', '--eval', 'db.dropDatabase()')
+        client.run(drop_db_command)
+        migrations_command = ('sudo', '-u', 'apache', 'pulp-manage-db')
+        client.run(migrations_command)
```
Ichimonji10: This can also be written as:

```python
        client.run("mongo pulp_database --eval 'db.dropDatabase()'".split())
        client.run('sudo -u apache pulp-manage-db')
```

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -76,6 +82,40 @@ def _gen_repo():
...
+        cls.services = tuple((
+            cli.Service(cls.cfg, service) for service in _SERVICES
+        ))
+
+        # Stop all Pulp services
+        for service in cls.services:
+            service.stop()
+
+        # Drop database and re-run migrations
+        client = cli.Client(server_config, cli.echo_handler)
```
Ichimonji10: Why use the echo handler? If we encounter an error when resetting the database, we probably want to know about it. Leaving this as the default `code_handler` makes more sense to me.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -76,6 +82,40 @@ def _gen_repo():
         'notes': {'_repo-type': 'puppet-repo'},
     }
 
+class _CleanUpTestCaseMixin(object):
+    """Provide functions to clean up database and filesystem of Pulp"""
+
+    @classmethod
+    def reset_database(cls, server_config):
```
Ichimonji10: This method references both `cls.cfg` and `server_config`. How about turning this into a static method?

```python
@staticmethod
def reset_database(server_config):
    ...
```

This makes it more obvious what's going on.

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -66,6 +66,12 @@
 )
 _PUPPET_QUERY = _PUPPET_MODULE['author'] + '-' + _PUPPET_MODULE['name']
 
+_SERVICES = {
+    'httpd',
+    'pulp_celerybeat',
+    'pulp_resource_manager',
+    'pulp_workers',
+}
```
Ichimonji10: Hooray for set comprehensions!

pulp_smash/tests/puppet/api_v2/test_sync_publish.py
```diff
@@ -76,6 +82,40 @@ def _gen_repo():
...
+            service.start()
+
+    @classmethod
+    def clean_filesystem(cls, server_config):
+        """Remove all content from /var/lib/pulp/[content, published]"""
+        client = cli.Client(server_config, cli.echo_handler)
+        content_command = ('rm', '-rf', '/var/lib/pulp/content/*')
+        published_command = ('rm', '-rf', '/var/lib/pulp/published/*')
+        client.run(content_command)
+        client.run(published_command)
```
Ichimonji10: This can also be written as:

```python
client.run('rm -rf /var/lib/pulp/content/*'.split())
client.run('rm -rf /var/lib/pulp/published/*'.split())
```

Ichimonji10:
Is there any use case for separately resetting the database and nuking
`/var/lib/pulp/{content,published}/*`, either right now or in the immediate
future? I can't think of any. If the goal is to reset Pulp, one will almost
certainly want to do these two actions together. Doing just one action will
probably put Pulp in a bad state. How about combining the `reset_database` and
`clean_filesystem` methods into a single `reset_pulp` method?

Is there any benefit to providing this functionality as a mixin? At this point,
I don't see any. Resetting Pulp doesn't require access to any instance or class
attributes, and currently, we explicitly call the `reset_database` and
`clean_filesystem` methods. On the other hand, inheritance bears significant
legibility costs. Developers reading the code at hand must traverse the entire
inheritance tree for any given class to understand how it works. Mixins also
fill the namespaces of the child classes, making accidental method overrides
more likely.

How about providing a single `pulp_smash.utils.reset_pulp` function?

```python
def reset_pulp(server_config):
    """Reset Pulp's database, nuke accumulated files, and restart services."""
    services = tuple((
        cli.Service(server_config, service) for service in
        {'httpd', 'pulp_celerybeat', 'pulp_resource_manager', 'pulp_workers'}
    ))
    for service in services:
        service.stop()

    # Reset the database.
    client.run("mongo pulp_database --eval 'db.dropDatabase()'".split())
    client.run('sudo -u apache pulp-manage-db'.split())

    # Nuke accumulated files.
    client.run('rm -rf /var/lib/pulp/content/*'.split())
    client.run('rm -rf /var/lib/pulp/published/*'.split())

    for service in services:
        service.start()
```

Ichimonji10:
See: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:reset-pulp

Ichimonji10:
See #83.

PR #83, opened 2016-02-03
Reset pulp

Ichimonji10:
Add a function for resetting pulp in the first commit, and make use of that function to twiddle with the tests in the second commit.


Ichimonji10:
@dkliban @asmacdo I've got to head to the doctor's, and may not be back later today. I've tested these changes locally, but you may with to review them too. I'm in a bit of a rush and may have missed something.

Ichimonji10:
@asmacdo @dkliban Thank you both!

Issue #84, opened 2016-02-04
pulp-smash should only run tests for plugins that are actually installed


dkliban:
Pulp smash should inspect the system under test to determine which plugins are installed. Based on that information some or all tests should be run.

Ichimonji10:
:+1:

What do you think is a good, canonical way of determining which plugins are installed? I of course like the idea of checking with a package manager. But when installing using the scripts shipped with Pulp, which is what happens when one executes `vagrant up`, the package manager may be left ignorant of that. Can you shed some light on the interactions between the package manager and the installation scripts, @dkliban?

omaciel:
Robotello uses a `skip_if_not_set` (or something similar) that will skip a test or tests if certain configurations are not available. For example, if there's no Docker installed locally, then docker related test are skipped.

I'm in favor of making tests as smart as possible, but let's be careful not to make them too smart :)  Folks will only look at test results if something goes wrong and if we skip tests without some type of vigilance, we run the risk of ignoring the testing of certain areas.

Ichimonji10:
The Pulp Smash code is organized in such a way that it's easy to select which plugins you'd like to test. Can we tell the test runner to blacklist certain packages (I mean a "package" in the Pythonic sense, not the package manager sense) during test discovery?

Ichimonji10:
@asmacdo or anyone else: is there some good, canonical way to determine whether a given plugin is installed? One option is to execute `test -d /path/to/plugin/dir` on the host. That's ugly and relies heavily on implementation-specific details, but it works. Another option is to ask pulp-admin on the target system, though that brings up its own set of possible bugs. Thoughts? See: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:ostree

Ichimonji10:
Is there, say, an API call I can make where I ask Pulp which plugins are installed and it responds with a list of them?

Pulp is only available through its API. Even the pulp-admin CLI client talks to Pulp via its API. And this is appropriate: it means that the target host can install Pulp however it likes (system-wide, per-use, in a virtualenv, etc), and client don't have to care. If we can maintain that level of abstraction, it could be nice.

[...]

PR #96, opened 2016-02-08
Conditionally execute OSTree tests

Ichimonji10:
Remove the unconditional `skipTest` exception raised by function `setUpModule` in module `pulp_smash.tests.ostree.api_v2.test_sync_publish`.

Add a `load_tests` function [1] to package `pulp_smash.tests.ostree`. This function talks to the Pulp server under test, asks it for a list of content unit types, and uses that information in choosing whether to run the OSTree tests.

This works because each plugin adds one (or more?) content unit type to Pulp. The types added by a given plugin are listed in the documentation for those plugins. For example, see:

* http://pulp-ostree.readthedocs.org/en/latest/tech-reference/type.html
* http://pulp-python.readthedocs.org/en/latest/reference/python-type.html

Due to Pulp issue 1642 [2], this only works well on Pulp 2.7. On Pulp 2.8, the OSTree plugin does not appear to be installed (even if present), so these tests are skipped.

Before this commit, all OSTree tests are skipped. Test results after this commit:

    2.7        18          OK
    dev (2.8)  0           OK

Test results after this commit against systems without the OSTree plugin installed:

    2.7        0           OK
    dev (2.8)  0           OK

[1] https://docs.python.org/3.5/library/unittest.html#load-tests-protocol
[2] https://pulp.plan.io/issues/1642

Related to #84

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -57,6 +57,13 @@ LOGIN_PATH = '/pulp/api/v2/actions/login/'
     https://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/authentication.html
 """
 
+PLUGIN_TYPES_PATH = '/pulp/api/v2/plugins/types/'
+"""See: `Retrieve All Content Unit Types`_.
+
+.. _Retrieve All Content Unit Types:
+   http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/server_plugins.html#retrieve-all-content-unit-types
+"""
+
 PULP_SERVICES = {
     'httpd',
     'pulp_celerybeat',
--- a/pulp_smash/tests/ostree/__init__.py
+++ b/pulp_smash/tests/ostree/__init__.py
@@ -1,3 +1,43 @@
 # coding=utf-8
 """Functional tests for Pulp's ostree plugin."""
 from __future__ import unicode_literals
+
+import os
+
+from pulp_smash import api, config
+from pulp_smash.constants import PLUGIN_TYPES_PATH
+
+
+def _get_plugin_type_ids():
+    """Get the ID of each of Pulp's plugins.
+
+    Each Pulp plugin adds one (or more?) content unit type to Pulp. Each of
+    these content unit types is identified by a certain unique identifier. For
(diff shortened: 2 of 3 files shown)


Ichimonji10:
@dkliban Thoughts?

Ichimonji10:
@asmacdo This pull request makes use of unittest's `load_tests` protocol. It's powerful, and its usage can be simple. I think the approach shown in this pull request can be adapted to to conditionally select all of Pulp Smash's plugin tests. Check it out if you have time.

dkliban:
We should file an issue against Pulp with regard to OSTree plugin not showing that it's content type is registered with Pulp 2.8.

Ichimonji10:
Aye, that'd be nice. https://pulp.plan.io/issues/1642 describes how almost none of the plugins register their content types with Pulp, but it'd be nice if we could have one issue for each of the misbehaving plugins and make https://pulp.plan.io/issues/1642 point to them.

Ichimonji10:
This pull request works, but with two caveats:

* Pulp 2.8 does not correctly report which plugins are installed, so these tests never run on that platform.
* These tests sometimes take an exceptionally long time to run. In my tests, the entire collection of OSTree tests typically executes in ~25 seconds, but on occasion, they'll take ~5 minutes to execute. Yikes. I don't know what's up with these spikes in execution time.

I'm going to merge these. #102 describes an improvement.

Ichimonji10:
The test results described in the commit message still hold true. Here's test results against Pulp 2.7:

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ PULP_SMASH_CONFIG_FILE=${PULP}.json python -m unittest2 discover pulp_smash.tests
s...s.....s....................ss...s...sssss.s.........ssss.................................................s........s...................s....
Ran 138 tests in 651.986s

OK (skipped=19)
```

Here's test results against Pulp 2.8:

```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ PULP_SMASH_CONFIG_FILE=${PULP}.json python -m unittest2 discover pulp_smash.tests
.......s...............s....ssss...........ssss.................................................s............................s....
Ran 127 tests in 734.564s

OK (skipped=12)
```

PR #97, opened 2016-02-09
Isolates the RPM publish test case by resetting Pulp during setUp

dkliban:
We are seeing a failure during the nightly Jenkins run of pulp-smash. My investigation led me to isolating this test.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -280,6 +280,7 @@ class PublishTestCase(_BaseTestCase):
            RPMs from both repositories.
         """
         super(PublishTestCase, cls).setUpClass()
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Do you have any evidence that this commit fixes the test failure? I'm provisioning fresh VMs right now and can test this commit afterwards, but if you have something in hand already, it'd be nice to see that.

The test that's currently failing is doing so with an HTTP 404 error code, but the linked-to issue ([1406](https://pulp.plan.io/issues/1406)) describes an HTTP 500 error. Any idea what accounts for the discrepancy?

dkliban:
....s.......s....ssss...........ssss.................................................s............................s....
Ran 115 tests in 245.377s

OK (skipped=12)

dkliban:
The output in previous comment is from after I applied the patch.

Ichimonji10:
Merged as 77f50a400c690d0eef594b3346816fecb1311413

@dkliban, I re-wrote the commit message.

Issue #90, opened 2016-02-05
Repository create and sync with --remove-missing 


preethit:
Test to verify that repositories can be created and synced when using --remove-missing. Assert that the --remove-missing is functioning as expected
https://pulp.plan.io/issues/1621

------
Will come back and add steps

preethit:
I see 2 options ( There may be 1 more) how we can go
Option 1
1. Create a repo  foowith a feed to a repo  bar on another pulp-server 
pulp-admin rpm repo create --repo-id foo --feed http://pulp-server1/bar --remove-missing true
Assert
 The repo create is successful with the options

2. Synchronize the repo
pulp-admin rpm repo sync run --repo-id foo

3. On  remove an rpm from the bar repo
pulp-admin rpm repo remove rpm --repo-id bar  --str-eq="name=dog"
Assert: rpm dog is removed from the repo 

4. Publish repo "bar"
pulp-admin rpm repo publish run --repo-id bar

5. Synchronise the foo repo
Assert
 1. Sync is successful
 2.The rpm that is removed from the parent repo is removed from the local repo as well

peterlacko:
@preethit @Ichimonji10, is there something more that needs to be tested? comment please. To PR I'll add also conditional execution of tests on 2.8, as they are failing currently.

Ichimonji10:
@peterlacko, there's no need to conditionally execute tests on Pulp 2.8. See:

* https://github.com/PulpQE/pulp-smash/pull/97
* https://github.com/PulpQE/pulp-smash/commit/77f50a400c690d0eef594b3346816fecb1311413

Ichimonji10:
> is there something more that needs to be tested? comment please.

There are not any tests that make use of the `--remove-missing` pulp-admin flag or its equivalent in the API. That's a good place to start.

omaciel:
@peterlacko hi there, what's the status for this issue?

Ichimonji10:
This is tested by module `pulp_smash.tests.rpm.api_v2.test_remove_unit`.

Issue #24, opened 2015-10-29
Fiddle with multi-line docstrings


Ichimonji10:
From https://github.com/PulpQE/pulp-smash/pull/20#discussion_r43385240:

> the recommended way for multiline docstring is like example below ( see https://www.python.org/dev/peps/pep-0008/#documentation-strings for more information):
>
> ```python
> """Return a foobang
>
> Optional plotz says to frobnicate the bizbaz first.
> """
> ```
>
> In other words, the closing `"""` should stay on its own line and don't need a blank line before it.

This makes sense. Let's implement this new format throughout the application. Making this change has value because it'll make the code base more palatable to other developers who are familiar with PEP8 (and PEP257).

elyezer:
:+1:

PR #103, opened 2016-02-10
Test the functionality of syncing RPM repos with remove-unit option

peterlacko:
Test the remove-unit option in RPM repositories (#90).

Following steps are executed in order to test correct functionality
of repository created with valid feed and remove_missing option set.

    1. Create repository foo with valid feed, run sync, add distributor to it
       and publish over http and https.
    2. Create second repository, bar, with feed pointing to first repository,
       set remove_missing=True and run sync.
    3. Assert that repositories contain same units.
    4. Remove random unit from repository foo and publish.
    5. Sync bar repository
    6. Assert that foo and bar contain same units

This commit adds 4 new tests to pulp-smash suite and does not affect
functionality of other tests.
Test results before commit:

    2.7        134         FAIL (ERRORS=1)
    dev (2.8)  127         OK

Test suite results after this commit:

    2.7        138         FAIL (ERRORS=1)
    dev (2.8)  131         OK

Error on 2.7 is due to internal server error in
pulp_smash.tests.rpm.api_v2.test_iso_crud.AddImporterDistributorTestCase.
Needs some more investigation.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_remove_unit.py
+++ b/pulp_smash/tests/rpm/api_v2/test_remove_unit.py
@@ -1,0 +1,205 @@
+# coding utf-8
+"""Test the functionality in RPM repos when `remove_missing`_ is set to True.
+
+Following steps are executed in order to test correct functionality
+of repository created with valid feed and remove_missing option set.
+
+1. Create repository foo with valid feed, run sync, add distributor to it
+   and publish over http and https.
+2. Create second repository bar, with feed pointing to first repository,
+   set remove_missing=True and run sync on them.
+3. Assert that repositories contain same set of units.
+4. Remove random unit from repository foo and publish.
+5. Sync bar repository.
+6. Assert that both repositories contain same units.
+
+.. _remove missing:
+    https://pulp-rpm.readthedocs.org/en/latest/tech-reference/yum-plugins.html
+"""
+
+from __future__ import unicode_literals
+
+import random
+
+try:  # try Python 3 import first
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin  # pylint:disable=C0411,E0401
+
+
+from pulp_smash import api, utils
+from pulp_smash.constants import (
+    REPOSITORY_PATH,
+    RPM_FEED_URL,
+)
+
+_PUBLISH_DIR = 'pulp/repos/'
+
+
+def _gen_repo():
+    """Return a semi-random dict for use in creating an RPM repostirory."""
+    return {
+        'id': utils.uuid4(),
+        'importer_config': {},
+        'importer_type_id': 'yum_importer',
+        'notes': {'_repo-type': 'rpm-repo'},
+    }
+
+
+def _gen_distributor():
+    """Return a semi-random dict for use in creating a YUM distributor."""
+    return {
+        'auto_publish': False,
+        'distributor_id': utils.uuid4(),
+        'distributor_type_id': 'yum_distributor',
+        'distributor_config': {
+            'http': True,
+            'https': True,
+            'relative_url': utils.uuid4() + '/',
+        },
+    }
+
+
+class RemoveMissingTestCase(utils.BaseAPITestCase):
+    """Test functionality of --remove-missing option enabled."""
+
+    @classmethod
+    def setUpClass(cls):  # pylint:disable=arguments-differ
+        """Create two repositories, first is feed of second one.
+
+        Provides server config and set of iterable to delete.
+        Following steps are executed:
+            1. Create repository foo with feed, sync and publish it.
+            2. Create repository bar with foo as a feed and run sync.
+            3. Get content of both repositories.
+            4. Remove random unit from repository foo and publish foo.
+            5. Sync repository bar.
+            6. Get content of both repositories.
+        """
+        super(RemoveMissingTestCase, cls).setUpClass()
+
+        cls.responses = {}
+        client = api.Client(cls.cfg, api.safe_handler)
+
+        bodies = tuple((_gen_repo() for _ in range(2)))
+        bodies[0]['importer_config']['feed'] = RPM_FEED_URL
+        repos = []
+        repos.append(client.post(REPOSITORY_PATH, bodies[0]).json())
+        sync_path = urljoin(repos[0]['_href'], 'actions/sync/')
+        # Run sync and wait for the task to complete
+        cls.responses['sync'] = []
+        cls.responses['sync'].append(client.post(
+            sync_path, {'override_config': {}}
+        ))
+        # Add distributor and publish
+        cls.responses['distribute'] = client.post(
+            urljoin(repos[0]['_href'], 'distributors/'),
+            _gen_distributor(),
+        )
+        cls.responses['publish'] = []
+        cls.responses['publish'].append(client.post(
+            urljoin(repos[0]['_href'], 'actions/publish/'),
+            {'id': cls.responses['distribute'].json()['id']},
+        ))
+
+        # Use http feed instead of https to avoid possible config problems
+        bodies[1]['importer_config']['feed'] = urljoin(
+            # Create http url from base_url
+            cls.cfg.base_url,
+            _PUBLISH_DIR +
+            cls.responses['distribute'].json()['config']['relative_url'],
+        )
+        bodies[1]['importer_config']['remove_missing'] = True
+        # Turning off validation is neccessary, as each https feed requires
+        # certificate to be specified
+        bodies[1]['importer_config']['ssl_validation'] = False
+        # Create and sync second repo
+        repos.append(client.post(REPOSITORY_PATH, bodies[1]).json())
+        sync_path = urljoin(repos[1]['_href'], 'actions/sync/')
+        cls.responses['sync'].append(client.post(
+            sync_path, {'override_config': {}}
+        ))
+        # Get content of both repositories
+        body = {'criteria': {}}
+        cls.responses['units before removal'] = [
+            client.post(urljoin(repo['_href'], 'search/units/'), body)
+            for repo in repos
+        ]
+        # Get random unit from first repository to remove
+        rpms = [unit['metadata']['name']
+                for unit in cls.responses['units before removal'][0].json()
+                if unit['unit_type_id'] == 'rpm']
+        cls.removed_unit = random.choice(rpms)
+        # Remove unit from first repo and publish again
+        cls.responses['remove unit'] = client.post(
+            urljoin(repos[0]['_href'], 'actions/unassociate/'),
+            {'criteria':
+                {'fields':  # pylint:disable=bad-continuation
+                    {'unit': ['name', 'epoch', 'version', 'release',  # noqa pylint:disable=bad-continuation,line-too-long
+                              'arch', 'checksum', 'checksumtype']},
+                    'type_ids': ['rpm'],  # noqa pylint:disable=bad-continuation,line-too-long
+                    'filters': {'unit': {'name': cls.removed_unit}}}},)  # noqa pylint:disable=bad-continuation,line-too-long
+        # Publish first repo again
+        cls.responses['publish'].append(client.post(
+            urljoin(repos[0]['_href'], 'actions/publish/'),
+            {'id': cls.responses['distribute'].json()['id']},
+        ))
+        # Sync second repo
+        sync_path = urljoin(repos[1]['_href'], 'actions/sync/')
+        cls.responses['sync'].append(client.post(
+            sync_path, {'override_config': {}}
+        ))
+        # Search for units in both repositories again
+        cls.responses['units after removal'] = [
+            client.post(urljoin(repo['_href'], 'search/units/'), body)
+            for repo in repos
+        ]
+        for repo in repos:
+            cls.resources.add(repo['_href'])
+
+    def test_status_code(self):
+        """Verify th HTTP status code of each server response."""
+        for step, code in (
+                ('sync', 202),
+                ('publish', 202),
+                ('units before removal', 200),
+                ('units after removal', 200),
+        ):
+            with self.subTest(step=step):
+                for response in self.responses[step]:
+                    self.assertEqual(response.status_code, code)
+        for step, code in (
+                ('distribute', 201),
+                ('remove unit', 202),
+        ):
+            with self.subTest(step=step):
+                self.assertEqual(self.responses[step].status_code, code)
+
+    def test_units_before_removal(self):
+        """Test that units in repositories before removal are the same."""
+        bodies = [re.json() for re in self.responses['units before removal']]
+        # Package category and package group will differ so we count only RPMs
+        self.assertEqual(
+            set(unit['unit_id'] for unit in bodies[0]
+                if unit['unit_type_id'] == 'rpm'),  # This test is fragile
+            set(unit['unit_id'] for unit in bodies[1]
+                if unit['unit_type_id'] == 'rpm'),  # due to hard-coded
+        )  # indices. But the data is complex, and this makes things simpler.
+
+    def test_unit_removed(self):
+        """Test that correct unit from first repository has been removed."""
+        body = self.responses['units after removal'][0].json()
+        units_names = set(unit['metadata']['name'] for unit in body
+                          if unit['unit_type_id'] == 'rpm')
+        self.assertNotIn(self.removed_unit, units_names)
+
+    def test_units_after_removal(self):
+        """Test that units in repositories after removal are the same."""
+        bodies = [re.json() for re in self.responses['units after removal']]
+        # Package category and package group will differ so we count only RPMs
+        self.assertEqual(
+            set(unit['unit_id'] for unit in bodies[0]
+                if unit['unit_type_id'] == 'rpm'),  # This test is fragile
+            set(unit['unit_id'] for unit in bodies[1]
+                if unit['unit_type_id'] == 'rpm'),  # due to hard-coded
+        )  # indices. But the data is complex, and this makes things simpler.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -41,6 +41,7 @@ developers, not a gospel.
     api/pulp_smash.tests.rpm.api_v2
     api/pulp_smash.tests.rpm.api_v2.test_broker
     api/pulp_smash.tests.rpm.api_v2.test_iso_crud
+    api/pulp_smash.tests.rpm.api_v2.test_remove_unit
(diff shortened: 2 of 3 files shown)


pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+        2. Create repository bar with foo as a feed and run sync.
+        3. Get content of both repositories.
+        4. Remove random unit from repository foo and publish foo.
+        5. Sync repository bar.
+        6. Get content of both repositories.
+
+    """
+
+    @classmethod
+    def setUpClass(cls, remove_missing=False):  # noqa pylint:disable=arguments-differ,line-too-long
```
Ichimonji10: Why are we disabling line-too-long throughout this entire method? If there's a case where a long line of code needs to be written, then disable this warning in as narrow a scope as possible.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+2. Create second repository, bar, with feed pointing to first repository and
+   set remove_missing=True (RemoveMissingTrueTestCase) and remove_missing=False
+   respectively (RemoveMissingFalseTestCase) and run sync on them.
+3. Assert that repositories contain same set of units.
+4. Remove random unit from repository foo and publish.
+5. Sync bar repo (RemoveMissingFalseTestCase, RemoveMissingFalseTestCase).
+6. Assert that:
+    * when remove_missing=True, content of foo and bar is same
+    * when remove_missing=False, content of foo and bar differ
+
```
Ichimonji10: See #24.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+    def test_unit_removed(self):
+        """Test that correct unit from first repository has been removed."""
+        body = self.responses['units after removal'][0].json()
+        units_names = set(unit['metadata']['name'] for unit in body
+                          if unit['unit_type_id'] == 'rpm')
+        self.assertNotIn(self.removed_unit, units_names)
+
+
+class _RemoveMissingTestCase(_BaseTestCase):
```
Ichimonji10: `_BaseTestCase` and `_RemoveMissingTestCase` can be collapsed into a single class. I don't see any advantage to splitting them up.
classes

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+        cls.responses['publish'] = []
+        cls.responses['publish'].append(client.post(
+            urljoin(repos[0]['_href'], 'actions/publish/'),
+            {'id': cls.responses['distribute'].json()['id']},
+        ))
+
+        # Use http feed instead of https to avoid possible config problems
+        bodies[1]['importer_config']['feed'] = urljoin(
+            # Create http url from base_url
```
Ichimonji10: If SSL errors are raised when the test suite is running and the errors are caused by Pulp, I think a good response is to file a bug with Pulp and work around that bug as narrowly as possible.

If SSL errors are raised because the Pulp system the user has is incorrectly configured, then it's the user's job to either fix that SSL misconfiguration or set `verify=False` in the Pulp Smas config file.

Let's not ignore the `base_url` that the user has set in their configuration file. Ignoring the configuration file is a Bad Thing. Also, why is a private method used here?

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+                    []
+                )
+
+    def test_units_before_removal(self):
+        """Test that units in repositories before removal are the same."""
+        bodies = [re.json() for re in self.responses['units before removal']]
+        # Package category and package group will differ so we count only RPMs
+        self.assertEqual(
+            set(unit['unit_id'] for unit in bodies[0]
+                if unit['unit_type_id'] == 'rpm'),  # This test is fragile
```
Ichimonji10: Please don't use hanging indents. This request is purely stylistic; but they're used nowhere in Pulp Smash right now.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+        rpms = [unit['metadata']['name']
+                for unit in cls.responses['units before removal'][0].json()
+                if unit['unit_type_id'] == 'rpm']
+        cls.removed_unit = random.choice(rpms)
+        # Remove unit from first repo and publish again
+        cls.responses['remove unit'] = client.post(
+            urljoin(repos[0]['_href'], 'actions/unassociate/'),
+            {'criteria':
+             {'fields':
+              {'unit': ['name', 'epoch', 'version', 'release',
```
Ichimonji10: Please use four-space indents.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+                        'arch', 'checksum', 'checksumtype']},
```
Ichimonji10: Please use four-space indents, and please avoid hanging indents.

Ichimonji10:
The `_CommonTestMixin` provides methods very similar to what is already present elsewhere in Pulp Smash. For example:

* `test_status_code` is similar to [`pulp_smash.tests.rpm.api_v2.test_sync_publish.PublishTestCase.test_status_code`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.puppet.api_v2.test_sync_publish.html#pulp_smash.tests.puppet.api_v2.test_sync_publish.PublishTestCase.test_status_code).
* `test_task_error_traceback` is similar to [`pulp_smash.tests.puppet.api_v2.test_sync_publish.SyncInvalidFeedTestCase.test_task_error_traceback`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.puppet.api_v2.test_sync_publish.html#pulp_smash.tests.puppet.api_v2.test_sync_publish.SyncInvalidFeedTestCase.test_task_error_traceback) and [`pulp_smash.tests.puppet.api_v2.test_sync_publish.SyncValidFeedTestCase.test_task_error_traceback`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.puppet.api_v2.test_sync_publish.html#pulp_smash.tests.puppet.api_v2.test_sync_publish.SyncValidFeedTestCase.test_task_error_traceback).

More generally, the `RemoveMissingTrueTestCase` and `RemoveMissingFalseTestCase` act much like the existing sync and publish test cases. Each of these test cases performs actions like syncing, adding distributors, and adding publishers, and then records the responses from pulp and makes assertions about those responses.

In a way, it's nice to have these tests. More tests means better constraints on Pulp's behaviour, right? However, these new tests are largely redundant: module [`pulp_smash.tests.rpm.api_v2.test_sync_publish`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.puppet.api_v2.test_sync_publish.html) already provides assurance that Pulp is capable of creating RPM repositories and performing syncs. And redundant test code has several costs:

* Adding these tests will extend the amount of time it takes for Pulp Smash to execute. (It already takes at least ten minutes to execute the test suite on a pair of 2.7 and 2.8 systems on my development system.) A long test suite execution time is bad because it extends the feedback loop experienced by developers and other users. We want people to use Pulp Smash, not say "eh, I don't want to wait 30 minutes to see if my changes are OK."
* Adding these tests will increase the maintenance burden. If Pulp changes the format of its task reports, or if Pulp changes the status codes it returns, or anything else, it should be easy for us to update Pulp Smash accordingly. That's only possible if there's only a handful of tests (ideally, one test) for each behavior.

Can we make this new test module provide a unique set of tests?

p.s. Yes, the sync and publish tests are fairly redundant right now. Modules `pulp_smash.tests.{rpm,ostree,puppet}.api_v2.test_sync_publish` all contain very similar code. I'd like to pull out common code into a set of functions and mixins.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: Can you link to Pulp's documentation somewhere in this docstring? It'd be nice to have a reference for these tests.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
peterlacko: I needed to disable `arguments-differ` warning and by adding that line got 81 chars long, I didn't know though that it disables `line-too-long` warning for whole method. Renaming `remove_missing` to e.g. `rm_missing` would do the trick.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
peterlacko: There is no advantage, just conforming to the rest of the tests. Changing.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
peterlacko: OK, last time :)

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
peterlacko: Fixed.

peterlacko:
I think that in this case (`--remove-missing` option), having assertions about correctly executed sync, publish etc. makes sense. When testing (1621)[https://pulp.plan.io/issues/1621], sync in this test fails, even when `pulp_smash.tests.rpm.api_v2.test_sync_publish` tests pass. We could only test content of repos before and after removing unit and syncing, but checking also that sync, publish... finished successfuly adds almost no time to total execution time and provides valuable info in case of failure.

Putting common code into set of functions is good idea though. I can imagine also removing  `RemoveMissingFalseTestCase` class, since it's behavior is most probably the same, as without `--remove-missing` option at all and using repository with less units than zoo.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
peterlacko: There isn't much about this parameter, only thing I could find is https://pulp-rpm.readthedocs.org/en/latest/tech-reference/yum-plugins.html

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: Ah, sorry. I made a mistake here. This disables line-too-long for this single line. The code is correct.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: That helps.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: Thank you.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: Thank you.

Ichimonji10:
There's assertions, and then there's assertions.  I agree it makes sense to assert that the syncs in this test succeed. But I don't agree with the approach of writing 80+ lines of setup code and adding in redundant assertions. We already know, for example, that the server returns HTTP 202 responses when asked to sync and publish a repository - what benefit is there in checking again here?

Yes, syncs can fail when using the `--remove-missing` option. So let's do something like this to ensure the sync succeeds:

```python
url = urljoin(repo['_href'], 'actions/sync/')
client.post(url, {'override_config': {}})
```

Those two lines of code will recursively poll every task spawned by the "sync" action and verify that their status codes are sane.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+"""
+
+from __future__ import unicode_literals
+
+try:  # try Python 3 import first
+    from urllib.parse import urljoin, urlparse
+except ImportError:
+    from urlparse import urljoin, urlparse  # pylint:disable=C0411,E0401
+
+import random
```
Ichimonji10: Nitpick: This is a standard library import. Please place a few lines up, with the other standard lib imports.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+import unittest2
+
+from pulp_smash import api, config, utils
+from pulp_smash.constants import (
+    REPOSITORY_PATH,
+)
```
Ichimonji10: Nitpick: This import can fit on one line.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
```diff
@@ -0,0 +1,283 @@
...
+
+_PUBLISH_DIR = 'pulp/repos/'
+_FEED_URL = 'https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/'
```
Ichimonji10: See just-added commit, c01635fc88963106da1a7fc5e7aa79fc5bc8b119

Ichimonji10:
> We could only test content of repos before and after removing unit and syncing, but checking also that sync, publish... finished successfuly adds almost no time to total execution time and provides valuable info in case of failure.

All things considered, I do see some logic behind this. I guess another way to approach this is to keep the test structure basically as you have it, and to raise the priority of creating mixins and functions to refactor out redundant code.

peterlacko:
Reply to [previous comment](https://github.com/PulpQE/pulp-smash/pull/103#issuecomment-182877834): When I poll for a task status, the response is 200, and 404 only if requested [task is not found found](http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/tasks.html#polling-task-progress), right? `poll_spawned_tasks` method from `pulp_smash.utils` also doesn't check if task was successful, just if it is in one of states: `'canceled', 'error', 'finished', 'skipped', 'timed out'`. So even when task failed, we won't see a traceback, i.e. when I remove `test_task_error_traceback` method from tests, only test failure I will get is 

````
$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_remove_unit
..F.
FAIL: test_units_after_removal (pulp_smash.tests.rpm.api_v2.test_remove_unit.RemoveMissingTrueTestCase)
Test that units in repositories after removal are the same.
  File "pulp_smash/tests/rpm/api_v2/test_remove_unit.py", line 229, in test_units_after_removal
    set(unit['unit_id'] for unit in bodies[1]
AssertionError: Items in the second set but not the first:
u'de46e3ae-4e43-4d9a-b65e-aea003f95077'

Ran 4 tests in 34.682s

FAILED (failures=1)
````
which I think isn't enough. But the code really is redundant, so what about checking task status in `poll_spawned_tasks` method, or better, adding another handler, that would check also task's state and not only return code?

File: pulp_smash/tests/rpm/api_v2/test_remove_unit.py (hunk 4 above)
Comment (peterlacko) [inline pulp_smash/tests/rpm/api_v2/test_remove_unit.py]: Having http there is also valid, since we are publishing over both, http and https. I see advantage in testing on content published over http as: If I want to use pulp only over insecure connection, there's no point in using https (and we even prevent user from running test). If I want to use https feed, there should be other tests that verify that everything works. And I'm not 100% sure about this, but I think that certificate that is configured in pulp smash config file, is used only for talking to pulp's api, while for using secure feed one must correctly configure pulp itself.

ad `_replace`: I found it as a best solution for replacing one item in tuple, that would fit to one line. Please suggest something better if you know :)

**Comment by Ichimonji10** [comment]
`poll_spawned_tasks` is pretty good at just waiting for tasks to complete. Let's leave it as is, and not give it a new responsibility.

I like the idea of adding a new handler that checks both return codes and task states. It's nice when code is safe by default. Writing code like that has the potential to prevent an entire class of bugs. :smile: Good suggestion.

It might be appropriate to edit `pulp_smash.api._handle_202`:

```diff
diff --git a/pulp_smash/api.py b/pulp_smash/api.py
index 1975739..868f78e 100644
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -48,7 +48,11 @@ def _handle_202(server_config, response):
     """Check for an HTTP 202 response and handle it appropriately."""
     if response.status_code == 202:  # "Accepted"
         _check_http_202_content_type(response)
-        tuple(utils.poll_spawned_tasks(server_config, response.json()))
+        call_report = response.json()
+        tasks = tuple(utils.poll_spawned_tasks(server_config, call_report))
+        # Should we re-read `call_report` now that tasks are complete?
+        _check_call_report(call_report)  # FIXME: Write this function.
+        _check_tasks(tasks)  # FIXME: Write this function.
 
 
 def echo_handler(server_config, response):  # pylint:disable=unused-argument
```

Ichimonji10:
I should warn you that I'm going to want a separate commit for changes to `pulp_smash.api`. You can add that here, or it might be easier to submit those changes as a separate pull request.

pulp_smash/tests/rpm/api_v2/test_remove_unit.py
Ichimonji10: > I see advantage in testing on content published over http as: If I want to use pulp only over insecure connection, there's no point in using https

I agree that it's nonsensical to make content available over HTTPS and then ignore SSL verification errors. But I've heard from several people who try to use Pulp Smash and immediately get confused when tests fail due to incorrect HTTPS configurations. My solution to that is to let them disable HTTPS verification errors via the configuration file. 

> And I'm not 100% sure about this, but I think that certificate that is configured in pulp smash config file, is used only for talking to pulp's api, while for using secure feed one must correctly configure pulp itself.

Pulp Smash's configuration file cannot be used to configure SSL certificates. It only supports the "verify" option, which controls whether `pulp_smash.api.Client` verifies SSL certificates when requests are made. This class is used for all HTTP and HTTP requests. It's used when making API calls, and it's used when fetching files published by Pulp.

> If I want to use https feed, there should be other tests that verify that everything works.

It's valuable to test that both HTTP and HTTPS publishing works. So let's make a separate test case for that, rather than ignoring the "base_url" set by the user.

Ichimonji10:
Merged as 570b840807800005512a58edbdfaefbdf2f59786, with a follow-up commit of 524aa5360cd94199deb80ebc0d0be69d9943d8d2. Thank you, @peterlacko.

PR #106, opened 2016-02-12
Adds login command to the docker cli test setUp

dkliban:
Tests were failing because the commands were being executed without logging in
or providing -u and -p options.

### DIFF
--- a/pulp_smash/tests/docker/cli/test_crud.py
+++ b/pulp_smash/tests/docker/cli/test_crud.py
@@ -21,6 +21,10 @@ class CreateTestCase(unittest2.TestCase):
         """Provide a server config and a repository ID."""
         self.cfg = config.get_config()
         self.repo_id = utils.uuid4()
+        cli.Client(self.cfg).run(
+            'pulp-admin login -u {} -p {}'
+            .format(self.cfg.auth[0], self.cfg.auth[1]).split()
+        )
 
     def tearDown(self):
         """Delete created resources."""
--- a/pulp_smash/tests/docker/cli/test_sync.py
+++ b/pulp_smash/tests/docker/cli/test_sync.py
@@ -29,6 +29,10 @@ class _BaseTestCase(unittest2.TestCase):
         """Provide a server config and a repository ID."""
         cls.cfg = config.get_config()
         cls.repo_id = utils.uuid4()
+        cli.Client(cls.cfg).run(
+            'pulp-admin login -u {} -p {}'
+            .format(cls.cfg.auth[0], cls.cfg.auth[1]).split()
+        )
 
     @classmethod
     def tearDownClass(cls):


pulp_smash/tests/docker/cli/test_crud.py
```diff
@@ -21,6 +21,9 @@ def setUp(self):
         """Provide a server config and a repository ID."""
         self.cfg = config.get_config()
         self.repo_id = utils.uuid4()
+        cli.Client(self.cfg).run(
+            'pulp-admin login -u admin -p admin'.split()
```
Ichimonji10: `self.cfg` has an "auth" attribute. It looks like this:

```python
>>> cfg.auth
['admin', 'admin']
```

pulp_smash/tests/docker/cli/test_crud.py
```diff
@@ -21,6 +21,9 @@ def setUp(self):
         """Provide a server config and a repository ID."""
         self.cfg = config.get_config()
         self.repo_id = utils.uuid4()
+        cli.Client(self.cfg).run(
+            'pulp-admin login -u {} -p {}'.format(self.cfg.auth[0], self.cfg.auth[1]).split()
```
Ichimonji10: This won't pass `make lint`. Try something like:

```diff
diff --git a/pulp_smash/tests/docker/cli/test_crud.py b/pulp_smash/tests/docker/cli/test_crud.py
index 17ddf86..6648569 100644
--- a/pulp_smash/tests/docker/cli/test_crud.py
+++ b/pulp_smash/tests/docker/cli/test_crud.py
@@ -22,7 +22,8 @@ class CreateTestCase(unittest2.TestCase):
         self.cfg = config.get_config()
         self.repo_id = utils.uuid4()
         cli.Client(self.cfg).run(
-            'pulp-admin login -u {} -p {}'.format(self.cfg.auth[0], self.cfg.auth[1]).split()
+            'pulp-admin login -u {} -p {}'
+            .format(self.cfg.auth[0], self.cfg.auth[1]).split()
         )
 
     def tearDown(self):
diff --git a/pulp_smash/tests/docker/cli/test_sync.py b/pulp_smash/tests/docker/cli/test_sync.py
index ec170f8..46dd465 100644
--- a/pulp_smash/tests/docker/cli/test_sync.py
+++ b/pulp_smash/tests/docker/cli/test_sync.py
@@ -30,7 +30,8 @@ class _BaseTestCase(unittest2.TestCase):
         cls.cfg = config.get_config()
         cls.repo_id = utils.uuid4()
         cli.Client(cls.cfg).run(
-            'pulp-admin login -u {} -p {}'.format(cls.cfg.auth[0], cls.cfg.auth[1]).split()
+            'pulp-admin login -u {} -p {}'
+            .format(cls.cfg.auth[0], cls.cfg.auth[1]).split()
         )
 
     @classmethod
```

Ichimonji10:
It's currently impossible to execute `pip install -r requirements.txt`. That's why builds are currently failing. I'll look into it. In the meantime, I can verify that this PR passes `make test lint`. I have yet to execute the full test suite with this PR to see if test results change.

When you spin up a VM with Vagrant, is that system configured in such a way that `pulp-admin` does not require the `-u` or `-p` options? I ask because I've never had issues executing commands with pulp-admin, but all the VMs I've been executing commands on are Vagrant VMs.

If that's the case, might that be cause for concern? Making the Vagrant VMs as similar as possible to how a production system acts helps prevent issues like this, wherein users develop code that works in one context but not another.

dkliban:
Vagrant logs in and creates some repos for you. As a result you can run pulp-admin commands afterwards.

Ichimonji10:
So if a repository has been created, it's no longer necessary to provide credentials for subsequent `pulp-admin` commands?

dkliban:
Once you execute `pulp-admin login -u <username> -p <password>` you don't need to provide the credentials with each command. 

You can also just always provide -u and -p options with all commands.

Ichimonji10:
Ahhh. `pulp-admin login --help` tells me that it logs in and downloads a session certificate. Is this certificate saved as `~/.pulp/user-cert.pem` and re-used for subsequent commands if no username and password are provided?

dkliban:
That is correct.

Ichimonji10:
TIL

Thanks @dkliban.

Ichimonji10:
Merged as 738aecd58d77b985b720fbf4f99d0d107f3111be. I took the liberty of rewriting the commit message.

Thank you @dkliban! I especially appreciate the explanation of how `pulp-admin` works, and it's nice to know that that's done automatically for our Vagrant setup.

PR #109, opened 2016-02-12
Change how OSTree tests are selected

Ichimonji10:
A function named `load_tests` is in module `pulp_smash.tests.ostree`. This function is automatically called by the unittest (or unittest2) test runner according to the load_tests protocol. [1] This works well, but few if any third party test runners support the protocol. (To the best of my knowledge, only
nose2 supports the load_tests protocol, and this support is nominal.)

Many third party test runners add JUnit-style XML reporting. This reporting is valuable, especially when combined with a tool like Jenkins. Unfortunately, it's not so easy to get reports with the unittest test runner.

The conflict here is tha we want JUnit-style XML reporting *and* easy test selection. There's a couple routes forward:

* Drop support for unittest (and unittest2) and adopt the test selection features of a third party test runner.
* Plow forward and put more work into figuring out how to make unittest produce reports.
* Don't use the load_tests protocol. Instead, add `setUpModule` functions that perform the same task into each individual module of tests.

This commit adopts the third route. This route ensures that Pulp Smash continues to be compatible with the standard library's `unittest` module and allows reports to be generated by third party tools. The main drawback is some level of redundant code. (Options for reducing this redundancy may be available.)

Test results are as expected for the following two commands:

    python -m unittest2 discover pulp_smash.tests.ostree
    py.test pulp_smash/tests/ostree/


Ichimonji10:
@dkliban Sanity check, please.

pulp_smash/tests/ostree/api_v2/test_sync_publish.py
```diff
@@ -39,6 +39,30 @@
 )
 
 
+def setUpModule():
+    """Skip tests if the OSTree plugin is not installed."""
+    if 'ostree' not in _get_plugin_type_ids():
+        raise unittest2.SkipTest('These tests require the OSTree plugin.')
+
+
+def _get_plugin_type_ids():
```
dkliban: perhaps move this method into utils.py ?

Ichimonji10: Aye, I'd very much like to do that! I can't do so at the moment due to a circular import issue.

Ichimonji10: This method makes use of `api.Client` and `api.json_handler`, and the `api` module uses the `utils.poll_spawned_tasks` function.

Ichimonji10: Perhaps the right solution is to move `utils.poll_spawned_tasks` to the `api` module, given that it works with the API? I'll make an issue for it.

dkliban:
In that case this is good. You should file the other issue so we can refactor this later.

Ichimonji10:
See #110.

Issue #73, opened 2016-01-27
Automate Lazy sync - on_demand


preethit:
API info
https://github.com/pulp/pulp/blob/master/docs/dev-guide/integration/rest-api/repo/sync.rst#download-a-repository

1. Create a repo with on_demand download policy
pulp-admin rpm repo create --repo-id zoo --feed https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/  --download-policy on_demand
2. Run repo Sync
3. The repo is synced  (check /var/lib/pulp/content )
4. Register a consumer and bind the repo
5. Install a package on the consumer 
6. Make sure that the package is downloaded and installed
7. Make sure that only the package requested is downloaded

preethit:
This is from jcline- 

class SyncOnDemandTestCase(SyncValidFeedTestCase):
    """Assert that Yum supports on-demand syncing."""

    @classmethod
    def setUpClass(cls):
        """Create an RPM repository with a valid feed and sync it."""
        super(SyncOnDemandTestCase, cls).setUpClass()
        cls.post_sync_repo = handle_response(requests.get(
                cls.cfg.base_url + cls.attrs_iter[0]['_href'] + '?details=true',
                **cls.cfg.get_requests_kwargs()
            )
        )

    def test_no_local_units(self):
        """Assert no content units were downloaded."""
        self.assertEqual(0, self.post_sync_repo['locally_stored_units'])

    def test_repository_units(self):
        """Assert there are some content units in the repository."""
        self.assertGreater(0, self.post_sync_repo['total_repository_units'])

Ichimonji10:
Here's a revision of the sample code above:

```python
class SyncOnDemandTestCase(SyncValidFeedTestCase):
    """Assert Yum supports on-demand syncing."""

    @classmethod
    def setUpClass(cls):
         """Create an RPM repository with a valid feed and sync it."""
        super(SyncOnDemandTestCase, cls).setUpClass()
        # Parent class creates an RPM repository with a valid feed and syncs
        # it.
        cls.repo_info = client.get(cls.repo['_href'], params={'details': True})

    def test_no_local_units(self):
        """Assert no content units were downloaded."""
        self.assertEqual(self.repo_info['locally_stored_units'], 0)

    def test_repository_units(self):
        """Assert there is at least one content unit in the repository."""
        self.assertGreater(self.repo_info['total_repository_units'], 0)
```

[...]

PR #117, opened 2016-02-16
Add tests for lazy on-demand syncing for RPM.

jeremycline:
This closes issue #73

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -31,6 +31,7 @@ Assertions not explored in this module include:
 """
 from __future__ import unicode_literals
 
+import hashlib
 from itertools import product
 try:  # try Python 3 import first
     from urllib.parse import urljoin
@@ -46,6 +47,7 @@ from pulp_smash.constants import (
     REPOSITORY_PATH,
     RPM,
     RPM_FEED_URL,
+    RPM_SHA256_CHECKSUM,
 )
 
 
@@ -420,3 +422,99 @@ class PublishTestCase(utils.BaseAPITestCase):
         for i, module in enumerate(self.rpms[1:]):
             with self.subTest(i=i):
                 self.assertEqual(self.rpms[0], module)
+
+
+class SyncOnDemandTestCase(utils.BaseAPITestCase):
+    """Assert Yum supports on-demand syncing."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create an RPM repository with a valid feed and sync it.
+
+        Do the following:
+
+        0. Reset Pulp, including the Squid cache.
+        1. Sync/publish a repository using the 'on_demand' download policy.
+        2. Download an RPM from the published repository.
+        3. Download the same RPM to ensure it is served by the cache.
+        """
+        super(SyncOnDemandTestCase, cls).setUpClass()
+
+        # Required to ensure the `locally_stored_units` is 0 before we start.
+        utils.reset_squid(cls.cfg)
+        utils.reset_pulp(cls.cfg)
+
+        client = api.Client(cls.cfg, api.json_handler)
+
+        # Create the repository
+        body = _gen_repo()
+        body['importer_config'] = {
+            'feed': RPM_FEED_URL,
+            'download_policy': 'on_demand',
+        }
+        distributor = _gen_distributor()
+        distributor['auto_publish'] = True
+        distributor['distributor_config']['relative_url'] = body['id']
+        body['distributors'] = [distributor]
+
+        repo = client.post(REPOSITORY_PATH, body)
+
+        # Sync the repository
+        sync_path = urljoin(repo['_href'], 'actions/sync/')
+        client.post(sync_path, {'override_config': {}})
+
+        cls.repo = client.get(repo['_href'], params={'details': True})
+
+        client.response_handler = api.safe_handler
+        url = urljoin(
+            '/pulp/repos/',
+            repo['id'] + '/',
+        )
+        url = urljoin(url, RPM)
+        cls.rpm = client.get(url)
+        cls.same_rpm = client.get(url)
+
+    def test_local_units(self):
+        """Assert no content units were downloaded besides metadata."""
+        metadata_unit_count = sum([
+            count for name, count in self.repo['content_unit_counts'].items()
+            if name not in ('rpm', 'drpm', 'srpm')
+        ])
+        self.assertEqual(
+            self.repo['locally_stored_units'],
+            metadata_unit_count
+        )
+
+    def test_repository_units(self):
+        """Assert there is at least one content unit in the repository."""
+        total_units = sum(self.repo['content_unit_counts'].values())
+        self.assertEqual(self.repo['total_repository_units'], total_units)
+
+    def test_request_history(self):
+        """Assert that the initial request received a 302 Redirect."""
+        self.assertTrue(self.rpm.history[0].is_redirect)
+
+    def test_rpm_checksum(self):
+        """Assert the checksum of the downloaded RPM matches the metadata."""
+        checksum = hashlib.sha256(self.rpm.content).hexdigest()
+        self.assertEqual(RPM_SHA256_CHECKSUM, checksum)
+
+    def test_rpm_cache_lookup_header(self):
+        """Assert the first request resulted in a cache miss from Squid."""
+        self.assertIn('MISS', self.rpm.headers['X-Cache-Lookup'])
+
+    def test_rpm_cache_control_header(self):
+        """Assert that the request has the Cache-Control header set."""
+        self.assertEqual(
+            {'s-maxage=86400', 'public', 'max-age=86400'},
+            set(self.rpm.headers['Cache-Control'].split(', '))
+        )
+
+    def test_same_rpm_checksum(self):
+        """Assert the checksum of the second RPM matches the metadata."""
+        checksum = hashlib.sha256(self.same_rpm.content).hexdigest()
+        self.assertEqual(RPM_SHA256_CHECKSUM, checksum)
+
+    def test_same_rpm_cache_header(self):
+        """Assert the second request resulted in a cache hit from Squid."""
+        self.assertIn('HIT', self.same_rpm.headers['X-Cache-Lookup'])
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -69,8 +69,8 @@ def reset_pulp(server_config):
     prefix = '' if client.run(('id', '-u')).stdout.strip() == '0' else 'sudo '
     client.run('mongo pulp_database --eval db.dropDatabase()'.split())
     client.run('sudo -u apache pulp-manage-db'.split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/content/*').split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/published/*').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/content').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/published').split())
 
     for service in services:
         service.start()
@@ -106,3 +106,25 @@ class BaseAPITestCase(unittest2.TestCase):
         client = api.Client(cls.cfg)
         for resource in cls.resources:
             client.delete(resource)
+
+
+def reset_squid(server_config):
+    """Stop Squid, reset its cache directory, and restart it.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server being targeted.
+    :returns: Nothing.
+    """
+    squid_service = cli.Service(server_config, 'squid')
+    squid_service.stop()
+
+    # Clean out the cache directory and reinitialize it.
+    client = cli.Client(server_config)
+    prefix = '' if client.run(('id', '-u')).stdout.strip() == '0' else 'sudo '
+    client.run((prefix + 'rm -rf /var/spool/squid').split())
+    client.run((prefix + 'mkdir --context=system_u:object_r:squid_cache_t:s0' +
+                ' --mode=750 /var/spool/squid').split())
+    client.run((prefix + 'chown squid:squid /var/spool/squid').split())
+    client.run((prefix + 'squid -z').split())
+
+    squid_service.start()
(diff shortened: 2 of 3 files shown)


pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -77,6 +79,39 @@ def _gen_distributor():
     }
 
 
+def _sync_and_publish_repo(server_config, importer=None, distributors=None):
+    """Create and sync an RPM repository.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
+    :param importer: A dict-like object with importer configuration.
+    :returns: A tuple of
```
Ichimonji10: What is returned?

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -77,6 +79,39 @@ def _gen_distributor():
     }
 
 
+def _sync_and_publish_repo(server_config, importer=None, distributors=None):
+    """Create and sync an RPM repository.
+
+    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
```
Ichimonji10: The `param` directive accepts either two or three arguments. Try something like this:

```
        Pulp server being targeted.
```

Then run `make docs-html` and check out the result. It's pretty cool!

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -77,6 +79,39 @@ def _gen_distributor():
     }
 
 
+def _sync_and_publish_repo(server_config, importer=None, distributors=None):
```
Ichimonji10: Is there any reason to make this function as generic as it is? This function is generic, so it's not going to get widespread use. In addition, it is used only once:

```python
repo, tasks = _sync_and_publish_repo(
    cls.cfg,
    importer=importer_config
)
```

I ask because, in my opinion, it is a liability to add in capabilities that are not needed or used. How about changing this function signature to the following?

```python
def _sync_and_publish_repo(server_config, importer):
```

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -438,3 +473,85 @@ def test_unit_integrity(self):
...
+        utils.reset_pulp(cls.cfg)
+        importer_config = {
+            'feed': RPM_FEED_URL,
+            'download_policy': 'on_demand',
+        }
+        repo, tasks = _sync_and_publish_repo(
+            cls.cfg,
+            importer=importer_config
+        )
+        client = api.Client(cls.cfg, api.echo_handler)
```
Ichimonji10: Why is `api.echo_handler` used here? It's a no-op callback; it returns server responses untouched.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -438,3 +473,85 @@ def test_unit_integrity(self):
...
+        self.assertEqual(RPM_SHA256_CHECKSUM, checksum)
+
+    def test_rpm_cache_lookup_header(self):
+        """Assert the first request resulted in a cache miss from Squid."""
+        self.assertIn('MISS', self.rpm.headers['X-Cache-Lookup'])
+
+    def test_rpm_cache_control_header(self):
+        """Assert that the request has the Cache-Control header set."""
+        self.assertEqual(
+            'public, s-maxage=86400, max-age=86400',
```
Ichimonji10: Can these headers legally appear in some other order? That is, can the server legally return one of the following?

* `public, max-age=86400, s-maxage=86400`
* `s-maxage=86400, max-age=86400, public`
* …another seven combinations

pulp_smash/utils.py
```diff
@@ -67,8 +67,29 @@ def reset_pulp(server_config):
...
-    client.run((prefix + 'rm -rf /var/lib/pulp/published/*').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/content').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/published').split())
 
     for service in services:
         service.start()
+
+
+def reset_squid(server_config):
+    """
```
Ichimonji10: Please document this function.

pulp_smash/utils.py
```diff
@@ -67,8 +67,29 @@ def reset_pulp(server_config):
     prefix = '' if client.run(('id', '-u')).stdout.strip() == '0' else 'sudo '
     client.run('mongo pulp_database --eval db.dropDatabase()'.split())
     client.run('sudo -u apache pulp-manage-db'.split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/content/*').split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/published/*').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/content').split())
+    client.run((prefix + 'rm -rf /var/lib/pulp/published').split())
```
Ichimonji10: Why is this change being made? Consider explaining this change in the commit message.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
jeremycline: I originally copy/pasted its functionality from a different test and since you mentioned in the issue it might be nice to have a function that syncs a valid repo, I added it (and promptly neglected to use the function where I originally copied from). 

Personally, I think the two use cases diverged enough that having a function to share the task is not worth the trouble, so I'm more than happy to remove it.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
jeremycline: It won't appear in any other order, although I suppose technically the specification doesn't forbid it. I'm happy changing these to `assertIn` tests to make it more flexible, though.

Ichimonji10:
Please write a proper commit message. For example commit messages, browse `git log` or the [online commit log](https://github.com/PulpQE/pulp-smash/commits/master). (Even better, [read up on the subject](https://duckduckgo.com/?q=how+to+write+a+commit+message). I still fail to do things like hard-wrap commit message bodies at 72 chars, and could probably focus more heavily on explaining the *why* in my commit messages.)

pulp_smash/utils.py
```diff
@@ -67,8 +67,29 @@ def reset_pulp(server_config):
...
 
     for service in services:
         service.start()
+
+
+def reset_squid(server_config):
```
Ichimonji10: Consider adding this function in a separate commit. It is easier to revert commits if each contains one logical change, among other benefits.

pulp_smash/utils.py
Ichimonji10: Again, consider making this change a separate commit.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -438,3 +473,85 @@ def test_unit_integrity(self):
...
+        cls.rpm = client.get(url)
+        cls.same_rpm = client.get(url)
+
+    def test_local_units(self):
+        """Assert no content units were downloaded besides metadata."""
+        metadata_units = filter(
+            lambda unit_tuple: unit_tuple[0] not in ('rpm', 'drpm', 'srpm'),
+            self.repo['content_unit_counts'].items()
+        )
+        metadata_unit_count = sum([unit[1] for unit in metadata_units])
```
Ichimonji10: Is `repo['content_unit_counts']` a dict in the following form?

```python
{
    'foo': 1,
    'bar': 3,
    'biz': 10,
    'baz': 5,
}
```

If so, this can be written as a list comprehension:

```python
>>> counts = {'foo': 1, 'bar': 2, 'biz': 3, 'baz': 4}
>>> sum((
...     number for name, number in counts.items()
...     if name not in {'foo', 'bar'}
... ))
7
```

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: There's a number of ways to tackle this. We could make this an `assertIn`, but that would leave the possibility of additional cache control headers being added. Is that something that should be guarded against?

This is the most equivalent test I can think off offhand:

```python
self.assertEqual(
    {'s-maxage=86400', 'public', 'max-age=86400'},
    set(self.rpm.headers['Cache-Control'].split(', '))
)
```

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
Ichimonji10: Removal is fine; trimming down the function is fine; some other more ingenious solution is fine. So long as the solution fits the problem.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
jeremycline: It could be. My personal preference is to use `filter`, but I'm fine with either way.

dkliban:
I got the following error when I tried running this PR:

`[jenkins@rhel7-vanilla-np-qeos-73966 pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_sync_publish
.................sE....
ERROR: setUpClass (pulp_smash.tests.rpm.api_v2.test_sync_publish.SyncOnDemandTestCase)
  File "pulp_smash/tests/rpm/api_v2/test_sync_publish.py", line 503, in setUpClass
    importer=importer_config
  File "pulp_smash/tests/rpm/api_v2/test_sync_publish.py", line 110, in _sync_and_publish_repo
    tasks = tuple(utils.poll_spawned_tasks(server_config, sync))
AttributeError: 'module' object has no attribute 'poll_spawned_tasks'

`

jeremycline:
@dkliban ahh, I rebased and I guess failed to resolve the merge conflicts correctly.

Ichimonji10:
FYI, as of c6308d4193bcbc0264daad307a98213741ad3a9d, `pulp_smash.api.safe_handler` and `pulp_smash.api.json_handler` check the final state of asynchronous tasks.

Ichimonji10:
The specific error you're seeing may be a result of 45cadf41ea6fd2320a2c97366fb68a541c3e680e.

jeremycline:
@Ichimonji10 It's going to be a while before I have time for this. It will probably be much more efficient for everyone involved if you simply take what's here and modify it to fit your needs.

Alternatively you can re-review it and it'll be a few days before I address the issues.

Ichimonji10:
@jeremycline OK, I may do that. Thanks for the status update. :smile:

Ichimonji10:
I took what is in this pull request and modified it in https://github.com/Ichimonji10/pulp-smash/tree/117. Unfortunately, most of the tests pass. It should be revisited for sanity.

jeremycline:
@Ichimonji10 Is there something else you're expecting from me? That link is no longer around and I'm not really sure what you mean by 'Unfortunately, most of the tests pass'.

Ichimonji10:
Last I checked, most of the tests fail. Can you take a look at them and see if they still fail or not? Sorry about the broken link. That was a snafu on my part.

pulp_smash/utils.py
jeremycline: The globs don't expand with sudo, it seems.

Ichimonji10: :+1:

Ichimonji10:
This test fails for me with the following error:

```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='dev.example.com', port=443): Max retries exceeded with url: /streamer/var/lib/pulp/content/units/rpm/76/78177c241777af22235092f21c3932dd4f0664e1624e5a2c77a201ec70f930/bear-4.1-1.noarch.rpm?policy=eyJleHRlbnNpb25zIjogeyJyZW1vdGVfaXAiOiAiMTkyLjE2OC4xMjEuMSJ9LCAicmVzb3VyY2UiOiAiL3N0cmVhbWVyL3Zhci9saWIvcHVscC9jb250ZW50L3VuaXRzL3JwbS83Ni83ODE3N2MyNDE3NzdhZjIyMjM1MDkyZjIxYzM5MzJkZDRmMDY2NGUxNjI0ZTVhMmM3N2EyMDFlYzcwZjkzMC9iZWFyLTQuMS0xLm5vYXJjaC5ycG0iLCAiZXhwaXJhdGlvbiI6IDE0NTYxNjY2NzR9;signature=UpAVdrBnJib6aVTJPuOSKPgRjXu7UbfGNdhNF8vmX_qZryUzXeiMuhrFdSEwknMVJDUBHmmo9Hf8DNJG9_FUhVCY_a4nq1l18bv14Q9sMDV2g1sP5tJr5F0VzIAX68raCvtrkCeiUeW3UrBRy29z43rhuJve9goeRQZd3tXnaaiJ-cNxfIQ-4ldri1XLNLITcokeLjgVcr8wKQUVg6zx5O4uVKY4pvrmZ1zHu5rpX8X5xsaYGew0gXpm0CRpwYWgH3-uVkAnMbucGBBOkBPDAGeL2acabyT0H6p6Yzb7zMvJLlDjTows_z5Qj8fIdg41WLwFm_IOySPwMp5UYtXHlQ%3D%3D (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7face40d5210>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

The interesting part there is the mention of "dev.example.com." That makes me think that this test will succeed when run from within a development Vagrant VM, but not in other contexts. Any insight here?

jeremycline:
@Ichimonji10 Is the Pulp server's hostname `dev.example.com`?

I'm guessing some setup automation needs to change to set up the reverse proxy, caching proxy, and streamer for lazy content syncing. The Ansible role we use in the development environment[0] is probably a good starting point, but the Squid configuration file needs to be different for RHEL6[1].

[0] https://github.com/pulp/pulp/tree/master/playpen/ansible/roles/lazy
[1] https://github.com/pulp/pulp/blob/master/docs/user-guide/deferred-download.rst#squid

Ichimonji10:
> @Ichimonji10 Is the Pulp server's hostname `dev.example.com`?

Yes. For what it's worth, I'm running the test suite outside of the VM itself.

jeremycline:
Yes, that's expected behavior. Lazy redirects (by default) the the FQDN of the host Pulp is running on.

Ichimonji10:
@dkliban, do the Pulp VMs used by Jenkins return correct hostnames?

@jeremycline, is this feature new in 2.8?

jeremycline:
@Ichimonji10 yes, it's new in 2.8

Ichimonji10:
Merged as cd92cd16c3365da575fcb7bbf552bb490fa8c333 and 036f67d9b3c0c14d925969360027a7ae7bfd8a83. Thank you, @jeremycline!

Issue #105, opened 2016-02-11
Test sync with dockerhub with unnamespaced repositories


asmacdo:
https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/docker/cli/test_sync.py#L16

The repository's real location is `library/busybox`, but when pulling from Dockerhub, the Docker client automatically prepends `library/` to any unnamespaced repos. Syncing from `upstream-name=busybox` should have exactly the same result.

Ichimonji10:
Should we have tests that do both? That is, should we use both `--upstream-name=busybox` and `--upstream-name=library/busybox`?

rbarlow:
@Ichimonji10 I don't seem to be able to self-assign this issue - can you assign it to me? I'm gonna hack on it this afternoon.

rbarlow:
And yeah, I'll just make a test for each style.

rbarlow:
Haha, I think @asmacdo is actually working on this, but hadn't assigned it when I asked. Sorry for the confusion!

PR #119, opened 2016-02-16
Test sync of a Docker repository from Dockerhub without a namespace

asmacdo:
The `upstream-name` of a Docker repository indicates the correct feed
to sync from the registry. For Dockerhub only, `registry-1.docker.io`,
some special repositories do not have a namespace, any it would be
correct to set their upstream name to be the repo_name by itself. In
this case, the sync will work exactly the same as if the upstream
name were `library/<repo_name>`.

https://pulp.plan.io/issues/1404
Closes: https://github.com/PulpQE/pulp-smash/issues/105

### DIFF
--- a/pulp_smash/tests/docker/cli/test_sync.py
+++ b/pulp_smash/tests/docker/cli/test_sync.py
@@ -94,6 +94,28 @@ class SyncV2TestCase(_SuccessMixin, _BaseTestCase):
         cls.completed_proc = _sync_repo(cls.cfg, cls.repo_id)
 
 
+class SyncUnnamespacedV2TestCase(_SuccessMixin, _BaseTestCase):
+    """Show it Pulp can sync an unnamespaced docker repo from a v2 registry."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a docker repository from an unnamespaced v2 registry.
+
+        This method requires Pulp 2.8 and above, and will raise a ``SkipTest``
+        exception if run against an earlier version of Pulp.
+        """
+        super(SyncUnnamespacedV2TestCase, cls).setUpClass()
+        if cls.cfg.version < Version('2.8'):
+            raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
+        kwargs = {
+            'feed': 'https://registry-1.docker.io',  # v2 feed
+            'repo_id': cls.repo_id,
+            'upstream_name': 'busybox',
+        }
+        cli.Client(cls.cfg).run(_CREATE_COMMAND.format(**kwargs).split())
+        cls.completed_proc = _sync_repo(cls.cfg, cls.repo_id)
+
+
 class InvalidFeedTestCase(_BaseTestCase):
     """Show Pulp behaves correctly when syncing a repo with an invalid feed."""
 


pulp_smash/tests/docker/cli/test_sync.py
```diff
@@ -94,6 +94,28 @@ def setUpClass(cls):
         cls.completed_proc = _sync_repo(cls.cfg, cls.repo_id)
 
 
+class SyncUnnamespacedV2TestCase(_SuccessMixin, _BaseTestCase):
+    """Show it Pulp can sync an unnamespaced docker repo from a v2 registry."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a docker repository and from an unnamespaced v2 registry.
```
Ichimonji10: "and from"?

Ichimonji10:
Good job with the added code. It looks great.

Can we twiddle with the commit message a bit? There's a couple things to do here:

* Make use of GitHub's ability to [close issues via commit messages](https://help.github.com/articles/closing-issues-via-commit-messages/).
* Make the commit message body valid (or valid-ish) markdown. I do this partly because markdown is an extremely popular and simple standard that is easily read as plain text, and partly because GitHub compiles commit message bodies using markdown before displaying them online. So - why not? In this case, all that needs to be done is to indent the sample code.
* Restrict the opening summary to the conventional 50 characters long. Yeah, it's super restrictive. But following it ensures that various commit parsing tools can deal nicely with it.

By the way, I want to say that the commit message is already quite nice. I especially appreciate that you took the time to write the summary in the imperative voice. That makes the output of `git shortlog` and other such tools much more useful. (Just think of the possibilities! How about weekly snapshot releases with auto-generated release notes, with a make target to automate the release process?)

Here's a sample of what I have in mind.

```
Add test for syncing an un-namespaced docker repo

The `upstream-name` of a Docker repository indicates the correct feed to
sync from the registry. For Dockerhub only, `registry-1.docker.io`, some
special repositories do not have a namespace, any it would be correct to
set their upstream name to be the repo_name by itself. In this case, the
sync will work exactly the same as if the upstream name were
`library/<repo_name>`.

The newly added tests pass:

    (pulp-smash)[vagrant@dev pulp-smash]$ python -m unittest2 pulp_smash.tests.docker.cli.test_sync
    s...........
    Ran 12 tests in 56.368s

Resolves #105

Also see https://pulp.plan.io/issues/1404
```

Ichimonji10:
That done, I'll walk you through rebasing and committing this PR. If possible, I'd like to have you do that.

asmacdo:
@Ichimonji10, I have fixed up the commit match your specs, but I am wondering if the restrictions and process are a little much. I recognize the benefits of the conventions (and I really love how consistent and beautiful they are), but I hope you will consider loosening some of them to lower the barrier to entry for contribution to the project. Particularly since we want to encourage casual contribution from the pulp dev team, style requirements (particularly regarding commit messages) should be as minimal as possible.

Whatever hard requirements for commits are decided, I'll volunteer to add them to the pulp-smash contribution docs.

Ichimonji10:
@asmacdo I tend to be loose about contribution requirements, and then fix things myself when committing others' PRs.

Ichimonji10:
Basically - that's why I like doing commits myself. I can edit commits to match whatever standards I have, while leaving the barrier to contribution a bit lower for other people. And that's why I'm going through this process with you: the hope is that there are some other people who I can trust to do the same.

Ichimonji10:
Thanks for updating this PR. I can confirm that the added tests pass. Test suite
results:

    2.7        6           OK (skipped=3)
    dev (2.8)  12          OK (skipped=1)

Results generated with:

    python -m unittest2 pulp_smash.tests.docker.cli.test_sync

Merged as 110657a6c12cdb1db314c6bf836c823e40b794e8

asmacdo:
@Ichimonji10, gotcha. Thanks!

Ichimonji10:
Thank *you*. It's good to have #105 resolved.

PR #120, opened 2016-02-17
Check Task report and Call report for errors.

peterlacko:
This commit extends functionality of pulp_smash.api.safe_handler
to check for errors in Task report or Call report and if error was
reported, raises TaskReportError or CallReportError respectively.

This commit does not change number of tests or their results.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -53,11 +53,34 @@ def _warn_http_202_content_type(response):
     warnings.warn(message, RuntimeWarning)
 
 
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Merged as c6308d4193bcbc0264daad307a98213741ad3a9d.

Thank you! This is a really nice systematic improvement that will allow us to more consistently validate asynchronous tasks. At the same time, it significantly reduces the amount of necessary test code. That's *awesome*.

Issue #122, opened 2016-02-17
Drop redundant call report and task report tests


Ichimonji10:
As of c6308d4193bcbc0264daad307a98213741ad3a9d,

> `pulp_smash.api.safe_handler` and `pulp_smash.api.json_handler`
> check for errors in task reports and call reports, and raise
> `TaskReportError` or `CallReportError` respectively if one is found.

As a result, many test methods are no longer necessary. Let's walk through tests that explicitly check call reports and task reports, drop redundant checks, and drop the code that bothers to collect and save those call reports and task reports if possible.

Ichimonji10:
#121 makes use of this new behaviour to great effect.

elyezer:
Had an IRC chat session with @Ichimonji10 and have clarified the following:

Both `api.safe_handler` and `api.json_handler` call `_handle_202`, which calls `_check_call_report` and `_check_tasks`. _check_tasks() ensures that each task's "error", "exception" and "traceback" fields. Given that, there's no reason to also check those fields in unit tests.

Ichimonji10:
Thanks for clarifying.

Issue #104, opened 2016-02-11
Decide how to host content for pulp-smash to test syncs


asmacdo:
Currently, the feed is pulp's fedorapeople for RPM, puppet forge, and dockerhub. It would be ideal for pulp-smash to have its own demo content, even if it is just a mirror of the content we currently use.

This will speed up the tests and will enable pulp-smash to test syncing with repositories that have the traits necessary to replicate certain bugs. For example:  https://github.com/PulpQE/pulp-smash/issues/93

Ichimonji10:
Being able to customize where Pulp Smash tests get content from would have several benefits:

* If some content source becomes unavailable, we could simply point Pulp Smash at some other content source. This would lower our reliance on outside systems that are outside our control.
* Tests can be *much* faster and more robust if the content is being pulled from a nearby source.
* If the content source is the local system, the test suite could be run completely offline.

Ichimonji10:
My main worry is getting the implementation details right. One proposal is to make demo content available and then asking users to download that content locally and then setting up a content server. This could be cool - but how hard would that be to do? Setting up server for RPM files is easy: it requires just executing `python -m http.server 80` at the root of some directory tree. But what about docker and puppet, where the content server has to implement an API?

One option is to use Pulp itself as a content source. The big advantages here are that:

* We'd be dog-fooding.
* Pulp does support serving all these content types.

Bug there's some caveats:

* Pulp can be slow and buggy at times. (I'm not attacking anyone here! But let's be realistic.)
* Pulp can be hard to install.
* Pulp *can't* be installed on many platforms.
* Forcing the test suite user to set up a content server before testing Pulp raises the barrier to entry!

These cons can be mitigated if we have a Pulp server publicly (?) available to anyone who wants to use it.

rbarlow:
The Python sync can't be done from a Pulp server (or a static content server) - it requires PyPI because of an interactive API.

Ichimonji10:
So that's at least docker, puppet and python that must be served from something that implements an API.

asmacdo:
https://botbot.me/freenode/docker-distribution/2016-02-22/?msg=60658740&page=1

This comment from a docker dev suggests that syncing a lot of content from docker hub could possibly result in being throttled.

omaciel:
@asmacdo should we create/maintain our own registry and perhaps have our tests pointing at it? This may solve one issue, but then, how can we make our own registry 'relevant' with what Docker Hub offers so that we're always up to date?

[...]

PR #126, opened 2016-02-18
Add tests cases for copying Docker units.

### LINKED ISSUE #98: Repo copy of docker 2.0 repo
1. Create a docker repo with feed of v2 

      Assert: Repo is created

2. Sync the repo

    Assert: Repo sync successful

3. Create a docker repo with no feed
     Assert:No feed repo created

4. Run repo copy from synced repo  to the destination repo

    Assert: Repo copy successful

5. Run repo list

    Assert: Lists repo 
             Repo content in the destination same as the source.

rbarlow:
This commit adds several test cases for copying Docker units.
Additionally, it adds a handy utils module that contains several
convenient functions that are useful for these test cases, as well
as for other future test cases.

Fixes #98

### DIFF
--- a/pulp_smash/tests/docker/cli/test_copy.py
+++ b/pulp_smash/tests/docker/cli/test_copy.py
@@ -1,0 +1,232 @@
+# coding=utf-8
+"""Tests for copying docker units between repositories."""
+from __future__ import unicode_literals
+import re
+
+import unittest2
+from packaging.version import Version
+
+from pulp_smash import utils
+from pulp_smash.tests.docker.cli import utils as docker_utils
+
+
+class CopyAllImagesTestCase(docker_utils.BaseTestCase,
+                            docker_utils.SuccessMixin):
+    """Test copying all Images from one repository to another."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create and sync a docker repository with a v1 registry."""
+        super(CopyAllImagesTestCase, cls).setUpClass()
+        if cls.cfg.version < Version('2.8'):
+            raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
+        # Create and sync a repository with a v1 feed to bring some Images in.
+        docker_utils.create_repo(
+            cls.cfg, cls.repo_id, upstream_name='library/busybox',
+            sync_v1=True, sync_v2=False)
+        docker_utils.sync_repo(cls.cfg, cls.repo_id)
+
+        # Now create a feedless repository and copy the images from the first
+        # repository over.
+        cls.copy_target = utils.uuid4()
+        docker_utils.create_repo(cls.cfg, cls.copy_target)
+        cls.completed_proc = docker_utils.copy(cls.cfg, 'image', cls.repo_id,
+                                               cls.copy_target)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the copy_target repo."""
+        super(CopyAllImagesTestCase, cls).tearDownClass()
+        docker_utils.delete_repo(cls.cfg, cls.copy_target)
+
+    def test_positive_copy_output(self):
+        """Assert that the list of image_ids in the copy output are correct."""
+        # We need to limit the fields to image_id as a hack to work around
+        # https://pulp.plan.io/issues/1696. Doing this has a side effect of
+        # showing the image_ids without splitting them with a line break
+        # (sigh).
+        units_in_src = docker_utils.search(self.cfg, 'image', self.repo_id,
+                                           fields=['image_id']).stdout
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the image_ids. Hello regex my old
+        # friend, I've come to talk with you again…
+        image_ids_in_src = set(re.findall(r'(?:Image Id:)\s*(.*)',
+                                          units_in_src))
+
+        image_ids_printed = set(re.findall(r'(?: {2})(\w*)',
+                                           self.completed_proc.stdout))
+
+        self.assertEqual(image_ids_printed, image_ids_in_src)
+
+    def test_positive_units_copied(self):
+        """Assert that all units were copied."""
+        units_in_src = docker_utils.search(self.cfg, 'image',
+                                           self.repo_id).stdout
+        units_in_dest = docker_utils.search(self.cfg, 'image',
+                                            self.copy_target).stdout
+
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the unit_ids. Hello regex my old
+        # friend, I've come to talk with you again…
+        units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
+        units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
+
+        self.assertEqual(units_in_src, units_in_dest)
+
+    def test_task_succeeded(self):
+        """Assert that "Copied" appears in stdout."""
+        self.assertIn('Copied', self.completed_proc.stdout)
+
+
+class CopyAllManifestsTestCase(docker_utils.BaseTestCase,
+                               docker_utils.SuccessMixin):
+    """Test copying all Manifests from one repository to another."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create and sync a docker repository with a v2 registry."""
+        super(CopyAllManifestsTestCase, cls).setUpClass()
+        if cls.cfg.version < Version('2.8'):
+            raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
+        # Create and sync a repository with a v2 feed to bring some Manifests
+        # in.
+        docker_utils.create_repo(
+            cls.cfg, cls.repo_id, upstream_name='library/busybox',
+            sync_v1=False, sync_v2=True)
+        docker_utils.sync_repo(cls.cfg, cls.repo_id)
+
+        # Now create a feedless repository and copy the manifests from the
+        # first repository over.
+        cls.copy_target = utils.uuid4()
+        docker_utils.create_repo(cls.cfg, cls.copy_target)
+        cls.completed_proc = docker_utils.copy(cls.cfg, 'manifest',
+                                               cls.repo_id, cls.copy_target)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the copy_target repo."""
+        super(CopyAllManifestsTestCase, cls).tearDownClass()
+        docker_utils.delete_repo(cls.cfg, cls.copy_target)
+
+    def test_positive_copy_output(self):
+        """Assert that the list of manifests in the copy output are correct."""
+        units_in_src = docker_utils.search(self.cfg, 'manifest', self.repo_id,
+                                           fields=['digest']).stdout
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the digests. Hello regex my old
+        # friend, I've come to talk with you again…
+        manifest_ids_in_src = set(re.findall(r'(?:Digest:)\s*(.*)',
+                                             units_in_src))
+
+        # The manifest digests are printed after "docker_manifest:" in the copy
+        # output.
+        manifest_ids = self.completed_proc.stdout.split('docker_manifest:')[1]
+        manifest_ids_printed = re.findall(r'(?: {2})(\w*:\w*)', manifest_ids)
+        # Due to https://pulp.plan.io/issues/1696 pulp-admin has split the last
+        # character of each of the manifest_ids_in_src items onto the next line
+        # which wasn't matched by the regex. That's stupid, but we're going to
+        # work around it by chopping the last character off of these as well.
+        manifest_ids_printed = set([d[:-1] for d in manifest_ids_printed])
+
+        self.assertEqual(manifest_ids_printed, manifest_ids_in_src)
+
+    def test_positive_units_copied(self):
+        """Assert that all units were copied."""
+        units_in_src = docker_utils.search(self.cfg, 'manifest',
+                                           self.repo_id).stdout
+        units_in_dest = docker_utils.search(self.cfg, 'manifest',
+                                            self.copy_target).stdout
+
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the unit_ids. Hello regex my old
+        # friend, I've come to talk with you again…
+        units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
+        units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
+
+        self.assertEqual(units_in_src, units_in_dest)
+
+    def test_task_succeeded(self):
+        """Assert that "Copied" appears in stdout."""
+        self.assertIn('Copied', self.completed_proc.stdout)
+
+
+class CopyAllTagsTestCase(docker_utils.BaseTestCase,
+                          docker_utils.SuccessMixin):
+    """Test copying all Tags from one repository to another."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create and sync a docker repository with a v2 registry."""
+        super(CopyAllTagsTestCase, cls).setUpClass()
+        if cls.cfg.version < Version('2.8'):
+            raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
+        # Create and sync a repository with a v2 feed to bring some Tags in.
+        docker_utils.create_repo(
+            cls.cfg, cls.repo_id, upstream_name='library/busybox',
+            sync_v1=False, sync_v2=True)
+        docker_utils.sync_repo(cls.cfg, cls.repo_id)
+
+        # Now create a feedless repository and copy the tags from the first
+        # repository over.
+        cls.copy_target = utils.uuid4()
+        docker_utils.create_repo(cls.cfg, cls.copy_target)
+        cls.completed_proc = docker_utils.copy(cls.cfg, 'tag', cls.repo_id,
+                                               cls.copy_target)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the copy_target repo."""
+        super(CopyAllTagsTestCase, cls).tearDownClass()
+        docker_utils.delete_repo(cls.cfg, cls.copy_target)
+
+    def test_positive_copy_output(self):
+        """Assert that the list of tags in the copy output are correct."""
+        units_in_src = docker_utils.search(self.cfg, 'tag', self.repo_id,
+                                           fields=['name']).stdout
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the digests. Hello regex my old
+        # friend, I've come to talk with you again…
+        tag_ids_in_src = set(re.findall(r'(?:Digest:)\s*(.*)', units_in_src))
+
+        # The tag digests are printed after "docker_tag:" in the copy output.
+        tag_ids = self.completed_proc.stdout.split('docker_tag:')[1]
+        tag_ids_printed = re.findall(r'(?: {2})(\w*:\w*)', tag_ids)
+        # Due to https://pulp.plan.io/issues/1696 pulp-admin has split the last
+        # character of each of the tag_ids_in_src items onto the next line
+        # which wasn't matched by the regex. That's stupid, but we're going to
+        # work around it by chopping the last character off of these as well.
+        tag_ids_printed = set([d[:-1] for d in tag_ids_printed])
+
+        self.assertEqual(tag_ids_printed, tag_ids_in_src)
+
+    def test_positive_units_copied(self):
+        """Assert that all units were copied."""
+        tags_in_src = docker_utils.search(self.cfg, 'tag', self.repo_id).stdout
+        tags_in_dest = docker_utils.search(self.cfg, 'tag',
+                                           self.copy_target).stdout
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the names. Hello regex my old
+        # friend, I've come to talk with you again…
+        tags_in_src = set(re.findall(r'(?:Name:)\s*(.*)', tags_in_src))
+        tags_in_dest = set(re.findall(r'(?:Name:)\s*(.*)', tags_in_dest))
+
+        # The Manifests should have been pulled over as well since they are
+        # recursively copied.
+        manifests_in_src = docker_utils.search(self.cfg, 'manifest',
+                                               self.repo_id).stdout
+        manifests_in_dest = docker_utils.search(self.cfg, 'manifest',
+                                                self.copy_target).stdout
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the unit_ids. Hello regex my old
+        # friend, I've come to talk with you again…
+        manifests_in_src = set(re.findall(r'(?:Name:)\s*(.*)',
+                                          manifests_in_src))
+        manifests_in_dest = set(re.findall(r'(?:Name:)\s*(.*)',
+                                           manifests_in_dest))
+
+        self.assertEqual(tags_in_src, tags_in_dest)
+        self.assertEqual(manifests_in_src, manifests_in_dest)
+
+    def test_task_succeeded(self):
+        """Assert that "Copied" appears in stdout."""
+        self.assertIn('Copied', self.completed_proc.stdout)
--- a/pulp_smash/tests/docker/cli/utils.py
+++ b/pulp_smash/tests/docker/cli/utils.py
@@ -1,0 +1,123 @@
+# coding=utf-8
+"""Common utilities that assist in testing the CLI with docker repositories."""
+from __future__ import unicode_literals
+
+import unittest2
+
+from pulp_smash import cli, config, constants, utils
+
+
+def copy(server_config, unit_type, src_repo_id, dest_repo_id):
+    """Use pulp-admin to copy content from one docker repository to another.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server targeted by this function.
+    :param unit_type: The type of content to copy, such as "image" or
+        "manifest." Run ``pulp-admin docker repo copy --help`` to get the full
+        set of available unit types.
+    :param src_repo_id: A value for the ``--from-repo-id`` option.
+    :param src_repo_id: A value for the ``--to-repo-id`` option.
+    """
+    cmd = 'pulp-admin docker repo copy {} --from-repo-id {} --to-repo-id {}'
+    cmd = cmd.format(unit_type, src_repo_id, dest_repo_id).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
+
+
+def create_repo(
+        server_config,
+        repo_id,
+        upstream_name=None,
+        sync_v1=False,
+        sync_v2=False):
+    """Use pulp-admin to create a repo with the given parameters.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server targeted by this function.
+    :param repo_id: A value for the ``--repo-id`` option.
+    :param upstream_name: A value for the ``--upstream-name`` option.
+    :param sync_v1: A value for the ``--enable-v1`` option.
+    :param sync_v2: A value for the ``--enable-v2`` option.
+    """
+    extra_flags = ''
+    if upstream_name:
+        extra_flags += ' --upstream-name {}'.format(upstream_name)
+
+    # Handle whether we are syncing, and if so which APIs
+    if sync_v1 or sync_v2:
+        # The Docker v2 feed URL can do Docker v1 as well
+        extra_flags += ' --feed {}'.format(constants.DOCKER_V2_FEED_URL)
+    if sync_v1:
+        extra_flags += ' --enable-v1 true'
+    else:
+        extra_flags += ' --enable-v1 false'
+    if sync_v2:
+        extra_flags += ' --enable-v2 true'
+    else:
+        extra_flags += ' --enable-v2 false'
+
+    command = 'pulp-admin docker repo create --repo-id {}{}'
+    command = command.format(repo_id, extra_flags).split()
+    return cli.Client(server_config, cli.echo_handler).run(command)
+
+
+def delete_repo(server_config, repo_id):
+    """Delete the repo given by repo_id."""
+    cmd = 'pulp-admin docker repo delete --repo-id {}'.format(repo_id).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
+
+
+def search(server_config, unit_type, repo_id, fields=None):
+    """
+    Search the given repo for all units of given unit_type.
+
+    unit_type should be a string: "image", "blob", "manifest", or "tag".
+    repo_id is the repo_id you wish to search.
+    fields is a list of strings of field names you want to retrieve.
+    """
+    extra_flags = ''
+    if fields:
+        extra_flags += ' --fields {}'.format(','.join(fields))
+    cmd = 'pulp-admin docker repo search {} --repo-id {}{}'
+    cmd = cmd.format(unit_type, repo_id, extra_flags).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
+
+
+def sync_repo(server_config, repo_id):
+    """Synchronize the given repo."""
+    return cli.Client(server_config, cli.echo_handler).run(
+        'pulp-admin docker repo sync run --repo-id {}'.format(repo_id).split()
+    )
+
+
+class BaseTestCase(unittest2.TestCase):
+    """A base class for testing Docker content. It logs in for you."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Provide a server config and a repository ID."""
+        cls.cfg = config.get_config()
+        cls.repo_id = utils.uuid4()
+        cmd = 'pulp-admin login -u {} -p {}'.format(*cls.cfg.auth).split()
+        cli.Client(cls.cfg).run(cmd)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created repository."""
+        cmd = 'pulp-admin docker repo delete --repo-id {}'.format(cls.repo_id)
+        cli.Client(cls.cfg).run(cmd.split())
+
+
+class SuccessMixin(object):
+    """Add some common assertion to test cases."""
+
+    def test_return_code(self):
+        """Assert the "sync" command has a return code of 0."""
+        self.assertEqual(self.completed_proc.returncode, 0)
+
+    def test_task_succeeded(self):
+        """Assert the phrase "Task Succeeded" is in stdout."""
+        self.assertIn('Task Succeeded', self.completed_proc.stdout)
+
+    def test_task_failed(self):
+        """Assert the phrase "Task Failed" is not in stdout."""
+        self.assertNotIn('Task Failed', self.completed_proc.stdout)
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -23,7 +23,7 @@ DOCKER_V1_FEED_URL = 'https://index.docker.io'
 This URL can be used as the "feed" property of a Pulp Docker registry.
 """
 
-DOCKER_V2_FEED_URL = 'https://registry-1.docker.io'
+DOCKER_V2_FEED_URL = DOCKER_V1_FEED_URL
 """The URL to a V2 Docker registry.
 
 This URL can be used as the "feed" property of a Pulp Docker registry.


pulp_smash/tests/docker/cli/utils.py
```diff
@@ -0,0 +1,118 @@
+# coding=utf-8
+"""Common utilities that assist in testing the CLI with docker repositories."""
+from __future__ import unicode_literals
+
+import unittest2
+
+from pulp_smash import cli, config, utils
+
+
+FEED = 'https://index.docker.io'
```
Ichimonji10: As of d93de79f0cec13779a719d45d3a8e4dac1cd34dd, there are now docker feed constants. Can we use one of those constants, or at least make `FEED` private?

By the way, as per our discussion on the mailing lists, I'm not entirely sure if those constants have correct values. It might make more sense to have something like this in the constants file:

```python
DOCKER_V1_FEED_URL = '...'
DOCKER_V2_FEED_URL = DOCKER_V1_FEED_URL
```

Or even something else. Basically: if the docker-related constants don't seem right to you, please speak up and/or submit a pull request.

Ichimonji10:
Hmm. Let's start with the easy stuff. One, please document function and method arguments, especially when they're public. Second, don't be afraid to use positional string interpolation - the results are often more legible. Third, please avoid hanging indents. It's a purely stylistic thing, but they're used no-where else in the codebase right now. Here's a diff that implements these three suggestions for the `utils` module: (read the following comments before investing time into this diff)

```diff
diff --git a/pulp_smash/tests/docker/cli/utils.py b/pulp_smash/tests/docker/cli/utils.py
index 1802cf3..84e83d9 100644
--- a/pulp_smash/tests/docker/cli/utils.py
+++ b/pulp_smash/tests/docker/cli/utils.py
@@ -11,26 +11,43 @@ FEED = 'https://index.docker.io'
 
 
 def copy(server_config, unit_type, src_repo_id, dest_repo_id):
-    """Use pulp-admin to copy all instances of unit_type from src to dest."""
-    command = ('pulp-admin docker repo copy {unit_type} --from-repo-id {src} '
-               '--to-repo-id {dest}')
-    command = command.format(unit_type=unit_type, src=src_repo_id,
-                             dest=dest_repo_id)
-    command = command.split()
-
-    return cli.Client(server_config, cli.echo_handler).run(command)
-
-
-def create_repo(server_config, repo_id, upstream_name=None, sync_v1=False,
-                sync_v2=False):
-    """Use pulp-admin to create a repo with the given parameters."""
+    """Use pulp-admin to copy content from one docker repository to another.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server targeted by this function.
+    :param unit_type: The type of content to copy, such as "image" or
+        "manifest." Run ``pulp-admin docker repo copy --help`` to get the full
+        set of available unit types.
+    :param src_repo_id: A value for the ``--from-repo-id`` option.
+    :param src_repo_id: A value for the ``--to-repo-id`` option.
+    """
+    cmd = 'pulp-admin docker repo copy {} --from-repo-id {} --to-repo-id {}'
+    cmd = cmd.format(unit_type, src_repo_id, dest_repo_id).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
+
+
+def create_repo(
+        server_config,
+        repo_id,
+        upstream_name=None,
+        sync_v1=False,
+        sync_v2=False):
+    """Use pulp-admin to create a repo with the given parameters.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server targeted by this function.
+    :param repo_id: A value for the ``--repo-id`` option.
+    :param upstream_name: A value for the ``--upstream-name`` option.
+    :param sync_v1: A value for the ``--enable-v1`` option.
+    :param sync_v2: A value for the ``--enable-v2`` option.
+    """
     extra_flags = ''
     if upstream_name:
-        extra_flags += ' --upstream-name {n}'.format(n=upstream_name)
+        extra_flags += ' --upstream-name {}'.format(upstream_name)
 
     # Handle whether we are syncing, and if so which APIs
     if sync_v1 or sync_v2:
-        extra_flags += ' --feed {f}'.format(f=FEED)
+        extra_flags += ' --feed {}'.format(FEED)
     if sync_v1:
         extra_flags += ' --enable-v1 true'
     else:
@@ -40,19 +57,15 @@ def create_repo(server_config, repo_id, upstream_name=None, sync_v1=False,
     else:
         extra_flags += ' --enable-v2 false'
 
-    command = ('pulp-admin docker repo create --repo-id '
-               '{repo_id}{extra_flags}').format(
-                   repo_id=repo_id, extra_flags=extra_flags).split()
-
+    command = 'pulp-admin docker repo create --repo-id {}{}'
+    command = command.format(repo_id, extra_flags).split()
     return cli.Client(server_config, cli.echo_handler).run(command)
 
 
 def delete_repo(server_config, repo_id):
     """Delete the repo given by repo_id."""
-    command = 'pulp-admin docker repo delete --repo-id {repo_id}'.format(
-        repo_id=repo_id).split()
-
-    return cli.Client(server_config, cli.echo_handler).run(command)
+    cmd = 'pulp-admin docker repo delete --repo-id {}'.format(repo_id).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
 
 
 def search(server_config, unit_type, repo_id, fields=None):
@@ -65,14 +78,10 @@ def search(server_config, unit_type, repo_id, fields=None):
     """
     extra_flags = ''
     if fields:
-        extra_flags += ' --fields {f}'.format(f=','.join(fields))
-
-    command = ('pulp-admin docker repo search {unit_type} --repo-id '
-               '{repo_id}{eflags}')
-    command = command.format(unit_type=unit_type, repo_id=repo_id,
-                             eflags=extra_flags).split()
-
-    return cli.Client(server_config, cli.echo_handler).run(command)
+        extra_flags += ' --fields {}'.format(','.join(fields))
+    cmd = 'pulp-admin docker repo search {} --repo-id {}{}'
+    cmd = cmd.format(unit_type, repo_id, extra_flags).split()
+    return cli.Client(server_config, cli.echo_handler).run(cmd)
 
 
 def sync_repo(server_config, repo_id):
@@ -83,27 +92,25 @@ def sync_repo(server_config, repo_id):
 
 
 class BaseTestCase(unittest2.TestCase):
-    """A Base class for testing Docker content. It logs in for you."""
+    """A base class for testing Docker content. It logs in for you."""
 
     @classmethod
     def setUpClass(cls):
         """Provide a server config and a repository ID."""
         cls.cfg = config.get_config()
         cls.repo_id = utils.uuid4()
-        cli.Client(cls.cfg).run(
-            'pulp-admin login -u {} -p {}'
-            .format(cls.cfg.auth[0], cls.cfg.auth[1]).split()
-        )
+        cmd = 'pulp-admin login -u {} -p {}'.format(*cls.cfg.auth).split()
+        cli.Client(cls.cfg).run(cmd)
 
     @classmethod
     def tearDownClass(cls):
         """Delete the created repository."""
-        command = 'pulp-admin docker repo delete --repo-id {}'
-        cli.Client(cls.cfg).run(command.format(cls.repo_id).split())
+        cmd = 'pulp-admin docker repo delete --repo-id {}'.format(cls.repo_id)
+        cli.Client(cls.cfg).run(cmd.split())
 
 
 class SuccessMixin(object):
-    """Add some common assertion to Test cases."""
+    """Add some common assertion to test cases."""
 
     def test_return_code(self):
         """Assert the "sync" command has a return code of 0."""
```

Ichimonji10:
I like the idea behind adding a `pulp_smash.tests.docker.cli.utils` module. If we can share code, why not? However, I'm not so sure about the implementation of things as it stands. The benefit behind having such a module is that its code can be shared - but no sharing is done right now.

Making re-usable code is *hard*. Really, I know! In my experience, a good time to introduce re-usable code is when there really are multiple uses for a given piece of code. This has the obvious benefit of ensuring that you've correctly designed that bit of code. But it also has the side benefit of ensuring that changes are better isolated in the commit history, and therefore can be rolled back or understood after-the-fact.

How about adding the new utilities directly into module `pulp_smash.tests.docker.cli.test_copy` as private functions, and then extracting them out at a later point when the re-use is needed?

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+        cls.completed_proc = docker_utils.copy(cls.cfg, 'image', cls.repo_id,
+                                               cls.copy_target)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the copy_target repo."""
+        super(CopyAllImagesTestCase, cls).tearDownClass()
+        docker_utils.delete_repo(cls.cfg, cls.copy_target)
+
+    def test_positive_copy_output(self):
```
Ichimonji10: In just about all of the tests in Pulp Smash, communications with Pulp occur in `setUpClass`, and assertions occur in test methods. However, this test is structured so that communications with Pulp occur in test methods. I don't think that's ideal. Can we push these communications into `setUpClass` and make the test methods concern themselves just with assertions?

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
+        # pulp-admin to limit the output to the unit_ids. Hello regex my old
+        # friend, I've come to talk with you again…
+        units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
+        units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
+
+        self.assertEqual(units_in_src, units_in_dest)
+
+    def test_return_code(self):
```
Ichimonji10: This is done by `SuccessMixin`.

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+        """Assert that the copy command has a return code of 0."""
+        self.assertEqual(self.completed_proc.returncode, 0)
```
Ichimonji10: This is done by `SuccessMixin`, right?

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+        # friend, I've come to talk with you again…
+        manifests_in_src = set(re.findall(r'(?:Name:)\s*(.*)',
+                                          manifests_in_src))
+        manifests_in_dest = set(re.findall(r'(?:Name:)\s*(.*)',
+                                           manifests_in_dest))
+
+        self.assertEqual(tags_in_src, tags_in_dest)
+        self.assertEqual(manifests_in_src, manifests_in_dest)
+
+    def test_return_code(self):
```
Ichimonji10: This is done by `SuccessMixin`, right?

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+        units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
+        units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
+
+        self.assertEqual(units_in_src, units_in_dest)
+
+    def test_return_code(self):
+        """Assert that the copy command has a return code of 0."""
+        self.assertEqual(self.completed_proc.returncode, 0)
+
+    def test_task_succeeded(self):
```
Ichimonji10: This same assertion appears in all three test cases. How about adding this to `SuccessMixin`?

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,244 @@
...
+        manifest_ids_printed = re.findall(r'(?: {2})(\w*:\w*)', manifest_ids)
+        # Due to https://pulp.plan.io/issues/1696 pulp-admin has split the last
+        # character of each of the manifest_ids_in_src items onto the next line
+        # which wasn't matched by the regex. That's stupid, but we're going to
+        # work around it by chopping the last character off of these as well.
+        manifest_ids_printed = set([d[:-1] for d in manifest_ids_printed])
+
+        self.assertEqual(manifest_ids_printed, manifest_ids_in_src)
+
+    def test_positive_units_copied(self):
```
Ichimonji10: A nearly identical test method is present in all three test cases. Consider adding a method like this to `SuccessMixin`:

```python
def test_positive_Units_copied(self):
    """Assert all units were copied."""
        # Due to https://pulp.plan.io/issues/1693 it is not possible to get
        # pulp-admin to limit the output to the unit_ids.
        src_units = set(re.findall(r'(?:Unit Id:)\s*(.*)', self.src_units))
        dest_units = set(re.findall(r'(?:Unit Id:)\s*(.*)', self.dest_units))
        self.assertEqual(units_in_src, units_in_dest)
```

…and then setting `self.src_units` and `self.dest_units` in `setUpClass`.

Ichimonji10:
Ah, I should mention one more thing: the stubs in the `docs/api/` directory are not auto-generated. Adding documentation for new modules requires that a new stub be added.

No fun, I know. :( I am under the impression that Sphinx can auto-generate such, but I've not looked into it.

pulp_smash/tests/docker/cli/utils.py
rbarlow: Ah I didn't realize there were constants. Those values are no longer correct, and also there is now one endpoint that does v1 and v2. I'll change my PR to update those constants. We might as well keep them separate in case docker changes them again later.

rbarlow: Fixed.

rbarlow:
I don't believe there is value in spending time refactoring this PR to not have reusable code, only to immediately refactor the code again on Monday to have reusable code. That's not a good way to use time just because we *might* want to undo a commit later (which is also highly unlikely with this completely additive PR.) Let's focus on spending our time on what *will* benefit us.

Furthermore, the sync.py can/will use this same code, and so I wrote it this way with that in mind as per our discussion in IRC yesterday. I did not do that in this PR because @asmacdo was working in the same area. It was upon your advice to integrate them later.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: What is the benefit of refactoring it this way? This seems like a preference thing to me. The code in question is reading data that is specifically relevant to the assertion. In fact, keeping it this way has the very practical benefit of having the code next to where it is used, instead of off somewhere else. Much more readable and easier to tell what's happening when things go wrong. It's also much easier to create the test this way. What benefit is there to the alternative?

Ichimonji10: One benefit is an improved ability to re-use test methods. See an example [here](https://github.com/PulpQE/pulp-smash/pull/126/files#r53498450). In short, many of these test methods contain highly similar logic, and pushing that logic into `setUpClass` lets us create more reusable test methods.

A second benefit is assertion isolation. For example, given this code:

```python
self.assertEqual(tags_in_src, tags_in_dest)
self.assertEqual(manifests_in_src, manifests_in_dest)
```

...if the first assertion fails, the second will not run, which is likely undesirable. Given that we have only one assertion in this test method, it does not apply in this particular case.

A third benefit is an increased ability to add or remove assertions about a particular outcome. Talking to the server and saving the munged responses just once means that we can write as many assertions as we want targeting the response(s).

Ichimonji10:
I see. I'll revisit this on Monday.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: In your cited example, if the tags aren't in dest, the manifests won't be either. We will know there is a problem and we will fix it. I don't think that's a benefit.

Also, the general pattern in pulp-smash of performing all the "work" in setUpClass means that we need a lot more classes. This means that we sync from docker n times instead of one time in this copy module. I'm not a fan of that, and I'm sure Docker wouldn't be either. If anything, we should have one setUpClass that does the sync, and then each test should do whatever copying it needs along with whatever assertions it needs.

It's easy to add assertions to tests, so I don't think that benefit is meaningfully different between the two approaches.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: Fixed.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: Fixored.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: Fixed.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: These are specific to copying. We could make a copying mixin, but that seems overkill to me.

pulp_smash/tests/docker/cli/test_copy.py
rbarlow: They aren't actually identical, and the SuccessMixin isn't specific to copying. I don't think we need to worry about their similarities, they do work.

pulp_smash/tests/docker/cli/utils.py
Ichimonji10: Thank you.

pulp_smash/tests/docker/cli/test_copy.py
Ichimonji10: Thank you.

pulp_smash/tests/docker/cli/test_copy.py
Ichimonji10: Thank you.

pulp_smash/tests/docker/cli/test_copy.py
Ichimonji10: Thank you.

pulp_smash/tests/docker/cli/test_copy.py
Ichimonji10: Here you go.

```python
class _CopyMixin(object):  # pylint:disable=too-few-public-methods
    """Add assertions common to the test cases in this module."""

    def test_task_succeeded(self):
        """Assert "Copied" appears in stdout."""
        self.assertIn('Copied', self.completed_proc.stdout)
```

pulp_smash/tests/docker/cli/test_copy.py
Ichimonji10: > In your cited example, if the tags aren't in dest, the manifests won't be either.

I've seen other cases where a collection of assertions should intuitively all either succeed or fail, but did not do so. As an example, this set of assertions was added some time back:

```python
counts = self.repo_after_sync.get('content_unit_counts', {})
self.assertEqual(counts.get('rpm'), 32)
self.assertEqual(counts.get('erratum'), 4)
self.assertEqual(counts.get('package_group'), 2)
self.assertEqual(counts.get('package_category'), 1)
```

A bug later cropped up that affected only the RPM counts.

> Also, the general pattern in pulp-smash of performing all the "work" in setUpClass means that we need a lot more classes. This means that we sync from docker n times instead of one time in this copy module.

If repeatedly syncing from docker is problematic, there are several solutions. One is to perform a sync in `setUpModule`. A second is to follow the suggestion you give — and after some thought, I think the solution you give may be better, so props for bringing it up.

> I'm not a fan of that, and I'm sure Docker wouldn't be either.

Is your worry that repeatedly syncing content consumes unnecessary bandwidth, thus slowing down the tests and driving up the cost of operating the docker hub?

If so: sure! As mentioned above, there are several patterns we can adopt to reduce the amount of downloading performed. In addition, #104 helpfully outlines the benefits of fetching data from local sources.

> It's easy to add assertions to tests, so I don't think that benefit is meaningfully different between the two approaches.

I have found that placing business logic in `setUpClass` and assertions in test methods tends to make it easier to write well-isolated complex assertions and re-use test methods. Now, this is only a trend, and there are certainly classes of problems that are better solved other ways, but I would still hesitate to dismiss the benefits of this approach.

Ichimonji10:
> I don't believe there is value in spending time refactoring this PR to not have reusable code, only to immediately refactor the code again on Monday to have reusable code. That's not a good way to use time just because we *might* want to undo a commit later (which is also highly unlikely with this completely additive PR.) Let's focus on spending our time on what *will* benefit us.

Ahh. So the plan is to take the code in `pulp_smash.tests.docker.cli.utils` and re-use it in the immediate future? That changes the equation. Can we explicitly enumerate how the code will be re-used? Naming existing modules (e.g. `test_sync`) and open issues requesting more test cases will certainly aid me as I review this PR, and the added traceability may be of use in the future.

So: if there's immediate re-use planned for the future, that provides some rationale for adding these two modules in one go. I would still rather have atomic commits that make easily reverted, isolated changes. (There are several ways to do this. One option is to place all new test code in a single module and pull out common code in a later commit, as originally suggested. A second option is to pull common code out of the `test_sync` module in one commit and to add these tests in a second commit. There are more options, too.) But I'm willing to compromise in response to the practical desire to add test coverage sooner rather than later, and in response to the apparently strong feelings expressed thus far.

pulp_smash/tests/docker/cli/test_copy.py
```diff
@@ -0,0 +1,232 @@
...
+            cls.cfg, cls.repo_id, upstream_name='library/busybox',
+            sync_v1=True, sync_v2=False)
+        docker_utils.sync_repo(cls.cfg, cls.repo_id)
+
+        # Now create a feedless repository and copy the images from the first
+        # repository over.
+        cls.copy_target = utils.uuid4()
+        docker_utils.create_repo(cls.cfg, cls.copy_target)
+        cls.completed_proc = docker_utils.copy(cls.cfg, 'image', cls.repo_id,
+                                               cls.copy_target)
```
Ichimonji10: I made a comment regarding code style earlier. Similar comments apply to this module. Here's a diff:

```diff
diff --git a/pulp_smash/tests/docker/cli/test_copy.py b/pulp_smash/tests/docker/cli/test_copy.py
index dec4616..9a1faaa 100644
--- a/pulp_smash/tests/docker/cli/test_copy.py
+++ b/pulp_smash/tests/docker/cli/test_copy.py
@@ -10,8 +10,16 @@ from pulp_smash import utils
 from pulp_smash.tests.docker.cli import utils as docker_utils
 
 
-class CopyAllImagesTestCase(docker_utils.BaseTestCase,
-                            docker_utils.SuccessMixin):
+class _CopyMixin(object):  # pylint:disable=too-few-public-methods
+    """Add assertions common to the test cases in this module."""
+
+    def test_task_succeeded(self):
+        """Assert "Copied" appears in stdout."""
+        self.assertIn('Copied', self.completed_proc.stdout)
+
+
+class CopyAllImagesTestCase(
+        docker_utils.BaseTestCase, docker_utils.SuccessMixin, _CopyMixin):
     """Test copying all Images from one repository to another."""
 
     @classmethod
@@ -22,16 +30,24 @@ class CopyAllImagesTestCase(docker_utils.BaseTestCase,
             raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
         # Create and sync a repository with a v1 feed to bring some Images in.
         docker_utils.create_repo(
-            cls.cfg, cls.repo_id, upstream_name='library/busybox',
-            sync_v1=True, sync_v2=False)
+            cls.cfg,
+            cls.repo_id,
+            upstream_name='library/busybox',
+            sync_v1=True,
+            sync_v2=False,
+        )
         docker_utils.sync_repo(cls.cfg, cls.repo_id)
 
         # Now create a feedless repository and copy the images from the first
         # repository over.
         cls.copy_target = utils.uuid4()
         docker_utils.create_repo(cls.cfg, cls.copy_target)
-        cls.completed_proc = docker_utils.copy(cls.cfg, 'image', cls.repo_id,
-                                               cls.copy_target)
+        cls.completed_proc = docker_utils.copy(
+            cls.cfg,
+            'image',
+            cls.repo_id,
+            cls.copy_target,
+        )
 
     @classmethod
     def tearDownClass(cls):
@@ -45,41 +61,46 @@ class CopyAllImagesTestCase(docker_utils.BaseTestCase,
         # https://pulp.plan.io/issues/1696. Doing this has a side effect of
         # showing the image_ids without splitting them with a line break
         # (sigh).
-        units_in_src = docker_utils.search(self.cfg, 'image', self.repo_id,
-                                           fields=['image_id']).stdout
+        units_in_src = docker_utils.search(
+            self.cfg,
+            'image',
+            self.repo_id,
+            fields=['image_id'],
+        ).stdout
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the image_ids. Hello regex my old
-        # friend, I've come to talk with you again…
-        image_ids_in_src = set(re.findall(r'(?:Image Id:)\s*(.*)',
-                                          units_in_src))
-
-        image_ids_printed = set(re.findall(r'(?: {2})(\w*)',
-                                           self.completed_proc.stdout))
-
+        # pulp-admin to limit the output to the image_ids.
+        image_ids_in_src = set(
+            re.findall(r'(?:Image Id:)\s*(.*)', units_in_src)
+        )
+        image_ids_printed = set(
+            re.findall(r'(?: {2})(\w*)', self.completed_proc.stdout)
+        )
         self.assertEqual(image_ids_printed, image_ids_in_src)
 
     def test_positive_units_copied(self):
         """Assert that all units were copied."""
-        units_in_src = docker_utils.search(self.cfg, 'image',
-                                           self.repo_id).stdout
-        units_in_dest = docker_utils.search(self.cfg, 'image',
-                                            self.copy_target).stdout
+        units_in_src = docker_utils.search(
+            self.cfg,
+            'image',
+            self.repo_id,
+        ).stdout
+        units_in_dest = docker_utils.search(
+            self.cfg,
+            'image',
+            self.copy_target,
+        ).stdout
 
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the unit_ids. Hello regex my old
-        # friend, I've come to talk with you again…
-        units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
-        units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
+        # pulp-admin to limit the output to the unit_ids.
+        regex = r'(?:Unit Id:)\s*(.*)'
+        units_in_src = set(re.findall(regex, units_in_src))
+        units_in_dest = set(re.findall(regex, units_in_dest))
 
         self.assertEqual(units_in_src, units_in_dest)
 
-    def test_task_succeeded(self):
-        """Assert that "Copied" appears in stdout."""
-        self.assertIn('Copied', self.completed_proc.stdout)
-
 
-class CopyAllManifestsTestCase(docker_utils.BaseTestCase,
-                               docker_utils.SuccessMixin):
+class CopyAllManifestsTestCase(
+        docker_utils.BaseTestCase, docker_utils.SuccessMixin, _CopyMixin):
     """Test copying all Manifests from one repository to another."""
 
     @classmethod
@@ -91,16 +112,24 @@ class CopyAllManifestsTestCase(docker_utils.BaseTestCase,
         # Create and sync a repository with a v2 feed to bring some Manifests
         # in.
         docker_utils.create_repo(
-            cls.cfg, cls.repo_id, upstream_name='library/busybox',
-            sync_v1=False, sync_v2=True)
+            cls.cfg,
+            cls.repo_id,
+            upstream_name='library/busybox',
+            sync_v1=False,
+            sync_v2=True,
+        )
         docker_utils.sync_repo(cls.cfg, cls.repo_id)
 
         # Now create a feedless repository and copy the manifests from the
         # first repository over.
         cls.copy_target = utils.uuid4()
         docker_utils.create_repo(cls.cfg, cls.copy_target)
-        cls.completed_proc = docker_utils.copy(cls.cfg, 'manifest',
-                                               cls.repo_id, cls.copy_target)
+        cls.completed_proc = docker_utils.copy(
+            cls.cfg,
+            'manifest',
+            cls.repo_id,
+            cls.copy_target,
+        )
 
     @classmethod
     def tearDownClass(cls):
@@ -110,18 +139,24 @@ class CopyAllManifestsTestCase(docker_utils.BaseTestCase,
 
     def test_positive_copy_output(self):
         """Assert that the list of manifests in the copy output are correct."""
-        units_in_src = docker_utils.search(self.cfg, 'manifest', self.repo_id,
-                                           fields=['digest']).stdout
+        units_in_src = docker_utils.search(
+            self.cfg,
+            'manifest',
+            self.repo_id,
+            fields=['digest'],
+        ).stdout
+
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the digests. Hello regex my old
-        # friend, I've come to talk with you again…
-        manifest_ids_in_src = set(re.findall(r'(?:Digest:)\s*(.*)',
-                                             units_in_src))
+        # pulp-admin to limit the output to the digests.
+        manifest_ids_in_src = set(
+            re.findall(r'(?:Digest:)\s*(.*)', units_in_src)
+        )
 
         # The manifest digests are printed after "docker_manifest:" in the copy
         # output.
         manifest_ids = self.completed_proc.stdout.split('docker_manifest:')[1]
         manifest_ids_printed = re.findall(r'(?: {2})(\w*:\w*)', manifest_ids)
+
         # Due to https://pulp.plan.io/issues/1696 pulp-admin has split the last
         # character of each of the manifest_ids_in_src items onto the next line
         # which wasn't matched by the regex. That's stupid, but we're going to
@@ -132,26 +167,27 @@ class CopyAllManifestsTestCase(docker_utils.BaseTestCase,
 
     def test_positive_units_copied(self):
         """Assert that all units were copied."""
-        units_in_src = docker_utils.search(self.cfg, 'manifest',
-                                           self.repo_id).stdout
-        units_in_dest = docker_utils.search(self.cfg, 'manifest',
-                                            self.copy_target).stdout
+        units_in_src = docker_utils.search(
+            self.cfg,
+            'manifest',
+            self.repo_id,
+        ).stdout
+        units_in_dest = docker_utils.search(
+            self.cfg,
+            'manifest',
+            self.copy_target,
+        ).stdout
 
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the unit_ids. Hello regex my old
-        # friend, I've come to talk with you again…
+        # pulp-admin to limit the output to the unit_ids.
         units_in_src = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_src))
         units_in_dest = set(re.findall(r'(?:Unit Id:)\s*(.*)', units_in_dest))
 
         self.assertEqual(units_in_src, units_in_dest)
 
-    def test_task_succeeded(self):
-        """Assert that "Copied" appears in stdout."""
-        self.assertIn('Copied', self.completed_proc.stdout)
-
 
-class CopyAllTagsTestCase(docker_utils.BaseTestCase,
-                          docker_utils.SuccessMixin):
+class CopyAllTagsTestCase(
+        docker_utils.BaseTestCase, docker_utils.SuccessMixin, _CopyMixin):
     """Test copying all Tags from one repository to another."""
 
     @classmethod
@@ -162,16 +198,24 @@ class CopyAllTagsTestCase(docker_utils.BaseTestCase,
             raise unittest2.SkipTest('These tests require Pulp 2.8 or above.')
         # Create and sync a repository with a v2 feed to bring some Tags in.
         docker_utils.create_repo(
-            cls.cfg, cls.repo_id, upstream_name='library/busybox',
-            sync_v1=False, sync_v2=True)
+            cls.cfg,
+            cls.repo_id,
+            upstream_name='library/busybox',
+            sync_v1=False,
+            sync_v2=True,
+        )
         docker_utils.sync_repo(cls.cfg, cls.repo_id)
 
         # Now create a feedless repository and copy the tags from the first
         # repository over.
         cls.copy_target = utils.uuid4()
         docker_utils.create_repo(cls.cfg, cls.copy_target)
-        cls.completed_proc = docker_utils.copy(cls.cfg, 'tag', cls.repo_id,
-                                               cls.copy_target)
+        cls.completed_proc = docker_utils.copy(
+            cls.cfg,
+            'tag',
+            cls.repo_id,
+            cls.copy_target,
+        )
 
     @classmethod
     def tearDownClass(cls):
@@ -181,11 +225,14 @@ class CopyAllTagsTestCase(docker_utils.BaseTestCase,
 
     def test_positive_copy_output(self):
         """Assert that the list of tags in the copy output are correct."""
-        units_in_src = docker_utils.search(self.cfg, 'tag', self.repo_id,
-                                           fields=['name']).stdout
+        units_in_src = docker_utils.search(
+            self.cfg,
+            'tag',
+            self.repo_id,
+            fields=['name']
+        ).stdout
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the digests. Hello regex my old
-        # friend, I've come to talk with you again…
+        # pulp-admin to limit the output to the digests.
         tag_ids_in_src = set(re.findall(r'(?:Digest:)\s*(.*)', units_in_src))
 
         # The tag digests are printed after "docker_tag:" in the copy output.
@@ -201,32 +248,37 @@ class CopyAllTagsTestCase(docker_utils.BaseTestCase,
 
     def test_positive_units_copied(self):
         """Assert that all units were copied."""
+        regex = r'(?:Name:)\s*(.*)'
+
         tags_in_src = docker_utils.search(self.cfg, 'tag', self.repo_id).stdout
-        tags_in_dest = docker_utils.search(self.cfg, 'tag',
-                                           self.copy_target).stdout
+        tags_in_dest = docker_utils.search(
+            self.cfg,
+            'tag',
+            self.copy_target
+        ).stdout
+
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the names. Hello regex my old
-        # friend, I've come to talk with you again…
-        tags_in_src = set(re.findall(r'(?:Name:)\s*(.*)', tags_in_src))
-        tags_in_dest = set(re.findall(r'(?:Name:)\s*(.*)', tags_in_dest))
+        # pulp-admin to limit the output to the names.
+        tags_in_src = set(re.findall(regex, tags_in_src))
+        tags_in_dest = set(re.findall(regex, tags_in_dest))
 
         # The Manifests should have been pulled over as well since they are
         # recursively copied.
-        manifests_in_src = docker_utils.search(self.cfg, 'manifest',
-                                               self.repo_id).stdout
-        manifests_in_dest = docker_utils.search(self.cfg, 'manifest',
-                                                self.copy_target).stdout
+        manifests_in_src = docker_utils.search(
+            self.cfg,
+            'manifest',
+            self.repo_id
+        ).stdout
+        manifests_in_dest = docker_utils.search(
+            self.cfg,
+            'manifest',
+            self.copy_target
+        ).stdout
+
         # Due to https://pulp.plan.io/issues/1693 it is not possible to get
-        # pulp-admin to limit the output to the unit_ids. Hello regex my old
-        # friend, I've come to talk with you again…
-        manifests_in_src = set(re.findall(r'(?:Name:)\s*(.*)',
-                                          manifests_in_src))
-        manifests_in_dest = set(re.findall(r'(?:Name:)\s*(.*)',
-                                           manifests_in_dest))
+        # pulp-admin to limit the output to the unit_ids.
+        manifests_in_src = set(re.findall(regex, manifests_in_src))
+        manifests_in_dest = set(re.findall(regex, manifests_in_dest))
 
         self.assertEqual(tags_in_src, tags_in_dest)
         self.assertEqual(manifests_in_src, manifests_in_dest)
-
-    def test_task_succeeded(self):
-        """Assert that "Copied" appears in stdout."""
-        self.assertIn('Copied', self.completed_proc.stdout)
```

Ichimonji10:
This past weekend, I spent some time working with the code in this PR. How about I take that work, polish it a bit further, and then post back here when I'm done? It may be easier to show than to tell.

rbarlow:
@Ichimonji10 why not merge this PR as is, and then apply the patch you made on it immediately thereafter? I am eager to move on to other work, and I already have a need to use the utils module from this PR in a follow up PR.

Ichimonji10:
See:

1. 9e282eec1f14d6fb06a968a297316426d895309a
2. 012fa3a81b6660b3beece9c6125fe03353ff80d3
3. 8ac192b2f3442b6bd6ec38da9001555f0f1f21db

rbarlow:
Thanks! This will be useful as I work on https://pulp.plan.io/issues/1710

Ichimonji10:
Def. I hope the utilities fulfill your needs. Feel free to expand them or add more - I think their structure is very consistent, and I hope you'll find the same. Thank you for providing the tests.

rbarlow:
The assertions that the return codes are 0 and the dockblock params in utils are both missing in your commits. You made comments about both of those, so you might want to add those back.

Ichimonji10:
There's no need to assert that CLI return codes are 0. The CLI client does that automatically. See [`pulp_smash.cli.code_handler`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.cli.html#pulp_smash.cli.code_handler).

By the way, similar functionality is present by default in the API client. That's the rationale for #122.

I've documented the meaning of the CLI parameters in the module-wide docstring. See [`pulp_smash.tests.docker.cli.utils`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.docker.cli.utils.html). It states the following:

> All of the functions in this module share a common structure. The first argument is a [`pulp_smash.config.ServerConfig`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.config.html#pulp_smash.config.ServerConfig), and all other arguments correspond to command-line options. Most arguments are named after a flag. For example, an argument `to_repo_id` corresponds to the flag `--to-repo-id`.
>
> For the meaning of each argument, see pulp-admin.

Individual parameters could be documented, but they will be redundant:

```python
# Yes, this function has an annoying number of arguments. It may be better to
# adopt some other solution such as providing a string for interpolation.
def repo_create(  # pylint:disable=too-many-arguments
        server_config,
        enable_v1=None,
        enable_v2=None,
        feed=None,
        repo_id=None,
        upstream_name=None):
    """Execute ``pulp-admin docker repo create``.

        Pulp server being targeted.
    :param enable_v1: A value for ``--enable-v1``.
    :param enable_v2: A value for ``--enable-v2``.
    :param feed: A value for ``--feed``.
    :param repo_id: A value for ``--repo-id``.
    :param upstream_name: A value for ``--upstream-name``.
    """
```

It's fine to have this sort of documentation, I suppose. Submit a patch if you'd really like to have them in place.

rbarlow:
The CLI raising an Exception when pulp-admin returns a non-0 exit code goes against a principle you were pushing in this pull request. In this pull request, you asserted the importance of having only one assertion per method so that when tests fail we can see the full list of failed assertions (i.e., no previous failure causes a later failure or success to be masked). However, the combination of having the CLI raise an Exception when encountering non-zero exit codes, and having interactions with Pulp in setUpClass() methods means that no assertions will happen when there are bugs in Pulp that generate a non-zero exit code in pulp-admin. I don't personally care either way, but it is frustrating that you made me jump through so many hoops here and then didn't apply that same principle to your code.

I wasn't personally interested in the docblocks, but I found it surprising that you placed a demand on their existence but then didn't apply that same demand to your commits, with a message about how you suppose it's fine to have them. Again this is frustrating since you made me jump through these hoops, but you lowered the standard for yourself.

Ichimonji10:
> I wasn't personally interested in the docblocks, but I found it surprising that you placed a demand on their existence but then didn't apply that same demand to your commits, with a message about how you suppose it's fine to have them. Again this is frustrating since you made me jump through these hoops, but you lowered the standard for yourself.

The functions originally added by this pull request did not have a uniform logic to them. Each method had a unique signature, where the name of each method varied, and where each argument might or might not be required or be given a value. In such a situation, documenting each function individually is necessary.

The utility functions recently added have a uniform logic to them. Each method is named after the pulp-admin command executed. Each argument is named after the corresponding pulp-admin flag, and no argument is given a default value. In such a situation, a single description suffices.

Ichimonji10:
> The CLI raising an Exception when pulp-admin returns a non-0 exit code goes against a principle you were pushing in this pull request [...] of having only one assertion per method so that when tests fail we can see the full list of failed assertions

Test code has the dual responsibilities of providing a rich set of tests while not being cumbersome. Checking for non-zero return codes and ensuring that asynchronous tasks complete successfully are two examples of checks that should be done in almost all tests, but which are also highly redundant. Using `pulp_smash.cli.code_handler` to automatically perform these checks has several benefits:

* Less code needs to be written to create a test that performs a given set of assertions, which encourages the test writer to concentrate on making an interesting test instead of getting bogged down in details.
* Having less code lowers the maintenance burden of this test suite.
* The automatic checks are applied even in cases where a human might forget, or where it would be especially kludgy to save the relevant responses.
* The automatic checks are all identical, which helps prevent human errors from causing or hiding a bug.

Given these significant benefits, I do use the automatic checks where possible. (See: #122.) In addition, I write somewhat-overlapping sets of tests.

> it is frustrating that you made me jump through so many hoops here and then didn't apply that same principle to your code.

The tests added in 012fa3a81b6660b3beece9c6125fe03353ff80d3. Check it out - no assertion failure will ever prevent another assertion from executing. Further, nearly every test method already in pulp smash has an isolated set of assertions.

Issue #72, opened 2016-01-27
Automate Lazy sync  - background


preethit:
Background policy skip downloading files during sync, but once the sync is completed it will dispatch a task to download the file
Case 2
1. Create a repo with background download policy
pulp-admin rpm repo create --repo-id zoo --feed https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/  --download-policy background
2. Run repo Sync
 pulp-admin rpm repo sync run --repo-id zoo
3. Assert that there is a pulp task for downloading files
  pulp-admin tasks list

jeremycline:
I can't assign this to myself, but I'm going to work on this.

Ichimonji10:
Fixed by 12bb39ee7eb959c252d4217b0d70b3338833f182

PR #130, opened 2016-02-23
Add tests for `background` lazy sync.

jeremycline:
This adds tests for creating and syncing a repository with the
`background` download policy. It asserts that a download task is
dispatched, all repository units are reported as being local, and
that the RPM is served directly by Pulp.

Fixes issue #72

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -524,3 +524,86 @@ class SyncOnDemandTestCase(utils.BaseAPITestCase):
     def test_same_rpm_cache_header(self):
         """Assert the second request resulted in a cache hit from Squid."""
         self.assertIn('HIT', self.same_rpm.headers['X-Cache-Lookup'])
+
+
+class SyncBackgroundTestCase(utils.BaseAPITestCase):
+    """Assert Yum supports background syncing."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create an RPM repository with a valid feed and sync it.
+
+        Do the following:
+
+        0. Reset Pulp, including the Squid cache.
+        1. Sync/publish a repository using the 'background' download policy.
+        2. Assert all the content is downloaded by a background task after the
+           sync.
+        """
+        super(SyncBackgroundTestCase, cls).setUpClass()
+        if cls.cfg.version < Version('2.8'):
+            raise unittest2.SkipTest('This test requires Pulp 2.8 or greater.')
+
+        # Required to ensure content is actually downloaded.
+        utils.reset_squid(cls.cfg)
+        utils.reset_pulp(cls.cfg)
+
+        client = api.Client(cls.cfg, api.json_handler)
+
+        # Create the repository
+        body = _gen_repo()
+        body['importer_config'] = {
+            'feed': RPM_FEED_URL,
+            'download_policy': 'background',
+        }
+        distributor = _gen_distributor()
+        distributor['auto_publish'] = True
+        distributor['distributor_config']['relative_url'] = body['id']
+        body['distributors'] = [distributor]
+
+        repo = client.post(REPOSITORY_PATH, body)
+        cls.resources.add(repo['_href'])
+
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I haven't reviewed this test in full, but I did `git checkout` the code and run the tests, and I see they're skipped when run against Pulp 2.7. Thanks for taking care of that. :smile:

Ichimonji10:
I ran these tests from within freshly provisioned Pulp 2.7 and 2.8 VMs, and the new tests fail on 2.8. Can you address this issue, either by fixing the bug or skipping the test based on the state of an open bug?

```
(pulp-smash)[vagrant@dev pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_sync_publish.SyncBackgroundTestCase
FF..
FAIL: test_repo_local_units (pulp_smash.tests.rpm.api_v2.test_sync_publish.SyncBackgroundTestCase)
Assert that all content is downloaded for the repository.
  File "pulp_smash/tests/rpm/api_v2/test_sync_publish.py", line 583, in test_repo_local_units
    sum(self.repo['content_unit_counts'].values()),
AssertionError: 7 != 39

FAIL: test_request_history (pulp_smash.tests.rpm.api_v2.test_sync_publish.SyncBackgroundTestCase)
Assert that the request was serviced directly by Pulp.
  File "pulp_smash/tests/rpm/api_v2/test_sync_publish.py", line 593, in test_request_history
    self.assertEqual(0, len(self.rpm.history))
AssertionError: 0 != 1

Ran 4 tests in 63.992s

FAILED (failures=2)
```

jeremycline:
The tests are likely failing because the Pulp server is not configured to trust the certificate of its Apache web server. This can be ignored by forcing the requests to be HTTP, which is a configuration option in `/etc/pulp/server.conf`'s `[lazy]` section. If you set `https_retrieval: false` the tests should pass.

Ichimonji10:
Yes, making that change does allow tests to pass.

Is this something users should have to worry about? As a user of the Vagrantfile supplied by Pulp, I find it surprising that this additional step must be taken in order to make Pulp work. As someone who wants the Jenkins tests to pass, I worry that merging these tests will cause additional failures. And as a Pulp user, this goes way over my head and seems like an implementation detail. (I did read through https://github.com/pulp/pulp/blob/master/docs/dev-guide/design/deferred-download.rst). Is there some solution available to address this issue?

jeremycline:
As a Vagrant user, probably not. My personal preference is that the Vagrant box set up the certificates appropriately, but I haven't gotten around to it yet.

Jenkins will need to be adjusted in a similar manner (either by not using HTTPS or setting up certificates correctly).

As a user, this absolutely is something you have to worry about. I know secure by default isn't fun, but I'm a firm believer in it.

Ichimonji10:
> As a user, this absolutely is something you have to worry about. I know secure by default isn't fun, but I'm a firm believer in it.

I think it's time I pick up a book on SSL. <_<

> My personal preference is that the Vagrant box set up the certificates appropriately, but I haven't gotten around to it yet.

Makes sense. I'll just make this change manually in the mean time.

> Jenkins will need to be adjusted in a similar manner

OK, good to know. I'd like to make sure that new tests don't cause failures in Jenkins. */me marks an item on his to-do list.*

jeremycline:
@Ichimonji10 I'm working on adding a trusted CA and signing certificates for various Pulp services in our Ansible playbook for Vagrant, so it should be as simple as dragging and dropping into the Ansible playbook for Jenkies

Ichimonji10:
Awesome! I'm excited for that to happen.

jeremycline:
@Ichimonji10 I've made a PR for the pulp_packaging repo: https://github.com/pulp/pulp_packaging/pull/101

elyezer:
I will review this PR as https://github.com/pulp/pulp_packaging/pull/101 is merged.

elyezer:
Merged as 12bb39ee7eb959c252d4217b0d70b3338833f182, added an additional 30914b819c240275b7f91e5a60e30f8210deddef.

PR #139, opened 2016-02-26
Add assertions testing the Docker v2 publish results.

rbarlow:
This commit adds several assertions to an existing sync/publish
CLI test case that ensure that the result of the pulp_docker v2
publish has the expected data that the Docker client will need.


rbarlow:
These tests are for https://pulp.plan.io/issues/1051.

pulp_smash/tests/docker/cli/test_sync_publish.py
```diff
@@ -0,0 +1,281 @@
...
+        string.
+        """
+        manifest = self._get_manifest()
+        self.assertTrue(isinstance(manifest['fsLayers'], list))
+        self.assertTrue(
+            all([isinstance(l, dict) for l in manifest['fsLayers']]))
+        self.assertTrue(
+            all([l.keys() == ['blobSum'] for l in manifest['fsLayers']]))
+        self.assertTrue(
+            all([isinstance(l['blobSum'], (str, unicode))
```
Ichimonji10: This fails on Python 3.

```
>>> isinstance('foo', (str, unicode))
NameError: name 'unicode' is not defined
```

How about this?

```python
isinstance('foo', type(''))
```

rbarlow: Yeah I noticed that on Friday but it was 5 PM so I figured I'd save it for Monday. The type solution seems good. I was considering using the six library to fix this. Do you prefer your technique? It does avoid adding a dependency so that's good.

Ichimonji10: Six would work well, and I certainly respect it as a lib. But it does seem excessive to add six as a dependency when there's so few use cases for it. That's the same reason I haven't added a random data generation lib - `uuid4()` and a few `random.randint()` calls have worked well enough.

rbarlow: @Ichimonji10 I have made the change you suggested, and the tests now pass.

Ichimonji10:
Merged as 423cb9ab445f300505e977923261e700b75c8979, with follow-up commits of 06d96a92a8b5cef1ab8b2e8a4f11fef7a0fecd8d and 44b802a091a2bfb4935ff4e27841bb1841bd349a. Thank you, @rbarlow!

PR #144, opened 2016-02-29
Add redundant publish metadata test

### LINKED ISSUE #127: Assert that two publishes do not modify the RPM repodata
Test setup:
- Create a repo and sync+publish it (ie: zoo)

The test:
- Get the contents of repodata from the published repo
- publish zoo a second time
- Get the contents of repodata from the published repo now

Assert that all the metadata files and their filenames are identical.

If Pulp is publishing a second time, when there is nothing to do, it should do nothing. In the past a republish caused metadata to get republished in all cases which changes the checksum included in the filename. That was undesirable since the repodata itself didn't change the checksums shouldn't change.

This was believed to already be fixed so there isn't an open issue. For now this guarantee is only expected to be on RPM content, but I think it makes sense for all content types.

pcreech:
Test that metadata creation is identical before and after a redundant
publish

closes #127

Depends on https://pulp.plan.io/issues/1724

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -501,3 +501,99 @@ class SyncOnDemandTestCase(utils.BaseAPITestCase):
     def test_same_rpm_cache_header(self):
         """Assert the second request resulted in a cache hit from Squid."""
         self.assertIn('HIT', self.same_rpm.headers['X-Cache-Lookup'])
+
+
+class RedundantPublishSameMetadata(utils.BaseAPITestCase):
+    """Redundantly publish a repo without changes."""
+
+    @classmethod
+    def _download_repomd_xml(cls, url):
+        cls.client.response_handler = api.safe_handler
+        result = cls.client.get(url)
+        cls.client.response_handler = api.json_handler
+        return result
+
+    @classmethod
+    def setUpClass(cls):
+        """Test a redundant publish on a repo.
+
+        Do the following:
+
+        1. Create a new repo with zoo's feed
+        2. Publish the repo a first time
+        3. Publish the repo a second time, without any changes
+        """
+        super(RedundantPublishSameMetadata, cls).setUpClass()
+        if selectors.bug_is_untestable(1724):
+            raise unittest2.SkipTest('https://pulp.plan.io/issues/1724')
+
+        cls.repo_id = utils.uuid4()
+        relative_url = urljoin('repos/', cls.repo_id)
+        feed = 'https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/'
+
+        repomd_xml = '/pulp/repos/{relative_url}/repodata/repomd.xml'\
+            .format(relative_url=relative_url)
+
+        sync_data = {'override_config': {}}
+
+        publish_data = {'override_config': {}, 'id': 'yum_distributor'}
+
+        body = {
+            'display_name': 'TwoPublishes test repo',
+            'description': 'TwoPublishes test repo - delete me',
+            'distributors': [
+                {
+                    'distributor_id': 'yum_distributor',
+                    'auto_publish': False,
+                    'distributor_config': {
+                        'http': False,
+                        'relative_url': relative_url,
+                        'https': True
+                    },
+                    'distributor_type_id': 'yum_distributor'
+                }
+            ],
+            'notes': {'_repo-type': 'rpm-repo'},
+            'importer_type_id': 'yum_importer',
+            'importer_config': {
+                'feed': feed
+            },
+            'id': cls.repo_id
+        }
+
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        try:
+            repo = cls.client.post(REPOSITORY_PATH, body)
+
+            sync_path = urljoin(repo['_href'], 'actions/sync/')
+            publish_path = urljoin(repo['_href'], 'actions/publish/')
+
+            cls.client.post(sync_path, sync_data)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish1_repomd = cls._download_repomd_xml(repomd_xml)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish2_repomd = cls._download_repomd_xml(repomd_xml)
+
+        except:
+            cls.tearDownClass()
+            raise
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository we just created."""
+        if hasattr(cls, 'client'):
+            try:
+                cls.client.delete(urljoin(REPOSITORY_PATH, cls.repo_id))
+            except:  # pylint: disable=bare-except
+                pass
+
+    def test_compare_repomd(self):
+        """Assert the repodata is the same."""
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+            cls.tearDownClass()
+            raise
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository we just created."""
+        if hasattr(cls, 'client'):
+            try:
+                cls.client.delete(urljoin(REPOSITORY_PATH, cls.repo_id))
+            except:  # pylint: disable=bare-except
```
Ichimonji10: Pylint is correct to warn about this bare `except` clause. This is dangerous.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+        except:
+            cls.tearDownClass()
+            raise
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository we just created."""
+        if hasattr(cls, 'client'):
+            try:
+                cls.client.delete(urljoin(REPOSITORY_PATH, cls.repo_id))
```
Ichimonji10: The parent `BaseAPITestCase` defines a `tearDownClass` method, and that method already provides resource deletion logic. How about using that logic instead of redefining it here?

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+            cls.client.post(sync_path, sync_data)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish1_repomd = cls._download_repomd_xml(repomd_xml)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish2_repomd = cls._download_repomd_xml(repomd_xml)
+
+        except:
+            cls.tearDownClass()
```
Ichimonji10: `tearDownClass` is automatically called by the unittest framework. Is there a compelling reason to call it here?

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+
+            cls.client.post(sync_path, sync_data)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish1_repomd = cls._download_repomd_xml(repomd_xml)
+
+            cls.client.post(publish_path, publish_data)
+            cls.publish2_repomd = cls._download_repomd_xml(repomd_xml)
+
+        except:
```
Ichimonji10: Please no bare `except` clauses.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+        2. Publish the repo a first time
+        3. Publish the repo a second time, without any changes
+        """
+        super(RedundantPublishSameMetadata, cls).setUpClass()
+        if selectors.bug_is_untestable(1724):
+            raise unittest2.SkipTest('https://pulp.plan.io/issues/1724')
+
+        cls.repo_id = utils.uuid4()
+        relative_url = urljoin('repos/', cls.repo_id)
+        feed = 'https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/'
```
Ichimonji10: See `pulp_smash.constants`.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+
+        repomd_xml = '/pulp/repos/{relative_url}/repodata/repomd.xml'\
+            .format(relative_url=relative_url)
+
+        sync_data = {'override_config': {}}
+
+        publish_data = {'override_config': {}, 'id': 'yum_distributor'}
+
+        body = {
```
Ichimonji10: See `pulp_smash.tests.rpm.api_v2.utils`.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
     def test_same_rpm_cache_header(self):
         """Assert the second request resulted in a cache hit from Squid."""
         self.assertIn('HIT', self.same_rpm.headers['X-Cache-Lookup'])
+
+
+class RedundantPublishSameMetadata(utils.BaseAPITestCase):
+    """Redundantly publish a repo without changes."""
+
+    @classmethod
+    def _download_repomd_xml(cls, url):
```
Ichimonji10: Is this method necessary?

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -501,3 +501,99 @@ def test_same_rpm_checksum(self):
...
+                pass
+
+    def test_compare_repomd(self):
+        """Assert the repodata is the same."""
+        if selectors.bug_is_untestable(1724):
+            self.skipTest('https://pulp.plan.io/issues/1724')
+
+        self.assertEqual(self.publish1_repomd.content,
+                         self.publish2_repomd.content,
+                         'Expected identical metadata on redundant publish')
```
Ichimonji10: Consider putting this information into the docstring, so that there's a single good authoritative source of information about what this method does. Many test runners will at least emit the docstring summary line for failed tests.

Ichimonji10:
If you don't have time to review the comments I've made, that's OK. Just tell me, and I'll eventually write a follow-up commit myself.

pcreech:
@Ichimonji10 Thanks for the review!  I should be able to address them in a timely manner.  Will get back to you soon with updates

Ichimonji10:
@pcreech Thank you. :smile:

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
omaciel: @pcreech if want to make sure that a 'clean up' method is run in case the `setUp` methods fails, you can use a `addCleanup` method like [this](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/test.py#L151)

Ichimonji10: :+1: There's a couple options. I notice that you're calling `tearDownClass` in order to force deletion of a repository. If you have the repository href (you should), then do this somewhere in `setUpClass`:

```python
self.resources.add(repo_href)
```

...and the default `tearDownClass` will take care of things for you.

elyezer:
Merged as 48943e21ad3373349214cc59ddd4275ea1b23938, added an additional commit 7b939bafe0bd7baf8308a2c4a2db4fbbb4043629 to handle the comments.

Ichimonji10:
:+1:

PR #164, opened 2016-03-11
Move dynamic imports to compat module

asmacdo:
This commit creates a compat module that can be used to do any dynamic
importing (particularly to support both python 2 and 3) so it will not
need to be done everywhere.


Ichimonji10:
Great job! Thank you. Merged as af40b0119180ccb956d6250c4f179e385becd657.

PR #168, opened 2016-03-18
- configurable db name

midnightercz:
I thought it could be handy to have choice of configuring the db name.

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -63,6 +63,7 @@ class ServerConfig(object):  # pylint:disable=too-many-instance-attributes
     ...     auth=('username', 'password'),
     ...     verify=False,  # Disable SSL verification
     ...     version='2.7.5',
+    ...     pulp_database='pulp_database'
     ... )
     >>> response = requests.post(
     ...     cfg.base_url + '/pulp/api/v2/actions/login/',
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Can you explain this change, preferably in the commit message? There may be other positive changes that should be made to this pull request, but I don't know, as I don't know what's happening here.

Thank you in advance. :smile:

pulp_smash/utils.py
```diff
@@ -67,7 +67,8 @@ def reset_pulp(server_config):
     # Reset the database and nuke accumulated files.
     client = cli.Client(server_config)
     prefix = '' if client.run(('id', '-u')).stdout.strip() == '0' else 'sudo '
-    client.run('mongo pulp_database --eval db.dropDatabase()'.split())
+    mongo_drop_cmd = 'mongo {pulp_database} --eval db.dropDatabase()'
+    client.run((mongo_drop_cmd % server_config.pulp_database).split())
```
Ichimonji10: Does this even work?

```
>>> 'foo {bar}' % 'biz'
TypeError: not all arguments converted during string formatting
```

pulp_smash/config.py
```diff
@@ -135,11 +136,13 @@ def __init__(  # pylint:disable=too-many-arguments
             auth=None,
             verify=None,
             version=None,
+            pulp_database=None,
             cli_transport=None):
         """Initialize this object with needed instance attributes."""
         self.base_url = base_url
         self.auth = auth
         self.verify = verify
+        self.pulp_database = pulp_database
```
Ichimonji10: If the user doesn't provide a value, this attribute will be `None`, which seems wrong.

Ichimonji10:
Is there some specific test case which requires that we be able to choose the Pulp database name? Assuming there is, does the database name need to be in such a central and critical bit of code, or can it live in a specific test case? Unused code is a liability, and doubly so when it touches as much of the application as the `ServerConfig` class.

Ichimonji10:
I don't see any use case for this change at this time. Closing.

PR #169, opened 2016-03-21
Introduce tests covering yum distributor XML generation

rohanpm:
These tests exercise the yum distributor and importer.

The tests are currently limited in scope and only verify a handful of
properties on the generated XML.


rohanpm:
As well as the CI failure ... I'm reconsidering whether pulldom was the right XML API to use.

pulp_smash/tests/rpm/api_v2/utils.py
```diff
@@ -42,3 +46,82 @@ def sync_repo(server_config, href):
...
+        text = gzip.decompress(response.content)  # pylint: disable=no-member
+        text = str(text, 'utf-8')
+        return ElementTree.fromstring(text)
+    elif is_gzip:
+        # python 2
+        compressed = io.BytesIO(response.content)
+        decompressed = gzip.GzipFile(fileobj=compressed)
+        return ElementTree.parse(decompressed)
+    else:
+        return ElementTree.fromstring(response.text)
```
rohanpm: I think this is wrong.  I've found that requests may wrongly guess ISO-8859-1 as the encoding here.  Looks like it may be better to use the bytes in response.content instead and let ElementTree decide the encoding based on what's declared in the XML.

Similarly maybe line 75 should be removed.

pulp_smash/tests/rpm/api_v2/utils.py
```diff
@@ -42,3 +46,82 @@ def sync_repo(server_config, href):
...
+    if is_gzip and 'decompress' in dir(gzip):
+        # python 3
+        text = gzip.decompress(response.content)  # pylint: disable=no-member
+        text = str(text, 'utf-8')
+        return ElementTree.fromstring(text)
+    elif is_gzip:
+        # python 2
+        compressed = io.BytesIO(response.content)
+        decompressed = gzip.GzipFile(fileobj=compressed)
+        return ElementTree.parse(decompressed)
```
rohanpm: This should be changed because parse() returns an ElementTree and fromstring() returns an Element, which have similar but not the same API.  It should return the same type of object no matter the code path.

Ichimonji10: I was wondering about that when reviewing this yesterday. Thanks for the clarification.

pulp_smash/tests/rpm/api_v2/test_repomd.py
```diff
@@ -0,0 +1,60 @@
...
+        repo_url = urljoin('/pulp/repos/',
+                           distribute['config']['relative_url'])
+        repomd_url = urljoin(repo_url, 'repodata/repomd.xml')
+
+        client = api.Client(cls.cfg, xml_handler)
+        cls.repomd_tree = client.get(repomd_url)
+
+    def test_repomd_sanity(self):
+        """Test basic top-level structure of the generated repomd.xml file."""
+        namespace = 'http://linux.duke.edu/metadata/repo'
```
Ichimonji10: What's the significance of this URL? I'm unable to look into it myself, as it 404s.

pulp_smash/tests/rpm/api_v2/test_repomd.py
```diff
@@ -0,0 +1,60 @@
...
+from __future__ import unicode_literals
+
+from pulp_smash import api
+from pulp_smash.compat import urljoin
+from pulp_smash.constants import REPOSITORY_PATH
+from pulp_smash.tests.rpm.api_v2.utils import gen_repo, gen_distributor, \
+    BaseRepoMDTestCase, xml_handler
+
+
+class RepoMDTestCase(BaseRepoMDTestCase):
```
Ichimonji10: Is there any special reason to inherit from `BaseRepoMDTestCase`? As far as I can tell, this class uses nothing from that parent, and it can inherit from `BaseAPITestCase` instead.

pulp_smash/tests/rpm/api_v2/test_repomd.py
```diff
@@ -0,0 +1,60 @@
...
+    """Test repomd.xml generation via yum distributor."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Creates, publishes repo and fetches repomd.xml."""
+        super(RepoMDTestCase, cls).setUpClass()
+
+        cls.tasks = {}
+        cls.errata = {}
```
Ichimonji10: These class attributes are unused.

pulp_smash/tests/rpm/api_v2/test_updateinfo.py
```diff
@@ -0,0 +1,212 @@
...
+            {'id': distribute['id']})
+
+        repo_url = urljoin('/pulp/repos/',
+                           distribute['config']['relative_url'])
+        cls.updateinfo_tree = cls.get_repodata_xml(repo_url, 'updateinfo')
+
+    def assert_task_successful(self, task, msg):
+        """Assert that a task completed successfully.
+
+        This is more extensive than the checks built-in to the api Client,
```
Ichimonji10: This test is actually less extensive than what is done by `pulp_smash.api._handle_202`. It does, however, make a fine *additional* check.

Ichimonji10:
There's no need to respond to my comments by amending the commit in this PR. That's because I've put together a significant set of changes that I'll add as an additional commit when merging, and having the base PR change makes that harder to do. However, I'm likely to ask for more back-and-forth on future PRs, so please read the updated modules after the merge. (I'm fast-tracking this PR to ensure RCM has some basic coverage.)

Ichimonji10:
Merged as ffc3d90ebaa3d83567148ae244b2b31e788b316a, with a follow-up commit of 771c0b2068629a75e6de53ba6c1cfc6ac746fd30. Thank you, @rohanpm.

PR #170, opened 2016-03-22
- added test last_unit_removed

midnightercz:


requirements-dev.txt
```diff
@@ -13,3 +13,6 @@ sphinx
 
 # For `make package`
 wheel
+
+# For datetime tests
+isodate
```
Ichimonji10: The `setup.py` file lists dependencies that are necessary for Pulp Smash to function. In contrast, `requirements-dev.txt` lists dependencies that are necessary for extra tasks, such as generating documentation with Sphinx or running the linters. Given that this new requirements is necessary for Pulp Smash to function, please place it in setup.py.

Ichimonji10: Along those lines, is there any special reason why isodate was chosen? It seems like a decent library, with roughly 2,800 downloads per day and decent compatibility with various Python versions. However, there are other more popular libraries which target the same problem area. For example, python-dateutil has roughly 232,000 downloads per day, is compatible with Python 3.5, and has been around for ~6 years longer.

Even more fundamentally, do we need a new dependency at all? Does Pulp state whether it complies with ISO 8601, or the more liberal RFC 3339, or something else?

Ichimonji10:
There are already tests for unit disassociation in [`pulp_smash.tests.rpm.api_v2.test_unassociate`](http://pulp-smash.readthedocs.org/en/latest/api/pulp_smash.tests.rpm.api_v2.test_unassociate.html). Is there a particular reason to add a new module instead of adding a new test case to that module? Assuming a new module is desired, I'd appreciate if a new documentation stub is added. While not strictly necessary, doing so will cause Sphinx to parse the docstrings in this new module, which often reveals errors.

Ichimonji10:
Finally, can I please get a more thorough commit message and/or docstrings? I don't usually block on this point, but I really am unclear on what this commit is trying to accomplish. The module-wide docstring states that one of the test steps is to "assert unit X is not accessible", the one test method states that it will "verify the HTTP status code of each publish," and the one assertion compares a pair of time stamps, the commit message just says `- added test last_unit_removed`, etc.

Ichimonji10:
Work in progress: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:170

Ichimonji10:
Merged as ea811390f85899dc79f3722402ecb4f208bc2ace, with a follow-up commit of c9a2d0f422b43817a312b4dfddd34adfadbe4dd9. Thank you, @midnightercz.

PR #175, opened 2016-03-25
Respect target_platform_release bug tracker field

### LINKED ISSUE #173: Improvements to test skipping logic
Redmine contains a "target platform release" field. Smash's `bug_is_untestable` logic should be aware of this field so bugs that have been moved to `ON_QA` or `VERFIED` are not run on older versions of Pulp.

An additional improvement in this area would be to get a dump of all bugs that are now testable and their target platform release to avoid making another http request to redmine for each bug.

There are some potential issues with this method. First are bugs that are fixed in an older Z release. For example, a bug that is fixed in 2.7.3 and 2.8.1. For the example lets say that 2.7.3 is released before 2.8.1. We then know that bugfixes in 2.7.3 are also available in 2.8.1. We should be able to accurately determine whether a bugfix is present by knowing the target release and the order of the releases. 

A second option would be to use `--contains` on the commit. I think that this would require less logic and could potentially be more accurate (and would work for backports), but would require very strict book keeping in redmine, specifically making sure that every bug has a commit. We would need to explore what is necessary when there are more than 1 commit with this model.

Ichimonji10:
Change the signatures of `bug_is_testable` and `bug_is_untestable` in module `pulp_smash.selectors`, from this:

```python
bug_is_testable(bug_id)
bug_is_untestable(bug_id)
```

To this:

```python
bug_is_testable(bug_id, pulp_version)
bug_is_untestable(bug_id, pulp_version)
```

The new `pulp_version` parameter is "A ``packaging.version.Version`` object telling the version of the Pulp server we are testing." It lets the function determine whether a bug is testable or not based not only on the state of a bug, but the Pulp version in which the bug was fixed relative to the Pulp version we're testing against. Sample usage:

```python
>>> from pulp_smash.selectors import bug_is_testable
>>> bug_is_testable(1743, Version('2.7.0'))
False
>>> bug_is_testable(1743, Version('2.7.99999'))
False
>>> bug_is_testable(1743, Version('2.8'))
True
>>> bug_is_testable(1743, Version('2.8.0'))
True
>>> bug_is_testable(1743, Version('2.8.0.0'))
True
>>> bug_is_testable(1743, Version('2.8.1'))
True
>>> bug_is_testable(1743, Version('3'))
True
```

If a bug's "Target Platform Release" field is unset, the field is effectively ignored. [1] For example:

```python
>>> from pulp_smash.selectors import bug_is_testable
>>> bug_is_testable(1744, Version('2'))
True
>>> bug_is_testable(1744, Version('3'))
True
```

Test suite results do not change as a result of this commit. Results generated with:

```sh
```

Fix #173.

[1] More precisely, we assume that the bug has been fixed in Pulp version 0.

### DIFF
--- a/pulp_smash/selectors.py
+++ b/pulp_smash/selectors.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import warnings
+from collections import namedtuple
 from functools import wraps
 
 import requests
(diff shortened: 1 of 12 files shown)


docs/api/pulp_smash.selectors.rst
```diff
@@ -4,3 +4,4 @@
 Location: :doc:`/index` → :doc:`/api` → :doc:`/api/pulp_smash.selectors`
 
 .. automodule:: pulp_smash.selectors
+    :private-members:
```
Ichimonji10: Ehhh, I need to drop this.

Ichimonji10:
This pull request looks pretty gnarly at first glance, but it's not as bad as it seems. Twelve files are modified, but ten of those twelve are straightforward updates in response to changed method signatures. The simple cases look like this:

```diff
-        if selectors.bug_is_untestable(1710):
+        if selectors.bug_is_untestable(1710, cls.cfg.version):
```

The complicated cases look like this:

```diff
-        if (self.cfg.version >= Version('2.8') and
-                selectors.bug_is_untestable(1455)):
+        if selectors.bug_is_untestable(1455, self.cfg.version):
```

It's valid to roll these two expressions together because [Pulp issue 1455](https://pulp.plan.io/issues/1455) has its Target Platform Release field set to "2.8.0".

Of the two remaining files, one is `pulp_smash.selectors`, and the other is `tests.test_selectors`. The former module contains the real meat of the changes in this pull request, and the latter module contains unit tests for the former.

pulp_smash/selectors.py
```diff
@@ -32,62 +33,122 @@
...
+    bug_json = response.json()
+    _BUG_STATUS_CACHE[bug_id] = _Bug(
+        bug_json['issue']['status']['name'],
+        _convert_tpr(_get_tpr(bug_json)),
+    )
     return _BUG_STATUS_CACHE[bug_id]
 
 
-def bug_is_testable(bug_id):
+def bug_is_testable(bug_id, pulp_version):
```
elyezer: I'm wondering if wouldn't be easier to accept a string here as `pulp_version` and convert to `packaging.version.Version` internally in order to do the comparison.

I'm saying that because passing a string would be easy to write, since there is no need to the caller wrap into the version object and will not require an additional import. That said we would reduce the amount of code needed in order to use the functionality.

elyezer: Seems that I already got the answer for my question. The Version is being used everywhere and the Pulp Smash settings already returns a Version wrapped object. Changing this will require more refactoring and will not be so valuable to do.

That said let's follow the standard and always expect a `packaging.version.Version` object, this way one may not be confused if a string or a Version object should be the argument.

elyezer:
LGTM

pulp_smash/selectors.py
Ichimonji10: > Seems that I already got the answer for my question.

That's 100% correct.

> That said let's follow the standard and [...]

Can you expand on this?

Ichimonji10:
I re-ran the test suite under Python 2 and 3 against fresh Pulp 2.8.1 systems, and all tests still pass.

Ichimonji10:
Checked in with @elyezer one more time, and all seems OK. Want to make sure I don't get this wrong.

PR #176, opened 2016-03-25
Rewrite "Installation" documentation section

Ichimonji10:
Do not attempt to lay out a comprehensive set of installation strategies with wordy explanations. Instead, list a recipe for creating a virtualenv, followed by three strategies for installing Pulp Smash, followed by a link to the official Python documentation.

Drop a redundant set of instructions in the "About" page, and reference the recipes in the "Installation" page instead.

This change is made on the theory that those people who need instructions for installing a Python package will benefit from a concise recipe-oriented approach, and those people who are interested in more in-depth explanations of Python packaging are (should be!) looking elsewhere.

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -54,14 +54,9 @@ Contributing
 Contributions are encouraged. The easiest way to contribute is to submit a pull
 request on GitHub, but patches are welcome no matter how they arrive.
 
-You can create a development environment and verify its sanity like so::
-
-    virtualenv env  # or `python -m virtualenv env` is using Python >= 3.3
-    source env/bin/activate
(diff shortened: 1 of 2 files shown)


Ichimonji10:
@preethit, I think your perspective on this would be especially valuable.

preethit:
@Ichimonji10  Adding .. _virtualenv: http://virtualenv.readthedocs.org/en/latest/ would be beneficial too. But other than that, looks good to me.

Ichimonji10:
Done. Thank you.

elyezer:
Got late for the merge party, but I had already looked over this PR and it looks good to me.

PR #179, opened 2016-03-29
Add function pulp_smash.utils.is_root

Ichimonji10:
When called, this new function will "tell if we are root on the target
system." Test suite results don't change as a result of this commit.
Test results generated with:

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_download.py
+++ b/pulp_smash/tests/rpm/api_v2/test_download.py
@@ -20,13 +20,6 @@ from pulp_smash.constants import REPOSITORY_PATH, RPM_ABS_PATH, RPM_FEED_URL
 from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
 
 
-def _is_root(server_config):
-    """Tell if we are root on the target Pulp system."""
-    if cli.Client(server_config).run(('id', '-u')).stdout.strip() == '0':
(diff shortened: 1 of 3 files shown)


Ichimonji10:
@elyezer Mind eyeballing this?

pulp_smash/tests/rpm/api_v2/test_download.py
```diff
@@ -84,7 +77,7 @@ def setUpClass(cls):
 
         # Corrupt an RPM. The file is there, but the checksum isn't right.
         cli_client = cli.Client(cls.cfg)
-        sudo = '' if _is_root(cls.cfg) else 'sudo '
+        sudo = '' if utils.is_root(cls.cfg) else 'sudo '
```
elyezer: Seeing this sudo/prefix defined in more than one place made me wonder if `utils.sudo()` would be a good thing to have.

elyezer: Off course this could be added in future commits.

elyezer: Actually the sudo stuff sounds more like a CLI client feature.

Ichimonji10: /me makes a mental note. :+1:

elyezer:
LGTM

Issue #167, opened 2016-03-16
Three pulp-smash tests seem to be fragile about hostnames


rbarlow:
I have been running pulp-smash on an F23 system pointed at a Fedora Rawhide system that's running the offical Fedora packages of Pulp. All the tests pass except for three, which fail due to hostname mismatching in some cases, and in other cases due to one half of the assertion expecting a hostname and the other not. Here's example test output, but I've sanitized the hostnames to protect the innocent:

```
(pulp-smash)[fedora@rbarlow-test-dev-setup pulp-smash]$ time py.test pulp_smash/tests/
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.4.3, pytest-2.9.0, py-1.4.31, pluggy-0.3.1
rootdir: /home/fedora/devel/pulp-smash, inifile: 
collected 211 items 

pulp_smash/tests/docker/api_v2/test_crud.py ....
pulp_smash/tests/docker/cli/test_copy.py ............
pulp_smash/tests/docker/cli/test_crud.py ..........
pulp_smash/tests/docker/cli/test_sync_publish.py ......F.............s..
pulp_smash/tests/ostree/api_v2/test_sync_publish.py ..................
pulp_smash/tests/platform/api_v2/test_consumer.py ..
pulp_smash/tests/platform/api_v2/test_content_applicability.py .......
pulp_smash/tests/platform/api_v2/test_login.py ..s.
pulp_smash/tests/platform/api_v2/test_repository.py .F..s........
pulp_smash/tests/platform/api_v2/test_search.py ......ssss......
pulp_smash/tests/platform/api_v2/test_user.py ...............
pulp_smash/tests/puppet/api_v2/test_sync_publish.py .....................
pulp_smash/tests/rpm/api_v2/test_broker.py ..
pulp_smash/tests/rpm/api_v2/test_duplicate_uploads.py .
pulp_smash/tests/rpm/api_v2/test_iso_crud.py .F.........
pulp_smash/tests/rpm/api_v2/test_orphan_remove.py ..
pulp_smash/tests/rpm/api_v2/test_remove_unit.py ....
pulp_smash/tests/rpm/api_v2/test_retain_old_count.py ..
pulp_smash/tests/rpm/api_v2/test_schedule_sync.py ...........
pulp_smash/tests/rpm/api_v2/test_retain_old_count.py ..
pulp_smash/tests/rpm/api_v2/test_schedule_sync.py ...........
[...]

PR #180, opened 2016-03-30
Edit tests for the HTTP Location header

Ichimonji10:
According to RFC 7231, the HTTP Location header may be either an
absolute or relative URL. Thus, given this request:

    GET /index.html HTTP/1.1
    Host: www.example.com

These two responses are equivalent:

    HTTP/1.1 302 FOUND
    Location: http://www.example.com/index.php

    HTTP/1.1 302 FOUND
    Location: /index.php

Update tests to abide by the RFC and allow Pulp to return either
absolute or relative URLs. This change does not affect test results on
the author's test systems, and should fix spurious test failures on some
systems. Test results generated with:

    python -m unittest2 \
        pulp_smash.tests.rpm.api_v2.test_iso_crud.CreateTestCase \
        pulp_smash.tests.platform.api_v2.test_repository.CreateSuccessTestCase

Related to https://github.com/PulpQE/pulp-smash/issues/167

### DIFF
--- a/pulp_smash/tests/platform/api_v2/test_repository.py
+++ b/pulp_smash/tests/platform/api_v2/test_repository.py
@@ -19,7 +19,7 @@ from __future__ import unicode_literals
 from packaging.version import Version
 
 from pulp_smash import api, utils
-from pulp_smash.compat import urljoin
+from pulp_smash.compat import urljoin, urlparse
 from pulp_smash.constants import REPOSITORY_PATH, ERROR_KEYS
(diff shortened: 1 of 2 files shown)


Ichimonji10:
@rbarlow Does this fix test failures you're experiencing?

Ichimonji10:
Regardless of whether this pull request fixes the test failures at hand, it produces more standards-compliant tests, and it does not introduce any new failures. Merging.

PR #181, opened 2016-03-30
Add automated release and publish scripts

elyezer:
Add release.sh script that will run all the steps to release a new version and
publish make target to upload the created package to PyPi.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+set -euo pipefail
+
+# Make sure local fork is updated
+git fetch -p --all
+git checkout master
+git merge --ff-only upstream/master
+
+NEW_VERSION="$(date +%Y.%m.%d)"
+OLD_VERSION="$(git tag --list | tail -n 1)"
+
+if [ "${NEW_VERSION}" = "${OLD_VERSION}" ]; then
+    echo "Nothing to release"
+    exit 0
+fi
+
+# Bump version number
+echo "${NEW_VERSION}" > VERSION
+
+# Generate the package
+make package-clean package
+
+# Sanity check Pulp Smash packages on both Python 2 and Python 3
+for python in python{2,3}; do
+    venv="$(mktemp --directory)"
+    virtualenv -p "${python}" "${venv}"
+    set +u
+    source "${venv}/bin/activate"
+    set -u
+    for dist in dist/*; do
+        pip install --quiet "${dist}"
+        python -m pulp_smash 1>/dev/null
+        pip uninstall --quiet --yes pulp_smash
+    done
+    deactivate
+    rm -rf "${venv}"
+done
+
+# Get the changes from last release and commit
+git add VERSION
+git commit -m "Release version ${NEW_VERSION}" \
+    -m "Shortlog of commits since last release:" \
+    -m "$(git shortlog ${OLD_VERSION}.. | sed 's/^./    &/')"
+
+# Tag with the new version
+git tag "${NEW_VERSION}"
+
+fmt <<EOF
+
+This script has made only local changes: it has updated the VERSION file,
+generated a new commit, tagged the new commit, and performed a few checks along
+the way. If you are confident in these changes, you can publish them with
+commands like the following:
+EOF
+
+cat <<EOF
+
+    git push --tags origin master && git push --tags upstream master
+    make publish
+
+EOF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ help:
 	@echo "  test-coverage  to run unit tests and measure test coverage"
 	@echo "  package        to generate installable Python packages"
 	@echo "  package-clean  to remove generated Python packages"
+	@echo "  publish        to upload dist/* to PyPi"
 
 # Edit with caution! Travis CI uses this target. ¶ We run docs-clean before
 # docs-html to ensure a complete build. (Warnings are emitted only when a file
@@ -62,5 +63,8 @@ package:
 package-clean:
 	rm -rf build dist pulp_smash.egg-info
 
+publish: package
+	twine upload dist/*
+
 .PHONY: help all docs-html docs-clean lint-flake8 lint-pylint lint test \
-    test-coverage package package-clean
+    test-coverage package package-clean publish
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,3 +13,6 @@ sphinx
 
 # For `make package`
 wheel
+
+# For `make publish`
+twine


release.sh
```diff
@@ -0,0 +1,48 @@
...
+
+if [ "${VERSION}" = "${LASTEST_VERSION}" ]; then
+    echo "Nothing to release"
+    exit 0
+fi
+
+# Make sure local fork is updated
+git fetch -p --all
+git checkout master
+git merge upstream/master
```
Ichimonji10: How about `git merge --ff-only upstream/master`? This'll help prevent the creation of a merge commit in case something is weird about the branch structure.

release.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Bump version number
+echo ${VERSION} > VERSION
```
Ichimonji10: Double quotes, please?

release.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Generate the package
+make package-clean package
+
+# Sanity check Pulp Smash packages on both Python 2 and Python 3
+for python in python{2,3}; do
+    virtualenv -p ${python} pulpsmash-${python}
+    source pulpsmash-${python}/bin/activate
+    for dist in dist/*.{whl,tar.gz}; do
```
Ichimonji10: How about changing this to `for dist in dist/*; do`? It'll be a little shorter, and more importantly, it'll match what the `publish` make target does.

release.sh
```diff
@@ -0,0 +1,49 @@
...
+Shortlog of commits since last release:
+
+EOF
+git shortlog "${LASTEST_VERSION}".. | sed "s/^/    /" >> commit_message
+git commit -a -F commit_message
+rm commit_message
+
+# Tag and push to upstream
+git tag ${VERSION}
+git push --tags upstream master
```
Ichimonji10: Can we omit this last step? I do like the idea of being able to look thing over *one* last time before publishing changes. We can put something like this at the end:

```sh
fmt <<EOF
This script has made only local changes: it has updated the VERSION file, generated a new commit, tagged the new commit, and performed a few checks along the way. If you are confident with these changes, you can publish them with commands like the following:
EOF
cat <<EOF

    make publish
    git push --tags origin master && git push --tags upstream master

EOF

File: release.sh [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+done
+
+# Get the changes from last release and commit
+cat > commit_message <<EOF
+Realease version ${VERSION}
+
+Shortlog of commits since last release:
+
+EOF
+git shortlog "${LASTEST_VERSION}".. | sed "s/^/    /" >> commit_message
```
Ichimonji10: This sed expression will introduce whitespace lines. How about this?

```sh
git shortlog "${LASTEST_VERSION}.." | sed 's/^./    &/ >> commit_message
```

release.sh
```diff
@@ -0,0 +1,49 @@
...
+git commit -a -F commit_message
+rm commit_message
+
+# Tag and push to upstream
+git tag ${VERSION}
```
Ichimonji10: Double quotes, please?

release.sh
```diff
@@ -0,0 +1,63 @@
...
+        python -m pulp_smash
+        pip uninstall -y pulp_smash
+    done
+    deactivate
+    rm -rf pulpsmash-"${python}"
+done
+
+# Get the changes from last release and commit
+cat > commit_message <<EOF
+Realease version ${VERSION}
```
Ichimonji10: "Realease" should be spelled "Release", without the first "a"

release.sh
```diff
@@ -0,0 +1,63 @@
...
+
+Shortlog of commits since last release:
+
+EOF
+git shortlog "${LASTEST_VERSION}".. | sed "s/^./    &/" >> commit_message
+git commit -a -F commit_message
+rm commit_message
```
Ichimonji10: How about this?

```sh
git commit -m 'Release version 1' \
    -m 'Shortlog of commits since last release:' \
    -m "$(git shortlog | sed 's/^./    &/')"
```

elyezer: Not sure if it will match the new lines, let me try.

release.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+set -euo pipefail
+
+VERSION="$(date +%Y.%m.%d)"
+LASTEST_VERSION="$(git tag --list | tail -n 1)"
```
Ichimonji10: Would it be clearer to name these `OLD_VER` and `NEW_VER`?

elyezer: Decided to be more verbose and changed to OLD_VERSION and NEW_VERSION

release.sh
elyezer: It works as expected.

release.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Bump version number
+echo "${NEW_VERSION}" > VERSION
+
+# Generate the package
+make package-clean package
+
+# Sanity check Pulp Smash packages on both Python 2 and Python 3
+for python in python{2,3}; do
+    virtualenv -p ${python} pulpsmash-${python}
```
Ichimonji10: Ok, this really is my last suggestion! How about using `mktmp --directory` when choosing where to make a virtualenv? There shouldn't be an issue with making directories named `pulpsmash-python2` or `pulpsmas-python3`, so this isn't the most important thing in the world. All the same... it does seem proper.

```sh
for python in python{2,3}; do
    env="$(mktemp -d)"
    virtualenv --python "$python" "$env"
```

release.sh
Ichimonji10: :+1:

release.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -euo pipefail
+
+NEW_VERSION="$(date +%Y.%m.%d)"
+OLD_VERSION="$(git tag --list | tail -n 1)"
```
Ichimonji10: TL;DR All is well, and I learned something.

Interesting aside: this command returns the same value even if you have an old commit checked out. I was wondering if the ordering of this command and the `git fetch -p --all; ...` block below matters, and it seems the ordering doesn't matter.

elyezer: We can run fetch -p --all before just to make sure we get updates from upstream and then make sure that other folk release tag will be fetched before listing the tags.

Ichimonji10:
I tried running this script, and had to make some minor modifications for it to succeed. Here's a diff:

```diff
diff --git a/Makefile b/Makefile
index e20d51c..b21d4be 100644
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ CPU_COUNT=$(shell python -c "from multiprocessing import cpu_count; print(cpu_co
 help:
 	@echo "Please use \`make <target>' where <target> is one of:"
 	@echo "  help           to show this message"
+	@echo "  publish        to upload dist/* to PyPi"
 	@echo "  all            to to execute all following targets (except \`test')"
 	@echo "  docs-html      to generate HTML documentation"
 	@echo "  docs-clean     to remove documentation"
@@ -62,7 +63,7 @@ package:
 package-clean:
 	rm -rf build dist pulp_smash.egg-info
 
-publish:
+publish: package
 	twine upload dist/*
 
 .PHONY: help all docs-html docs-clean lint-flake8 lint-pylint lint test \
diff --git a/release.sh b/release.sh
index 2195082..f12441f 100755
--- a/release.sh
+++ b/release.sh
@@ -24,17 +24,20 @@ make package-clean package
 for python in python{2,3}; do
     venv="$(mktemp --directory)"
     virtualenv -p "${python}" "${venv}"
+    set +u
     source "${venv}/bin/activate"
+    set -u
     for dist in dist/*; do
-        pip install "${dist}"
-        python -m pulp_smash
-        pip uninstall -y pulp_smash
+        pip install --quiet "${dist}"
+        python -m pulp_smash 1>/dev/null
+        pip uninstall --quiet --yes pulp_smash
     done
     deactivate
     rm -rf "${venv}"
 done
 
 # Get the changes from last release and commit
+git add VERSION
 git commit -m "Release version ${NEW_VERSION}" \
     -m "Shortlog of commits since last release:" \
     -m "$(git shortlog ${OLD_VERSION}.. | sed 's/^./    &/')"
```

The new `--quiet` and `1>/dev/null` bits are there just to keep the script output to a reasonably small level. The locations of the virtualenvs are still printed, so if the script aborts, the user can still find and destroy what's been created in `/tmp/`.

release.sh
Ichimonji10: Agreed. It's possible that the fetch-and-merge could introduce new tags onto the master branch, so we should do the fetch-and-merge before setting `OLD_VERSION`.

elyezer:
I had added the quiet option before submitting but decided to use more verbose mode to make sure it is working, but relying in the return code is better.

elyezer:
Incorporated the diff from above comment, the PR should be good now.

elyezer:
Also run the scrip locally and it worked perfectly :smile:

Ichimonji10:
Looks good to me. I can execute `make`, `make publish`, and `./release.sh` all with expected results.

Ichimonji10:
ACK (finally! phew!)

elyezer:
@Ichimonji10 I'm gonna push to master then

Ichimonji10:
dpvm1

Ichimonji10:
:tada:

Issue #89, opened 2016-02-04
Cannot execute certain shell commands via cli.Client


Ichimonji10:
Python's `subprocess` module offers a very handy function under Python 3, `run`. It's used much like `cli.Client.run`. One of the few reasons I'm not currently using it is that it's unavailable on Python 2. Interest in Python 2 is also why I backported(ish) `pulp_smash.cli.CompletedProcess`. Here's a usage example:

``` python
>>> subprocess.run('echo foo'.split())
foo
CompletedProcess(args=['echo', 'foo'], returncode=0)
```

By default, `subprocess.run` does not allow you to use typical shell features like redirection:

``` python
>>> subprocess.run('echo foo > bar'.split())
foo > bar
CompletedProcess(args=['echo', 'foo', '>', 'bar'], returncode=0)
```

However, there's an escape hatch: you can ask for your command to be executed via a system shell. By default, it's `/bin/sh`. Here's an example:

``` python
>>> subprocess.run('echo foo > bar', shell=True)  # creates `bar`!
CompletedProcess(args='echo foo > bar', returncode=0)
```

Now, `cli.Client.run` lets you run commands in a manner similar to `subprocess.run`. It acts just like `subprocess.run` in most cases, even down to preventing redirection by default:

``` python
>>> client.run('echo foo > bar'.split())
CompletedProcess(args=['echo', 'foo', '>', 'bar'], returncode=0, stdout='foo > bar\n', stderr='')
```

However, the escape hatch is broken. We're trapped, and we cannot do things like pipe values. Here's an example where I try to do so, but instead get a `CommandNotFound` exception complaining about the missing command "e" (presumably because `''.split()` is called on the string):

``` python
>>> client.run('echo foo > bar')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/cli.py", line 199, in run
    code, stdout, stderr = self.machine[args[0]].run(args[1:], **kwargs)
  File "/home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.5/site-packages/plumbum/machines/local.py", line 208, in __getitem__
    return LocalCommand(self.which(cmd))
  File "/home/ichimonji10/.virtualenvs/pulp-smash3/lib/python3.5/site-packages/plumbum/machines/local.py", line 178, in which
    raise CommandNotFound(progname, list(cls.env.path))
[...]

PR #186, opened 2016-04-07
Fix a partially broken test

Ichimonji10:
From a comment in the source code:

> When executing commands over SSH, in a non-login shell, and as a non-root
> user, the PATH environment variable is quite short. For example:
>
>     /usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin

In cases where these conditions hold true, such as when Jenkins logs
into a Nodepool test system, several private functions that work with
`/usr/sbin/getenforce` will malfunction. Make these functions more
resilient by hard-coding in references to `/usr/sbin/getenforce`. This
approach is inflexible and inelegant. However:

> We're doing this it's simple (see _has_getenforce()), because Pulp is
> available on a limited number of platforms, and because we may move to
> an SSH client that allows for shell expansion.

Are there other solutions available? Sure! However:

> We cannot execute `PATH=${PATH}:/usr/sbin which getenforce` because
> Plumbum does a good job of preventing shell expansions. See:
> https://github.com/PulpQE/pulp-smash/issues/89

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_export.py
+++ b/pulp_smash/tests/rpm/api_v2/test_export.py
@@ -29,21 +29,30 @@ from pulp_smash.tests.rpm.api_v2.utils import (
 )
 
 
-def _has_selinux(server_config):
-    """Tell whether SELinux appears to be installed on the target system.
-
-    We use a fairly dumb heuristic to determine whether SELinux is installed:
-    we look for the ``getenforce`` CLI utility.
-    """
+def _has_getenforce(server_config):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I've run the relevant tests with this command:

    python -m unittest2 pulp_smash.tests.rpm.api_v2.test_export

They pass. I've run them both against my usual test systems, and also on a nodepool system that was configured to SSH into itself.

elyezer:
LGTM, pending travis.

PR #191, opened 2016-04-08
Refactor RPM unassociation tests

Ichimonji10:
Do at least the following:

* Rename class `RemoveAssociatedUnits` to `RemoveUnitsTestCase`.
* Update the assertions on `RemoveUnitsTestCase`. Drop the assertion
  that checks that a unit randomly selected from a list is present in
  the list. Of course it is - that's guaranteed by Python itself.
  Refactor the remaining assertions to, hopefully, provide more concise
  and useful messages upon test failure.
* Heavily refactor the private methods in the module, and make use of
  them as appropriate.

This commit does not introduce any failures. Test results generated
with:

    python -m unittest2 pulp_smash.tests.rpm.api_v2.test_unassociate

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_unassociate.py
+++ b/pulp_smash/tests/rpm/api_v2/test_unassociate.py
@@ -24,53 +24,61 @@ from pulp_smash.tests.rpm.api_v2.utils import (
     sync_repo,
 )
 
+_RPM_ID_FIELD = 'checksum'
+# RPM units do not have an ``id`` metadata field. This is problematic when
+# trying to select a specific RPM content unit, as the remaining metadata
+# fields may be non-unique. For example, two different RPM content units may
+# both have a name of "walrus" — they just provide different versions. This
+# constant attempts to name a substitute for an ID.
 
-def _remove_unit(client, repo_href, type_id, unit_id):
-    """Remove a unit from a repository.
 
-    :param pulp_smash.api.Client client: object to make calls to API
-    :param repo_href: url of the repo the unit is associated to
-    :param type_id: type of unit that will be removed
-    :param unit_id: id of the unit to be removed
-    :return: response from server
+def _get_unit_id(unit):
+    """Return the unit's _RPM_ID_FIELD if an RPM. Otherwise, return ID."""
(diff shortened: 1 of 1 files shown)


Ichimonji10:
For the record, this refactor was basically a chance for me to experiment with Pulp's [criteria](http://pulp.readthedocs.org/en/latest/user-guide/general-reference.html#criteria).

Ichimonji10:
As it happens, this change will also fix a bug in the test suite. In the case where two or more RPMs have the same name (such as with the "walrus" RPM, for which there are two versions), `pulp_smash.tests.rpm.api_v2.test_unassociate.RemoveAssociatedUnits.test_one_unit_removed` currently fails.

pulp_smash/tests/rpm/api_v2/test_unassociate.py
```diff
@@ -92,50 +100,28 @@ def setUpClass(cls):
...
+            unit = random.choice(_get_units_by_type(cls.units_before, type_id))
+            cls.units_removed.append(unit)
+            _remove_unit(cls.cfg, repo['_href'], unit)
+        cls.units_after = _search_units(cls.cfg, repo['_href'], cls.TYPE_IDS)
+
+    def test_units_not_in_search(self):
+        """Assert the removed units do not appear in search results."""
+        ids_removed = {_get_unit_id(unit) for unit in self.units_removed}
+        ids_after = {_get_unit_id(unit) for unit in self.units_after}
+        self.assertEqual(ids_removed & ids_after, set())  # s1.isdisjoint(s2)
```
elyezer: Smart move, using assertTrue and isdisjoint will give the same result, but using assertEqual with the two sets will actually prints what are the differences. :100:

elyezer:
All this code makes sense and are good to me.

Feel free to merge it.

pulp_smash/tests/rpm/api_v2/test_unassociate.py
Ichimonji10: :smile:

Ichimonji10:
Thanks for looking this over, @elyezer. This still looks good to me. Merging.

PR #192, opened 2016-04-11
Skip tests for which a plugin is not installed

Ichimonji10:
This pull request contains two commits. The first commit rewrites existing logic for skipping tests based on whether or not a plugin is installed. It doesn't expand that logic to any new modules. It just takes what's already in place and shuffles around the pieces. This makes it easy to verify: one need only run tests for two test modules and verify that nothing has broken. The second commit expands the per-plugin test skipping logic to the entire test suite.

### DIFF
--- a/pulp_smash/tests/docker/api_v2/test_crud.py
+++ b/pulp_smash/tests/docker/api_v2/test_crud.py
@@ -15,6 +15,11 @@ from pulp_smash.compat import urljoin
 from pulp_smash.constants import REPOSITORY_PATH
 
 
+def setUpModule():  # pylint:disable=invalid-name
+    """Skip tests if the Docker plugin is not installed."""
+    utils.skip_if_type_is_unsupported('docker_image')
+
+
 def _gen_docker_repo_body():
     """Generate a Docker repo create body.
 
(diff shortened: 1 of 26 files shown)


pulp_smash/tests/docker/api_v2/test_crud.py
```diff
@@ -15,6 +15,11 @@
 from pulp_smash.constants import REPOSITORY_PATH
 
 
+def setUpModule():  # pylint:disable=invalid-name
```
elyezer: I'm wondering if there is a way to avoid repeat this code over the other modules. Not sure if an import from other module will help.

elyezer:
It looks good to me. I'm just thinking if we can avoid that code repetition, but we can improve that later when we have a better way to handle it if we find a better way at all.

pulp_smash/tests/docker/api_v2/test_crud.py
Ichimonji10: It might be possible to define this once in another module and `import set_up_module as setUpModule`.

Ichimonji10:
The reptition in general sucks. It'd be better to use something like the load_tests protocol, but that's off the table due to the need to also support jUnit XML reports.

Ichimonji10:
Thanks for looking this over. I just realized that I failed to write up any unit tests - I'll bang some out in the next hour or so. Should be very easy.

Ichimonji10:
I've updated this PR with some unit tests. I've made no other changes. I made sure to re-run several of the affected test modules to ensure nothing breaks.

elyezer:
:+1:

Issue #182, opened 2016-03-30
The docker tests are not skipped when the docker plugin is not installed


rbarlow:
I noticed that the OSTree tests all skip if the OSTree plugin is not installed. It would be nice if the Docker tests also did this. Thanks!

Ichimonji10:
We skip tests very inconsistently. All the modules should be reviewed before closing this bug.

PR #194, opened 2016-04-12
Deduplicate test skipping logic

Ichimonji10:
A given Pulp server may have an arbitrary set of plugins installed.
Consequently, not all tests in Pulp Smash apply to a given Pulp server.
Pulp Smash chooses which tests to run by executing a `setUpModule`
function in each `test*` module, and this function talks to the Pulp
server and checks to see if the relevant plugin is installed.

This works, but it's not DRY. Every single `test*` module must define a
nearly identical function, and each of these functions contains a magic
keyword like "docker_image" or "rpm". Address this by defining one
`set_up_module` function in each of the following:

* pulp_smash.tests.docker.utils
* pulp_smash.tests.ostree.utils
* pulp_smash.tests.puppet.utils
* pulp_smash.tests.rpm.utils

Then, `import set_up_module as setUpModule` as needed. This change does
not change test results. Test results gathered with the following
commands:

    nose pulp_smash.tests
    nose2 pulp_smash.tests
    py.test --pyargs pulp_smash.tests

(Note that nose treats `SkipTest` exceptions as errors.)

Related to https://github.com/PulpQE/pulp-smash/issues/182.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -27,6 +27,7 @@ developers, not a gospel.
     api/pulp_smash.tests.docker.cli.test_crud
     api/pulp_smash.tests.docker.cli.test_sync_publish
     api/pulp_smash.tests.docker.cli.utils
+    api/pulp_smash.tests.docker.utils
     api/pulp_smash.tests.ostree
     api/pulp_smash.tests.ostree.api_v2
     api/pulp_smash.tests.ostree.api_v2.test_crud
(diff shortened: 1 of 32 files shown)


Ichimonji10:
@elyezer, this should partially address the duplicate method calls you mentioned in #192.

elyezer:
@Ichimonji10 that was fast :smile:

Thank you

Ichimonji10:
Thanks for the review. :smile:

PR #195, opened 2016-04-12
Collect all tests for RPM repo download policies

Ichimonji10:
An RPM repository may have one of several download policies for its
importer, including "background" and "on demand". These tests are split
between modules `test_download` and `test_sync_publish`, with names like
`SyncDownloadTestCase`. Testing of download policies is an important
subject, and in the author's opinion, the current situation makes it
hard to find these tests.

Collect these tests into a single new module, `test_download_policies`,
with new names like `BackgroundTestCase` (for the "background" download
policy). Doing this should improve navigability of the test suite, and
it definitely does allow for an improved test case implementation.

The tests in module `test_download_policies` succeed. Tests executed
with:

    python -m unittest2 \
        pulp_smash.tests.rpm.api_v2.test_download_policies


Ichimonji10:
I just re-ran the entire test suite with:

The test suite completed with no errors.

elyezer:
test_organization++

ACK

Ichimonji10:
Thanks for the prompt review, @elyezer. :smile:

elyezer:
Not a problem, I like good stuff.

Issue #18, opened 2015-10-19
Autogenerate API documentation stubs


Ichimonji10:
The `docs/api` directory contains files such as `pulp_smash.config.rst` and `pulp_smash.tests.platform.api_v2.test_login.rst`. There's one `.rst` file for each documented Python module. This is a burden to maintain and error-prone, especially when you consider that each page of the API documentation has an identical layout. Let's use something like Sphinx's `sphinx-autogen` script to generate these files automatically. See: [`sphinx.ext.autosummary` – Generate autodoc summaries](sphinx-doc.org/ext/autosummary.html).

elyezer:
:+1:

PR #197, opened 2016-04-13
Auto-generate API documentation stub files

Ichimonji10:
Create script `gen_api_docs.sh`. It uses the "find" utility to search
through the `pulp_smash` and `tests` directories and, for each file
found, generates a reStructuredText document in `docs/api/`. The
generated documents are semantically identical to what is present in
`docs/api/` already.

Also:

* Update the `docs-html` make target, and make it call this new script
  before Sphinx itself.
* Make git keep the `docs/api/` directory.
* Make git ignore the reST files the script places in `docs/api/`.
* Add module `pulp_smash.__main__` to the API documentation.

How does this solution compare the current state of affairs? One
downside is that developers still must maintain the index, in
`docs/api.rst`, of all modules to be included in the API documentation.
However, there are several benefits:

* reST files no longer need to be maintained in `docs/api/*`.
* Sphinx issues a warning if the index (in `docs/api.rst`) is out of
  sync with the set of reST stubs and, therefore, the set of modules in
  the application. This has already come in to play: module
  `pulp_smash/__main__` was missing from the API documentation, and it
  has been added in response to this warning.
* It is now much easier to update the API documentation pages, as there
  is now only one template, instead of dozens.

Two alternative solutions were considered. One option is to use
`sphinx-apidoc` to generate reST files, instead of this custom script.
Unfortunately, the reST files emitted by this script are unsatisfactory,
and it is impossible to make this script use a custom template. A second
option is to use the `autosummary` Sphinx directive in `docs/api.rst`
like so:

    .. autosummary::
        :toctree: api

        pulp_smash
        pulp_smash.__main__
        pulp_smash.api
        pulp_smash.cli
        …

One can then ask the `sphinx-autogen` script to generate one reST file
for each module listed. While this works, there are more downsides:

* The `sphinx-autogen` script names imports (!) in the reST files it
  generates. This is nonsensical and generates a tremendous number of
  warnings.
* The `autosummary` Sphinx directive emits tables, instead of lists.
  While arguably useful, tabulated output is much more verbose, which
  reduces legibility and makes it harder to scan through the list of
  modules. In addition, this cannot be overridden.
* Because the `autosummary` directive drives which reST files are
  generated, it is easy for the documentation to become out of sync with
  the actual set of modules which should be documented.

Fix https://github.com/PulpQE/pulp-smash/issues/18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 
 # Generated documentation
 /docs/_build/
+/docs/api/*.rst
 
 # Created by `make package`. `pulp_smash.egg_info` is also used when another
 # package does a `pip install -e …/pulp_smash`.
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ help:
 all: test-coverage lint docs-clean docs-html package-clean package
 
 docs-html:
+	./gen_api_docs.sh
 	@cd docs; $(MAKE) html
 
 docs-clean:
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -11,6 +11,7 @@ developers, not a gospel.
 .. toctree::
 
     api/pulp_smash
+    api/pulp_smash.__main__
     api/pulp_smash.api
     api/pulp_smash.cli
     api/pulp_smash.compat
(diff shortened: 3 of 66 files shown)


Ichimonji10:
Creating a `scripts` directory now seems like a much better idea.

elyezer:
:+1: for adding the scripts dir.

Ichimonji10:
See #201.

Issue #81, opened 2016-02-03
Add new test for "Uploading the same Content Unit twice"


omaciel:
As per this [issue](https://pulp.plan.io/issues/1406), there is a new regression in Pulp 2.8 Beta when one tries to upload the same content (ie. rpms, puppet modules, etc) to an existing repository. We need a new automated test that does the following:

* Create a new feed-less repository
* Manually import a valid content type (i.e. RPM, etc)
  * Assert that the content type was successfully uploaded into the repository
* Manually try to import the same exact content type to the same repository
  * Assert that a PulpCodedException is raised

We should make sure that all valid content types (rpms, puppet modules, docker images, etc) are used for this test to make sure that we get the same, expected behavior across the board.

* [ ] docker v1
* [x] puppet
* [x] python
* [x] rpm

<ins>Not testable: docker v2 and ostree. See: <a href="https://github.com/PulpQE/pulp-smash/issues/81#issuecomment-214412605">here</a>.</ins>

omaciel:
@peterlacko please consider this issue your top priority :)

peterlacko:
Sure, will do!

Ichimonji10:
I'm self-assigning this.

* #90 needs more work vis–à–vis #103, and #61 is still open.
* I want to close the "high priority" issues.
* I have an idea for how to use mixins to refactor redundant code out of the various packages, and taking on this issue will allow me to explore that idea.

Ichimonji10:
Branch https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:duplicate contains a first draft of tests for this issue. They all succeed at the moment, and they're structured so that common code can be re-used between the different plugins.

@bmbouter, this set of tests targets [Pulp #1406](https://pulp.plan.io/issues/1406). In that issue, @mhrivnak suggests that a `PulpCodedException` should be raised. Is this exception visible to an end user? Or is that an internal implementation detail?

As things stand, the test case asserts that it is possible to upload the same content to a single repository twice in a row, and that no errors are reported. From my understanding of the issue, this is how Pulp 2.7 and 2.8 should both act. Am I missing something here?

bmbouter:
@Ichimonji10 Any `PulpCodedException` is designed to be user-facing. When we raise them, the Pulp httpd middleware should convert them to JSON and hand them back to the user in a structured way. If we're generically raising a 500 without any coded exception, I consider that a bug.

[...]

PR #203, opened 2016-04-14
Add test for duplicate uploads of a Puppet module

Ichimonji10:
Add a new test case, `DuplicateUploadsTestCase`, in module
`pulp_smash.tests.puppet.api_v2.test_duplicate_uploads`. This module
uploads and imports a Puppet module into a single repository twice in a
row, and verifies that the procedure completes without any issues. The
new test case passes. Tests executed with:

    python -m unittest2 \
        pulp_smash.tests.puppet.api_v2.test_duplicate_uploads

Related to #81.

### DIFF
--- a/pulp_smash/tests/puppet/api_v2/test_duplicate_uploads.py
+++ b/pulp_smash/tests/puppet/api_v2/test_duplicate_uploads.py
@@ -1,0 +1,78 @@
+# coding=utf-8
+"""Tests for how well Pulp can deal with duplicate uploads.
+
+This module targets `Pulp #1406`_ and `Pulp Smash #81`_. The test procedure is
+as follows:
+
+1. Create a new feed-less repository.
+2. Upload content and import it into the repository. Assert the upload and
+   import was successful.
+3. Upload identical content and import it into the repository.
+
+The second upload should silently fail for all Pulp releases in the 2.x series.
+
+.. _Pulp #1406: https://pulp.plan.io/issues/1406
+.. _Pulp Smash #81: https://github.com/PulpQE/pulp-smash/issues/81
+"""
+from __future__ import unicode_literals
+
+from pulp_smash import api, utils
+from pulp_smash.compat import urljoin
+from pulp_smash.constants import (
+    CONTENT_UPLOAD_PATH,
+    PUPPET_MODULE_URL,
+    REPOSITORY_PATH,
+)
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This new module is highly similar to `pulp_smash.tsets.rpm.api_v2.test_duplicate_uploads`. I'm leaving the task of inventing a common parent class parent class, mixin, or something else to a future PR.

pulp_smash/tests/puppet/api_v2/test_duplicate_uploads.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def _upload_import_puppet_module(server_config, puppet_module, repo_href):
+    """Upload a puppet module and import it into a repository.
+
+    Return the JSON-decoded body of the call report received when importing the
+    Puppet module.
+    """
+    client = api.Client(server_config, api.json_handler)
+    malloc = client.post(CONTENT_UPLOAD_PATH)
```
elyezer: malloc?

elyezer:
LGTM

My question about malloc is that it does not makes 100% sense to me even though I got the idea of that variable being a placeholder for capturing some specific information about the Puppet Module Upload.

pulp_smash/tests/puppet/api_v2/test_duplicate_uploads.py
Ichimonji10: This is a reference to the C `malloc()` function. When called, `malloc()` allocates the requested amount of memory and returns a pointer to the beginning of that section of memory. The user must later `free()` that memory. A failure to do so results in a memory leak, and a duplicate attempt to free memory results in a segfault or some other critical error.

Pulp's handling of content uploads is reminiscent of `malloc()`. When one makes this HTTP call, the server returns an href (kinda like the C pointer) that must later be used to end the upload request. A failure to do so is a resource leak, and a duplicate attempt results (should result?) in an error.

Ichimonji10: > C dynamic memory allocation refers to performing manual memory management [via] malloc, realloc, calloc and free.

This is basically my sense of programming humour. :wink:

elyezer: Got it. :smile:

PR #205, opened 2016-04-15
Ensure docs/Makefile builds API docs

### LINKED ISSUE #204: Fix RTD API docs
The API documentation is not listed on Read The Docs right now. This bug was likely introduced by #197. A possible solution is to move `gen_api_docs.sh` to `docs/gen_api_docs.py` (note the file extension change) and ensure the script is called by the appropriate targets in `docs/Makefile`.

This obsoletes #201.

Ichimonji10:
Create a "phony" target, `gen-api-docs`, in `docs/Makefile`. This new
target generates reStructuredText API documents in `docs/api/`. Ensure
this target is called by other make targets as appropriate.

Fix #204.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ help:
 all: test-coverage lint docs-clean docs-html package-clean package
 
 docs-html:
-	./gen_api_docs.sh
 	@cd docs; $(MAKE) html
 
 docs-clean:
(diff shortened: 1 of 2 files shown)


Ichimonji10:
You can prove this fix works by performing these same two steps both without and with this commit applied:

```sh
git clean -dfx
make docs-html
git clean -dfx
cd docs && make html
```

Ichimonji10:
According to Travis, this commit correctly generates all API docs. Merging, as all checks pass, getting this in un-blocks some other work, and getting this in also re-enables API documentation on RTD.

elyezer:
Got late for the party but this look awesome!

PR #208, opened 2016-04-15
Add and use function pulp_smash.utils.sync_repo

Ichimonji10:
Add function `pulp_smash.utils.sync_repo`. Use it throughout the test
suite. Test results do not change as a result of this commit. Test
results generated with:


elyezer:
+62 -129 == :tada: 

LGTM

PR #210, opened 2016-04-18
Add test for duplicate uploads of a Python package

Ichimonji10:
Add a new test case, `DuplicateUploadsTestCase`, in module
`pulp_smash.tests.python.api_v2.test_duplicate_uploads`. This module
uploads and import a Python package into a single repository twice in a
row, and verifies that the procedure completes without any issues. The
new test case passes. Tests executed with:

    python -m unittest2 \
        pulp_smash.tests.python.api_v2.test_duplicate_uploads

Related to: https://github.com/PulpQE/pulp-smash/issues/81

This commit is highly similar to
4b11dbb73f820aa76b6756f550a2b84a5a7844f6. It would be beneficial to
refactor out the commonalities.)

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -126,6 +126,18 @@ PUPPET_MODULE_URL = ('{}/v3/files/{}-{}-{}.tar.gz'.format(
 ))
 """The URL to a Puppet module available at :data:`PUPPET_FEED`."""
 
+PYTHON_EGG_URL = (
+    'https://pypi.python.org/packages/source/p/pulp-smash/'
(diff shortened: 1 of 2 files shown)


Ichimonji10:
The "duplicate upload" and CRUD test cases are similar for all plugins. I think the tests in `pulp_smash.tests.python.api_v2` are the best implementation thus far in the test suite, and they would be a good basis for creating a common test case.

elyezer:
Very clean and readable code.

Ichimonji10:
Good to hear. :smile:

PR #211, opened 2016-04-19
Make API docs show on Read The Docs

Ichimonji10:
How is the API documentation built? The ``docs/api.rst`` file contains a
static index of documents in the ``docs/api/`` directory, and the
contents of the directory are automatically generated whenever a
relevant make target from ``docs/Makefile`` (such as "html") is
executed.

This works well so long as documentation is generated via one of the
make files. However, Read The Docs (RTD) does not use the make files.
Instead, it generates documentation by cd-ing into the ``docs``
directory and executing a command like this: [1]

    sphinx-build -b html . _build/html

What we really want is a system where:

* All users have access to the API documentation, including RTD.
* The API documentation stubs in ``docs/api/`` are auto-generated.
* An error or warning is shown if either ``docs/api.rst`` or the
  contents of ``docs/api/`` are incorrect.

Update the documentation build system to achieve these goals. Place the
documentation stubs in ``docs/api/`` under version control so RTD and
others can generate API documentation without without the make files.
And make the documentation-related make targets clean and regenerate the
contents of the ``docs/api/`` directory, thus accomplishing the other
two goals.

Fix https://github.com/PulpQE/pulp-smash/issues/204.

[1] https://read-the-docs.readthedocs.org/en/latest/builds.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ __pycache__/
 
 # Generated documentation
 /docs/_build/
-/docs/api/*.rst
 
 # Created by `make package`. `pulp_smash.egg_info` is also used when another
 # package does a `pip install -e …/pulp_smash`.
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -52,7 +52,7 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/* api/*
 
-gen-api-docs:
+gen-api-docs: clean
 	cd ..; ./scripts/gen_api_docs.sh
 
 html: gen-api-docs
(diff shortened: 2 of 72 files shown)


elyezer:
So for now on we are running the gen_docs locally whenever a new module is added/updated. This will automatically generate the stubs which will make git show the diffs, this way we are just a commit and push far from updating the docs on the Pulp Smash repository. Is that right?

If my assumption is ok, then feel free to merge this PR since it LGTM.

Ichimonji10:
The short answer is: yes.

From now on, we should make sure to run `make docs-html` and edit `docs/api.rst` each time a module is added or removed, and version control the changes. That's a pain in the butt. How is this better than two weeks ago?

* We no longer need to manually add, remove or update the documentation stub files. Instead, `scripts/gen_api_docs.sh` takes care of that for us.
* If we forget to update `docs/api.rst`, `make docs-html` will emit warnings.

PR #212, opened 2016-04-19
Unify and expand CRUD testing

Ichimonji10:
Create a new class, `pulp_smash.utils.BaseAPICrudTestCase`, use it for all CRUD tests, and add Puppet CRUD tests. This new base test case ships with seven test methods, so the number of tests is greatly expanded. And if we want to add any new test coverage or edit the existing test, it'll be much easier to do so now.

This PR is split in to three commits for the purpose of logically separating several different tasks.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -44,6 +44,7 @@ developers, not a gospel.
     api/pulp_smash.tests.platform.api_v2.test_user
     api/pulp_smash.tests.puppet
     api/pulp_smash.tests.puppet.api_v2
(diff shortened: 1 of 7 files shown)


elyezer:
This is what I call DRY taken seriously :smile: 

LGTM

Ichimonji10:
I just rebased and ran the checks again with:

    python -m unittest2 pulp_smash.tests.{python,docker,ostree,puppet}.api_v2.test_crud

All is well.

Issue #214, opened 2016-04-20
reboot_suggested is False by default if during errata upload there was not specified any value.


ipanova:
1. Upload errata into a repo and don't specify any value for 'reboot_suggested'
2. Publish repo
3. Check the updateinfo.xml --> reboot_suggested tag should not be None but False.

see: https://pulp.plan.io/issues/1782

PR #215, opened 2016-04-20
Test whether updateinfo.xml suggests a reboot

Ichimonji10:
If no value is specified for an errata's `reboot_suggested` field, the
errata is uploaded into a repository, and the repository is published,
the repository's `updateinfo.xml` file should not suggest a reboot. The
file's `<reboot_suggested>` tag should have a value of "False".

Add a test for: https://pulp.plan.io/issues/1782

Fix: https://github.com/PulpQE/pulp-smash/issues/214


Ichimonji10:
I dropped the test skipping logic from Pulp Smash and ran this test against a Pulp 2.8.1 system. The test failed, as the XML tag had a value of `"None"`. This is as expected.

Issue #161, opened 2016-03-08
Test pulp can sync drpm, srpm


preethit:
- Test Repository Sync success with 
    - [ ] drpm
    - [ ] srpm

Pulp-admin rpm repo create --repo-id drpm --feed https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/test_drpm_repo/
pulp-admin rpm repo sync run --repo-id drpm

Pulp-admin rpm repo create --repo-id srpm --feed  https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/test_srpm_repo/

pulp-admin rpm repo sync run --repo-id srpm

elyezer:
:tada:

PR #218, opened 2016-04-22
Sync DRPM and SRPM repositories

Ichimonji10:
Pulp's RPM plugin has the ability to sync not only RPM repositories, but
DRPM and SRPM repositories too. Add test cases which do this.
Implementation notes:

* Add two new constants, `DRPM_FEED_URL` and `SRPM_FEED_URL`.
* Create a new base test case, `SyncRepoBaseTestCase`, and three child
  classes, `SyncRpmRepoTestCase`, `SyncDrpmRepoTestCase` and
  `SyncSrpmRepoTestCase`.
* Class `SyncValidFeedTestCase` has effectively been renamed to
  `SyncRpmRepoTestCase`.

All affected tests pass. Tests executed with:

    python -m unittest2 pulp_smash.tests.rpm.api_v2.test_sync_publish

Fix https://github.com/PulpQE/pulp-smash/issues/161

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -40,10 +40,12 @@ from pulp_smash.compat import urljoin
 from pulp_smash.constants import (
     CALL_REPORT_KEYS,
     CONTENT_UPLOAD_PATH,
+    DRPM_FEED_URL,
     REPOSITORY_PATH,
     RPM,
     RPM_FEED_URL,
     RPM_URL,
+    SRPM_FEED_URL,
 )
 from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
 from pulp_smash.tests.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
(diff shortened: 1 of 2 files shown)


pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -101,72 +103,98 @@ def test_importer_config(self):
                 self.assertEqual(importers[0][key], body['importer_' + key])
 
 
-class SyncValidFeedTestCase(utils.BaseAPITestCase):
-    """Create an RPM repository with a valid feed and sync it.
+# This class is left public for documentation purposes.
+@unittest2.skip('Base test case.')
+class SyncRepoBaseTestCase(utils.BaseAPITestCase):
```
elyezer: Would not be better to have this base class on the utils module?

This way I think the skip will not be needed and we will not require to remember that one skip is due to this.

Ichimonji10: Good idea. I'll do that.

Ichimonji10: Ahhh, I should point out that I did place this class in module `pulp_smash.tests.rpm.api_v2.utils`, and the class was still run.

Ichimonji10: We could also place this class in `pulp_smash.utils`. I didn't do that b/c the placement seems odd. This class's `setUpClass` and `test_task_progress_report` methods are both specific to the RPM plugin.

elyezer: Either way works for me, leaving it where it is or moving to utils. Feel free to merge after deciding about the place.

Ichimonji10: I think I'll leave it as is. It's trivial to move up the directory hierarchy as needed, but at this point, the class is only used in this one module, and moving the class to a `utils` module doesn't help us.

PR #219, opened 2016-04-22
Conditionally skip Python duplicate upload tests

Ichimonji10:
Every module of tests should check to see if the necessary Pulp plugins
are installed before executing. If the plugin is missing, that module
should skip. Unfortunately, module
`pulp_smash.tests.python.api_v2.test_duplicate_uploads` is missing this
check. Add it in. This should fix a failing test in Jenkins. Test
results are sane, both when run against systems with and without the
plugin, with the following command:

    python -m unittest2 discover pulp_smash.tests.python.api_v2


elyezer:
Thank you

PR #221, opened 2016-04-25
Adds a test_repos generation script which creates the zoo repo

bmbouter:
Run by running `scripts/gen_test_repos.sh` from the root of the checkout. It will generate the test repos in a folder called `test_repos`.


scripts/gen_test_repos.sh
```diff
@@ -0,0 +1,49 @@
...
+cp scripts/test_repo_assets/rpm/*.rpm test_repos/zoo/
+createrepo -s sha256 \
+           -g ../../scripts/test_repo_assets/rpm/comps.xml \
+           test_repos/zoo
+
+# Add in an updateinfo.xml file using modifyrepo
+modifyrepo --mdtype=updateinfo \
+           scripts/test_repo_assets/rpm/updateinfo.xml \
+           test_repos/zoo/repodata/
+
```
Ichimonji10: Are lines 40 and below supposed to be in this script?

scripts/gen_test_repos.sh
```diff
@@ -0,0 +1,49 @@
...
+# using .gitignore.
+#
+# This script is primarily used by a Jenkins job which builds the assets and
+# copies the contents of /test_repos/ to [0].
+#
+# [0]:  https://repos.fedorapeople.org/repos/pulp/pulp/test_repos
+#
+# NOTE: This script should be run from the repository root directory. That is,
+# this script should be run from this script's parent directory.
+
```
Ichimonji10: A good technique for making more reliable scripts is to add:

```sh
set -euo pipefail
```

Those options will cause the script to execute under a much more strict set of conditions. See [this blog post](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

bmbouter:
Closing this because this work was taken into the [initial commit in a new repo](https://github.com/Ichimonji10/pulp-fixtures/commit/5b7cf748077610c4af38848947d6a87296da8a95). Thanks @Ichimonji10 !

scripts/gen_test_repos.sh
bmbouter: `set -euo pipefail` Looks good, thanks for the link to the blog post too.

PR #223, opened 2016-04-29
Fix test for Pulp bug 1787

Ichimonji10:
According to the bug [1], if an input package group has no
`display_order` element, then output XML should also omit said element.
The relevant test asserts that said element always appears exactly once
in the output — not good!

Thanks to @rohanpm for finding this issue.

Test results do not change as a result of this commit. (The test is
currently skipped.) Tests executed with:

    python -m unittest2 pulp_smash.tests.rpm.api_v2.test_comps_groups

[1] https://pulp.plan.io/issues/1787


elyezer:
:+1:

Issue #224, opened 2016-04-29
Add support for testing a Pulp agent


Ichimonji10:
An "agent" can be installed on a system. This agent allows Pulp to send commands to that system. See: http://pulp.readthedocs.io/en/latest/dev-guide/architecture.html We should be able to test interactions between a consumer (a system with an agent) and a Pulp server.

Let's build out support for this scenario. One option is to change how we use the `settings.json` file. (The file is badly named. It should be named `server_configs.json`, or something like that.) Currently, that file may have a single "default" section. We can change the semantics assigned to that file and expect at least a "server" section, corresponding to a Pulp server, and optionally allow for the presence of a "consumer" section, corresponding to a system with the Pulp agent installed. Modules could then choose whether or not to run based on the presence or absence of this config file section. Here's a verbose example of what could be in a test module:

```python
def setUpModule():  # pylint:disable=invalid-name
    """Skip tests if there is no system running the Pulp agent."""
    set_up_module()
    if 'agent' not in config.ServerConfig().sections():
        raise unittest2.SkipTest(
            'This module requires a system running the Pulp agent.'
        )
```

Also see: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:test-agent

elyezer:
Should we care about having more than one agent? If yes then we should consider expecting a list of agents or maybe make the settings smart and expect either a list or a dict for multiple and single scenario respectively.

Ichimonji10:
Yes, I did consider that possibility. It may be useful to be able to work with more than one agent, and that possibility is left open. That said, making `pulp_smash.settings` smart enough to deal with lists of configurations is a change in semantics that, I think, will introduce some complications.

Code that currently interacts with the `settings.json` file assumes that the file uses a simple labeling system. Want the system running Pulp? Call `ServerConfig().read('pulp')`. Want a system running the Pulp agent? Call `ServerConfig().read('pulp agent')`. What if the requested system isn't available? A `ConfigFileSectionNotFoundError` is raised. Easy enough.

Imagine we want to conditionally run some tests based on how many Pulp agents are available. If we require that a section named `pulp_agents` be present in the config file:

```python
# Require at least one agent
try:
    pulp_agents = ServerConfig().read('pulp agents')
except exceptions.ConfigFileSectionNotFoundError:
    raise unittest2.SkipTest()
if len(pulp_agents) == 0:
    raise unittest2.SkipTest()
```

[...]

PR #225, opened 2016-05-02
Change expected configuration file contents

Ichimonji10:
Fix https://github.com/PulpQE/pulp-smash/issues/224

It's technically already possible to create test cases that target both
a Pulp server and a Pulp agent. However, there is not proper
documentation for how to do this. Create such.

The most significant change in this commit regards the contents of
configuration files. Configuration files may now have two sections.
Here's what `python -m pulp_smash` has to say about the topic:

> The configuration file should have this structure:
>
>     {
>         "pulp": {
>             "base_url": "https://pulp.example.com",
>             "auth": ["username", "password"],
>             "verify": true,
>             "version": "2.7.5",
>             "cli_transport": "local"
>         },
>         "pulp agent": {
>             "base_url": "https://pulp-agent.example.com"
>         }
>     }
>
> Each section provides information about a single Pulp-related service
> and the host on which that service is installed. The "pulp" and "pulp
> agent" sections tell about the Pulp and Pulp Agent services,
> respectively. The former is required. The latter is optional, and if
> omitted, relevant tests are skipped.

The assertion that "relevant tests are skipped" will only hold true if
individual test tests (or test modules) implement such checks. A proof
of concept showing that this can be done has been written.

If a user calls `ServerConfig.read`, does not specify which
configuration file section to read, and the configuration file has a
single section named `default`, then that section is read and a
deprecation warning is raised.

### DIFF
--- a/pulp_smash/__main__.py
+++ b/pulp_smash/__main__.py
@@ -11,18 +11,30 @@ from pulp_smash.config import ServerConfig
 
 MESSAGE = tuple((
     '''\
-    Pulp Smash's command line interface has not yet been fleshed out. Please
-    create a configuration file at {} and call `python -m unittest2 discover
-    pulp_smash.tests`. The configuration file should have this structure:
+    Please create a configuration file at {} and call `python -m unittest2
+    discover pulp_smash.tests`. The configuration file should have this
+    structure:
     ''',
     '''\
-    {"default": {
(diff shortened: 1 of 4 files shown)


Ichimonji10:
See https://github.com/Ichimonji10/pulp-smash/tree/test-agent-demo for a demonstration of how to use this change.

Issue #107, opened 2016-02-12
Test dependency solving resolution on RPM repo copy


seandst:
When recursively copying rpm units from one repo to another, only the latest units should be copied. The zoo repo provides packages with dependencies that can be used to verify this functionality. Specifically, it provides two version of 'walrus', and 'chimpanzee' depends on walrus; only one version of walrus should be copied when recursively copying chimpanzee.

Here are the relevant pulp-admin commands:
```
pulp-admin rpm repo create --repo-id zoo --feed https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/
pulp-admin rpm repo sync run --repo-id zoo
pulp-admin rpm repo create --repo-id new
pulp-admin rpm repo copy rpm --from-repo-id zoo --to-repo-id new --str-eq="name=chimpanzee --recursive"
```

If this is working correctly, only walrus 5.21 is associated with the 'new' repository. If this is broken, both walrus 0.71 and walrus 5.21 will be associated with the 'new' repository.

This test could potentially be used to test dependency solving generally (or perhaps this could be a separate issue). If it's working correctly, the only latest versions of  camel, chimpanzee, fox, squirrel, and walrus will be in the 'new' repo. If the number of packages associated with the 'new' repo after copy is not exactly 5, and the names aren't the 5 aforementioned names, dependency solving is generally broken.

More information can be found in https://pulp.plan.io/issues/1432#note-9.

omaciel:
Thank you @seandst for the issue. I have set a `high priority` label to it and assigned it to @Ichimonji10 for automation.

Ichimonji10:
:+1:

PR #227, opened 2016-05-04
Test copying units between RPM repositories w/CLI

Ichimonji10:
Add module `pulp_smash.tests.rpm.cli.test_copy_units`. Add two test
cases:

1. `CopyTestCase`
2. `CopyRecursiveTestCase`

The first test case copies a single unit named "chimpanzee" from a
source repository to a target repository. The second test case
*recursively* copies a single unit named "chimpanzee" from a source
repository to a target repository.

The second test fails, and is skipped so long as Pulp issue 1895 is
open. See: https://pulp.plan.io/issues/1895

Test results are OK. Tests executed with:

    python -m unittest2 pulp_smash.tests.rpm.cli.test_copy_units

Fix: https://github.com/PulpQE/pulp-smash/issues/107


elyezer:
LGTM

PR #242, opened 2016-05-10
Skip tests affected by Pulp issue 1905

Ichimonji10:
Tests executed against both F23 and RHEL6 servers with the following
command:

    python -m unittest2 \
        pulp_smash.tests.rpm.api_v2.test_download_policies.BackgroundTestCase

When targeting F23, no test results change. When targeting RHEL 6, the
entire test case is skipped. This is because:

* Pulp issue 1905 [1] only affects RHEL 6.
* Letting the test run against a RHEL 6 server and skipping only the
  failing test methods appears to damage the Pulp services.

[1] https://pulp.plan.io/issues/1905


elyezer:
Thank you

PR #245, opened 2016-05-11
Refactor API search tests

Ichimonji10:
Each of the test cases in module
`pulp_smash.tests.platform.api_v2.test_search` needs one or more users
to be present. However, none of the test cases touch the users in any
way. Thus, it's wasteful to create new users in each test case. Create
and destroy users in `setUpModule` and `tearDownModule` functions,
respectively, and let the test cases use those users.

Tests executed with:

    git clean -dfx && time python -m unittest2 \
        pulp_smash.tests.platform.api_v2.test_search

This change does not affect test results. This change does make the
tests execute more quickly. On a F23 test system, execution time dropped
from 40s to 27s. On a RHEL6 test system, execution time dropped from 39s
to 25s.


Ichimonji10:
I want to drop the `@unittest2.skip` decorators from this module, and this was an excuse to re-familiarize myself with the tests.

elyezer:
LGTM, btw I was wondering why not use a list for the users in the way we can append new users and pop when removing. This way the global keyword can be dropped.

Ichimonji10:
Great suggestion. Commit amended.

elyezer:
I'm glad you like it.

PR #251, opened 2016-05-13
Fix test for RPM repo content unit counts

Ichimonji10:
Support for langpacks has been added in Pulp 2.9. In earlier versions of
Pulp, repositories do not track this type of content unit. Consequently,
tests that verify the number of content units in repositories must take
this in to account.

Update test `FixFileCorruptionTestCase.test_units_after_download` in
module `pulp_smash.tests.rpm.api_v2.test_download_policies`. This test
now succeeds when run against a Pulp 2.9 system, and there is no change
when the test is run against a Pulp 2.8 system. Tests executed with:

    python -m unittest2 pulp_smash.tests.rpm.api_v2.test_download_policies


bmbouter:
@Ichimonji10 This looks exactly correct.

elyezer:
:+1:

PR #252, opened 2016-05-13
Partially enable a test affected by Pulp 616

Ichimonji10:
Pulp issue 616 [1] describes how SELinux prevents Pulp from writing to
an export directory. `ExportDistributorTestCase.test_export_to_dir` is
skipped if that bug is open and SELinux is in enforcing mode on the
target system. While this behaviour prevents test failures from
occurring, it's too aggressive: it prevents us from knowing whether
exporting to a directory functions when SELinux is in permissive mode.

Update the test. If the bug is open and SELinux is in enforcing mode on
the target system, then disable SELinux for the duration of that one
test.

[1] https://pulp.plan.io/issues/616


Ichimonji10:
Manually tested in Pulp 2.8/F23, Pulp 2.9/F23, and Pulp 2.9/RHEL6. All three tests succeeded.

Ichimonji10:
For reference: https://pulp.plan.io/issues/616

Ichimonji10:
Commit message updated.

PR #253, opened 2016-05-16
- added ErratumContentRWRTest

midnightercz:
added ErratumContentRWRTest that works following way:
- upload erratum to pulp
- publish it 
- download and parse published data
- compare


Ichimonji10:
I see some binary blobs are added in this pull request. If binary blobs are needed, please submit a pull request against [Pulp Fixtures](https://github.com/PulpQE/pulp-fixtures). Those fixtures are uploaded to https://repos.fedorapeople.org/pulp/pulp/fixtures/.

Please address the static analysis issues.

A more thorough commit message is not strictly necessary but is appreciated.

See also: http://pulp-smash.readthedocs.io/en/latest/about.html#contributing

Ichimonji10:
Thanks for addressing my initial comments. I do have at least one more request: can we drop the binary blobs from this pull request? Pulp Fixtures exists in part so that we can better separate business logic from data, and as a bonus, the clone time on this repo stays remarkably low.

More generally, do we specifically need `dummy-tomcat-1.0-1.noarch.rpm`, `dummy-vim-1.0-1.noarch.rpm`, `dummy-apache-1.0-1.noarch.rpm` and `dummy-ssh-1.0-1.noarch.rpm`? Or, would it be possible to re-use the files we already have? While I don't know too much about the details of advisories, it seems we can add an advisory file to that repository, where the advisory file references files like `bear-4.1-1.noarch.rpm` and `camel-0.1-1.noarch.rpm`. I browsed through the test added by this PR, and it seems doable.

midnightercz:
I rewrote test to use zoo rpms and removed binary blobs from commit

Ichimonji10:
:+1:

Ichimonji10:
I tried running the tests added by this pull request, but unfortunately, a
variety of errors prevent the test from executing. Here's a couple of the errors
I encountered.

* The `_upload_import_rpm` function doesn't function correctly. The server
  complains about something related to headers. Rather than debugging the issue,
  I dropped the function and used `pulp_smash.utils.upload_import_unit` instead.

* The erratum has a hard-coded ID. This makes it impossible to run the test more
  than once if the test doesn't fully complete. I solved this by creating a
  `gen_erratum` function that loads the erratum from disk, generates a random
  ID, and returns that erratum.

* The call to the `strbool_convert` function always fails:

  ```python
  reboot_suggested = updateinfo.get('reboot_suggested',
                                    {'.cdata': 'False'})
  strbool = strbool_convert(reboot_suggested)  # Calls {…}.lower()
  updateinfo['reboot_suggested'] = strbool
  ```

* The `_cdata_to_text` function always fails:

  ```python
  elif isinstance(obj, list):
      for index in enumerate(obj):
          # `stack` is an immutable tuple
          stack.insert(0, (obj, index, obj[index]))
  ```

Can you update this test so that it can successfully be executed? Alternately,
I'll write the test myself if you provide me with a reference explaining the
functionality this test is exercising.

midnightercz:
_upload_import_rpm tried to fetch non-existing rpm from source, that's why yum importer failed. It's not possible to run test more then once, because all code is in setUpClass. I took inspiration in one of existing tests here, but I quickly realized it's stupid design idea. Workarounding this by generating repo/erratum everytime with different unit_key isn't solution. Potentially you would like to run tests on production environment where you don't want left any mess created by tests. strbool_convert and _cdata_to_text should be fixed now.

Ichimonji10:
*sigh*

Placing business logic in `setUpClass` has its advantages and disadvantages, and it isn't the right fit for every test.

One of the advantages is that it's a natural fit for situations where many assertions should be run against a single set-up. A simple example is the test cases [`pulp_smash.tests.platform.api_v2.test_login`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.platform.api_v2.test_login.html).

One of the disadvantages to placing business logic in `setUpClass` is that it's harder to implement reliable clean-up logic. So, if a test case doesn't run numerous assertions against a single set-up, there's little reason for it to follow this pattern.  `ErratumContentRWRTest` only has a single assertion, so it's a great candidate for having a different structure.

This question of how to both run numerous tests against a single set-up and also have more reliable tear down logic is something that's been on my mind since the inception of Pulp Smash. In fact, the current structure is a reaction to the not-so-great pattern of running the same set-up and tear-down procedures over and over, and executing slightly different assertions each time. My concern with this topic is also reflected in test cases such as [`pulp_smash.tests.rpm.api_v2.test_orphan_remove.OrphansTestCase`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_orphan_remove.html#pulp_smash.tests.rpm.api_v2.test_orphan_remove.OrphansTestCase), which experiments with pushing more logic to the individual `test*` methods. The test case also serves to highlight the very concerns that prompt the heavy usage of `setUpClass` in the first place. Can one simply put all logic in `test*` methods? Not when failure of a set-up action should force all remaining `test*` methods to fail. So how can one deal with that? It's an interesting question.

There's some options out there that may accomplish all of these goals. For example, this test method structure would both allow for a strict set-up procedure to execute, followed by independent assertions and a reliable clean-up:

```python
def test_foo(self):
    # perform set up, such as creating, syncing and publishing a repo
    client = api.Client(self.cfg)
    repo = client.post(...)
    utils.sync_repo(self.cfg, repo['_href'])
    with self.subTest(goal='verify foo has been frobnicated'):
        pass
    with self.subTest(goal='verify bar does not exist'):
        pass
```

Would I like to play around with even better test case designs, such as the above? Yes, I'd like to do that. But successful projects require compromise, and one of those compromises has been to heavily use a mediocre test case structure so that other goals can be accomplished. For example, one priority is to expand test coverage so the devs have more confidence in the state of Pulp. And that's happening. For example, just this morning, our matrix of tests executed **2,438 tests with 0 failures**. This is no mean feat when one considers that the test matrix includes **eight combinations** of operating systems and Pulp versions. (It's also a testament to the Pulp team's dedication to creating a quality product.)

One of my other goals is to assist your team in deploying new versions of Pulp. I think both of us want to see an easy deployment with as few bugs as possible. A good way to help ensure this will happen is to add test cases to Pulp Smash. So, please: let me help you. With all of that in mind:

> It's not possible to run test more then once, because all code is in setUpClass.

No, that's not true. It's impossible to re-run the test case after a setup failure because `ErratumContentRWRTest` creates resources and assigns them hard-coded IDs. *Any* unexpected error in the setup logic can trigger this situation, including when `setUpClass` encounters an error, or when the user aborts the test case.

> Workarounding this by generating repo/erratum everytime with different unit_key isn't solution.

This solution is widely used in test cases throughout Pulp Smash, and it works. It'd be even better if combined with more resilient tear down logic.

> I took inspiration in one of existing tests here, but I quickly realized it's stupid design idea.

As mentioned, there's advantages and disadvantages to placing business logic in `setUpClass`, and given that `ErratumContentRWRTest` has only one assertion, there's no reason for it to also follow that pattern.

Also: Ouch. I appreciate constructive criticism. This just burns.

> Potentially you would like to run tests on production environment where you don't want left any mess created by tests.

Please don't run Pulp Smash in a production environment. See: Pulp Smash → About → Scope and Limitations → [Destructiveness](http://pulp-smash.readthedocs.io/en/latest/about.html#destructiveness).

midnightercz:
Dude, almost TL;DR! Just to be clear here, as I mentioned earlier when we talked, I thought there was some "garbage collector" mechanism for pulp smash tests that uses all-in-setup way. However I think we can agree that fact that exceptions raised in setup methods prevents tear down method from run really sucks. I cannot imagine what was the reason to that in such impractical way. However I'm not unittest expert, so it's possible that deep in internals of unittest (and also nosetest) is good reason for that.

You can check https://github.com/PulpQE/pulp-smash/pull/267/files where I started using GC class that should do the job, however there is still few things that needs to be resolved. TestCase setUpClass is used only for setup of garbage collector and tearDown for destroying all the acquired resources. Whole setup should be in test0 which is executed as first. However as you mentioned it's not possible to depend on results of previous tests. So I thought of shared class variable that would hold information if initialization was successful or not. And in the beginning of every test depending on that initialization, there would be assertTrue for that shared flag (for example flag). I tried decorating test methods once and customizing nosetests test runner and things like that and I can tell that's really bad idea. That's why easy flag & assert mechanism. I think this design can work, of course I'm not trying to force you into anything, so it's up to you if you like it or not. 

Regarding to destructiveness of pulp smash, I was quite serious about intent running tests in production. Because there are no other tests that we could run to test that deployed pulp is operating correctly, right? And that's why I think random unit_key generator isn't solution here. I'm not saying it's not working, but I can hardly imagine that on production. So I directed all my comments to the fact that pulp_smash should be production ready one day. At least this is how I felt it. However I'm not sure about your opinion.

Ichimonji10:
> However I think we can agree that fact that exceptions raised in setup methods prevents tear down method from run really sucks. I cannot imagine what was the reason to that in such impractical way. However I'm not unittest expert, so it's possible that deep in internals of unittest (and also nosetest) is good reason for that.

I understand where you're coming from, but I disagree. I suggest reading [Perhaps unittest is wrong, and tearDown should always run](http://pythontesting.net/framework/unittest/unittest-wrong-teardown/) and the comments at the bottom of the page. The first comment by masklinn gets at the heart of the issue:

> I don’t think you’ve thought this through.
>
> Consider your own example: setup is supposed to create a directory, open a connection and set up some other stuff. setup blows up and throws, you try to run teardown anyway.
>
> And it blows up because it tries to del an attribute which was never created in the first place.
>
> If your setup blows up, there is no guarantee about its partial state, and the teardown simply can’t be run. It’s the same reason why __exit__ is *not* run if __enter__ throws.
>
> For what you want, you need to build a stack of setup/teardown operations (similar to exitStack), but setup and teardown are not up to the task

The comments include a variety of suggestions for solving the issue.

midnightercz:
Yeah, I hit that article also, but went through that very quickly and didn't read the comments. Maybe I should. But I think, in general, tearDown and tearDownClass should be called always and it should be up to programmer to write that in way where it will be able to deal with any state of resources initialized in setUp.

Ichimonji10:
> [I] didn't read the comments. Maybe I should.

Please do. They're informative. Even the original poster changed his tune in response to the comments:

> Update: Hmmm… changed my stance on this. … there are better ways to ensure teardown happens. But it was an interesting discussion…

Ichimonji10:
> Regarding to destructiveness of pulp smash, I was quite serious about intent running tests in production.

/me shrugs.

[Just know](http://pulp-smash.readthedocs.io/en/latest/about.html#destructiveness) that Pulp Smash will do the following to a system under test, and possibly more:

* It will drop databases.
* It will forcefully delete files from the filesystem.
* It will stop and start system services.
* It will attempt to trigger buggy behaviour.
* It may not clean up after itself, and this guarantee is impossible to provide.

If you and the rest of the ops team are OK with that, I am too.

midnightercz:
So how about subset of production save tests? I definitely wouldn't like to drop db in production or delete files from FS. Or do you have any other suggestion how we should tests production/preproduction?

Ichimonji10:
Hmm. If the intent is to run functional tests against production, then I'd strongly suggest selecting a subset of tests. It'd also recommend using a particular version. (See the [tags](https://github.com/PulpQE/pulp-smash/tags/).) And more broadly, I'd also focus on things like minimizing dev/prod differences and automating deployments.

All that said, I'm an amateur system administrator. A more experienced developer and/or an ops guy would be a better person to ask.

elyezer:
It is impossible for a testing software play well on both sides dev/production, the assumptions and premises are different. In this thread we can see that for development version we must trigger not working behavior and other destructive actions that will ensure a given development version is production ready.

On the other hand, for production a given installation can be health check by running a set of scripts or tests that must never do anything silly, like dropping database. Also the cleanup should be reinforced, but it should be careful tested because if some wrong id is used it will lead on data loss.

Pulp Smash initial design was to be a tool that exercises Pulp on its development environment and gives a positive or negative feedback to say if that development version is production ready. This does not mean that on production it will not raise issues. With that said, my concern here is to, if a production health check is needed a different set of tests should be created ensuring the constraints for the production check, Pulp Smash tools like its API and CLI client can be used for this.

In terms of roadmap, having a health check for production is not something listed on the top priority items because Pulp Smash still lacking of coverage for some areas in order to be able to say if a development version is production ready, maybe when we get there we can focus on providing production check.

With all that said, what we can do for this PR is to adequate it to run on development environment which is the only one environment where Pulp Smash is run right now. Production check should be thought better and careful created later, if one destructive operation goes hidden then it will be a massive loss, Pulp Smash must not be run on production environment until then.

midnightercz:
@elyezer Thanks for your comment. I think, for now, we can probably create custom nosetests configuration with production safe tests. But in the future I can imagine for example tests marked as production safe, some prerun script that would investigate all tests classes and creates production safe nose configuration that could be passed to some automation test runner (jenkins for example)

elyezer:
@midnightercz yes, this is something the tests would have when they are production ready. Thanks for the suggestion.

Ichimonji10:
I ran the tests in this PR again, and unfortunately, the tests fail. Here's what happens in a Python 2 virtualenv when testing Pulp 2.8:

```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.content.test_01_rwr             
E
ERROR: setUpClass (pulp_smash.tests.rpm.content.test_01_rwr.ErratumContentRWRTest)
  File "pulp_smash/tests/rpm/content/test_01_rwr.py", line 258, in setUpClass
    _updateinfo_transform(cls.parsed)
  File "pulp_smash/tests/rpm/content/test_01_rwr.py", line 155, in _updateinfo_transform
    for collection in collections:
TypeError: 'NoneType' object is not iterable

Ran 0 tests in 29.831s

FAILED (errors=1)
```

Here's what happens in a Python 3 virtualenv when testing Pulp master (2.9):

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ fc
python -m unittest2 pulp_smash.tests.rpm.content.test_01_rwr
E
ERROR: setUpClass (pulp_smash.tests.rpm.content.test_01_rwr.ErratumContentRWRTest)
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/rpm/content/test_01_rwr.py", line 250, in setUpClass
    'updateinfo')
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/rpm/content/test_01_rwr.py", line 81, in get_repo_md_type_url
    parsed = parse_repomd(StringIO.StringIO(metadata_f.content))
TypeError: initial_value must be str or None, not bytes

Ran 0 tests in 29.525s

FAILED (errors=1)
```

midnightercz:
Interesting, my run passed without any failures/errors. But I enhanced the code, so it shouldn't be problem anymore

Ichimonji10:
What version of Pulp and Python were used in the passing run?

Here's what I get with Python 2.7.11 against either Pulp 2.8.5 and master (2.9), with the current code:

```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.content.test_01_rwr             
F
FAIL: test_01_content_equal (pulp_smash.tests.rpm.content.test_01_rwr.ErratumContentRWRTest)
test if content is equal.
  File "pulp_smash/tests/rpm/content/test_01_rwr.py", line 274, in test_01_content_equal
    self.assertTrue(not diff, json.dumps(diff, indent=4, sort_keys=True))
AssertionError: False is not true : {
    "pkglist": [
        {
            "packages": [
                {
                    "reboot_suggested": {
                        "MISSING_IN_UPLOADED": "False"
                    }
                }, 
                {
                    "reboot_suggested": {
                        "MISSING_IN_UPLOADED": "False"
                    }
                }, 
                {
                    "reboot_suggested": {
                        "MISSING_IN_UPLOADED": "False"
                    }
                }, 
                {
                    "reboot_suggested": {
                        "MISSING_IN_UPLOADED": "False"
                    }
                }
            ]
        }
    ], 
    "pushcount": {
        "__TYPE_PUBLISHED__": "<type 'int'>", 
        "__TYPE_UPLOADED__": "<type 'unicode'>"
    }, 
    "references": [
        {
            "title": {
                "__TYPE_PUBLISHED__": "<type 'unicode'>", 
                "__TYPE_UPLOADED__": "<type 'NoneType'>"
            }
        }, 
        {
            "title": {
                "__TYPE_PUBLISHED__": "<type 'unicode'>", 
                "__TYPE_UPLOADED__": "<type 'NoneType'>"
            }
        }
    ]
}

Ran 1 test in 32.824s

FAILED (failures=1)
```

And, with Python 3.5.1 against Pulp 2.8.5 and master:

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.content.test_01_rwr             
E
ERROR: setUpClass (pulp_smash.tests.rpm.content.test_01_rwr.ErratumContentRWRTest)
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/rpm/content/test_01_rwr.py", line 252, in setUpClass
    'updateinfo')
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/rpm/content/test_01_rwr.py", line 81, in get_repo_md_type_url
    parsed = parse_repomd(StringIO.StringIO(metadata_f.content))
TypeError: initial_value must be str or None, not bytes

Ran 0 tests in 29.848s

FAILED (errors=1)
```

Both machines were provisioned this morning, using RPMs from repos.fedorapeople.org.

midnightercz:
I'm running on master branches for everything. Python 2.7 output is actually good, python 3 isn't. I will have to check that

Ichimonji10:
Subsumed by https://github.com/PulpQE/pulp-smash/pull/310.

Thanks for the efforts, @midnightercz.

PR #254, opened 2016-05-17
Fix broken sudo logic

Ichimonji10:
Function `utils.is_root` tells whether the Pulp Smash CLI client has
root access on a target system. A common use case for this function is
deciding whether to prefix a command with `sudo` or not. (`sudo` is an
optional system utility.) Fix an incorrect bit of logic that prepends
CLI commands with "sudo" when the test already has root access.


Ichimonji10:
Tested by adding a new (non-root) user to the target system, giving the user root access via sudo, configuring my local system to log in with this user, and executing the relevant test.

Also tested by configuring my local system to log in as root and executing the relevant test.

Ichimonji10:
This should fix several Jenkins tests.

elyezer:
Thank you!

Ichimonji10:
Thank *you*, @elyezer.

PR #260, opened 2016-05-19
Fix permissions in directory export tests

Ichimonji10:
Consider this series of commands:

```sh
d="$(mktemp --directory)"
setfacl -m user:apache:rwx "${d}"  # let apache user rwx in dir
sudo -u apache install -D /dev/null "${d}/foo/bar"
sha256sum "${d}/foo/bar"
```

In this scenario, `${d}/foo/` is owned by apache, which means that the
current user doesn't have permissions necessary to read `${d}/foo/bar`.

A similar issue is encountered by
`ExportDistributorTestCase.test_export_to_dir`. This test creates a
directory on the target Pulp server and asks Pulp to export a repository
to this directory. The "apache" user writes files into this directory,
meaning that the user executing this test may not have permissions
necessary for manipulating the exported files.

Fix this issue.

As an aside, fixing this issue *and* removing the logic related to Pulp
issue 1928, as added by f06d5fcde50331962a686835bc0e4b89f5db06b3, causes
`test_export_to_web` to start failing again, indicating that Pulp issue
1928 is valid.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_export.py
+++ b/pulp_smash/tests/rpm/api_v2/test_export.py
@@ -134,6 +134,8 @@ class ExportDistributorTestCase(utils.BaseAPITestCase):
         This test is skipped if selinux is installed and enabled on the target
         system an `Pulp issue 616 <https://pulp.plan.io/issues/616>`_ is open.
         """
+        sudo = '' if utils.is_root(self.cfg) else 'sudo '
+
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against:

* Pulp 2.8.3 on Fedora 23
* Pulp 2.9.0 on Fedora 22
* Pulp 2.9.0 on RHEL 7 (nodepool)

PR #261, opened 2016-05-19
Skip an abstract test case class

elyezer:
test_content_applicability makes use of an abstract test case class for the
success tests, make sure it will be skipped and not counted as a valid test.


Ichimonji10:
ACK. Travis is happy, and I'm happy. Let's ship it.

PR #262, opened 2016-05-20
Erratum content swr test

midnightercz:
To achive pulp data consistency, pulp should be able to preserve file
    data consistency after publish operation. This test syncs
    testing reposiory into pulp, publish it and then
    download and parse published data. Goal here is to have equal
    source and published data.


Ichimonji10:
Note to reviewers: this PR should be reviewed only after #253 is merged.

Ichimonji10:
I ran the tests in this PR again, and unfortunately, the tests fail. Module `test_02_swr` produces an error similar to what's mentioned in [this comment](https://github.com/PulpQE/pulp-smash/pull/253#issuecomment-223632075).

Module `test_01_swr` produces an error when the `_upload_import_rpm` function runs. The server responds with this message: "unexpected error occurred importing uploaded file: error reading package header". I remember us fixing this issue in #253 by dropping the custom `_upload_import_rpm` function and using the `pulp_smash.utils.upload_import_unit` function. Rebasing the commit in this PR so that it branches off of the work done in #253 may be a good idea. (i.e. try use git branches)

Ichimonji10:
The tests in this PR are still failing. I'm closing this so we can focus efforts elsewhere.

If you'd like to test the Pulp behaviour nominally tested by this pull request, consider either submitting a pull request with code that passes and that better integrates into the test suite, or consider opening an issue.

PR #265, opened 2016-05-23
Correct test for Pulp issue 1928

Ichimonji10:
Pulp issue 1928 only affects Pulp 2.9. Update tests accordingly, so that
a work-around is only applied a Pulp 2.9 system is under test and the
system under test is affected by the bug.


elyezer:
Finally gihub allowed me to review this.

Ichimonji10:
Tested against a pair of Pulp 2.8.3 and master (2.9.0) systems.

Issue #133, opened 2016-02-24
Test Search for content types


preethit:
Create a repository or repositories, and search for each of the following types of content:
- [ ] `distribution`
- [ ] `docker_blob`
- [ ] `docker_image`
- [ ] `docker_manifest`
- [ ] `docker_tag`
- [ ] `drpm`
- [ ] `erratum`
- [ ] `iso`
- [ ] `ostree`
- [ ] `package_category`
- [ ] `package_environment`
- [ ] `package_group`
- [ ] `package_langpacks`
- [ ] `puppet_module`
- [ ] `python_package`
- [x] `rpm`
- [x] `srpm`
- [ ] `yum_repo_metadata_file`
1. Create a repository
   
   ```
   pulp-admin rpm repo create --repo-id=ideate  \
       --feed=https://github.com/Katello/runcible/tree/master/test/fixtures/repositories/zoo5 
   ```
2. Synchronize the repository
   
   ```
   pulp-admin rpm repo sync run --repo-id=ideate
   ```
3. Display the contents (group) in the repo
   
   ```
   pulp-admin rpm repo content group --repo-id=ideate
   ```
4. Search the content using different options
   
   ```
   pulp-admin rpm repo content group --repo-id=ideate --fields=id
   pulp-admin -vv rpm repo content group --repo-id=ideate --in='id=bird,mammal'
   ```

This issue targets https://pulp.plan.io/issues/1527.

asmacdo:
Completion of this story should either include a port of the old unit search code or the creation of a separate issue:
https://github.com/RedHatQE/pulp-automation/blob/master/tests/general_tests/test_16_unit_search.py

asmacdo:
The use of the CLI here is helpful, but we will also need to use the API directly.

There was a regression recently resulting from unit searches. A test to catch this problem would be helpful.

Test Case:

1. Create and sync the zoo repository from feed
2. Check the "_id" field of a unit 
3. Use this search:
 `http --verify no --auth admin:admin POST https://localhost/pulp/api/v2/content/units/erratum/search/ criteria:='{"filters":{"_id":{"$in": ["21c1ab08-fd08-4443-91d9-f8dc6eb8ec38"]}}}'`
Note that the content type is `erratum` in this example.

The regression would have caused a 500 error, but we should also ensure that the correct unit and only that unit is returned.

Ichimonji10:
Here are the content types advertised by a current Pulp 2.8.3 system:

```
distribution
[...]

PR #268, opened 2016-05-24
Add tests that search for RPM & SRPM units

Ichimonji10:
Add module `pulp_smash.tests.rpm.api_v2.test_search`. This module has
two test cases:

* `SearchRpmTestCase`
* `SearchSrpmTestCase`

The `SearchRpmTestCase` creates and syncs an RPM repository, then
executes a variety of searches for RPM content units. The
`SearchSrpmTestCase` does the same, but for SRPM.

Add a new constant, `SRPM`. This new constant is the name of an SRPM
file, and it's used by `SearchSrpmTestCase`.

Related to https://github.com/PulpQE/pulp-smash/issues/133


Ichimonji10:
Tested against Pulp 2.8.3 and 2.9.0.

elyezer:
:+1:

Issue #233, opened 2016-05-05
Langpacks sync and Publish verification


bmbouter:
The zoo repo on fedorapeople contains a `repodata/comps.xml` file in which contains a specification for [langpacks which got added with Pulp 2.9.0](http://www.pulpproject.org/2016/05/05/langpacks-support-added-to-pulp-2-9-0/).

To accomplish this test:
1. Sync a zoo repo with the feed: `https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/rpm/`
2. Verify that the published comps.xml contains the expected langpacks definition from the fixture data. I think hardcoding it into the expected test would be simple and currently it is:

```
<langpacks>
  <match install="gorilla-%s" name="gorilla-en"/>
  <match install="zebra-%s" name="zebra-docs"/>
</langpacks>
```

The feature story:  https://pulp.plan.io/issues/1367
Pulp's pulp-admin [langpacks documentation](https://github.com/pulp/pulp_rpm/blob/c7a14c0129a241337ea182db6730955422705364/docs/user-guide/recipes.rst#package-langpacks)

Ichimonji10:
Pulp Fixtures now lists langpacks in its `comps.xml` file. See: https://github.com/PulpQE/pulp-fixtures/blob/master/rpm/assets/comps.xml#L60-L63

Two tests have langpack-specific logic in their assertions, `pulp_smash.tests.rpm.api_v2.test_orphan_remove.OrphansTestCase` and `pulp_smash.tests.rpm.api-v2.test_download_policies.FixFileCorruptionTestCase`. Here's a (trimmed) sample of relevant code:

```python
# Support for package langpacks has been added in Pulp 2.9. In earlier
# versions, langpacks are ignored.
locally_stored_units = 39  # See repo['content_unit_counts']
if self.cfg.version >= Version('2.9'):
    locally_stored_units += 1
self.assertEqual(self.repo_post_download['locally_stored_units'], locally_stored_units)
```

Do we need to do more to show that langpacks are being treated correctly after a sync?

bmbouter:
@Ichimonji10 It's so great that I can write a github issue and it gets implemented. Wow!

If I understand this, the unit count is being asserted to include the new one for langpacks. It would be good to also assert that the published comps.xml which is downloaded from the server contains the same [snippet of fixture data](https://github.com/PulpQE/pulp-fixtures/blob/master/rpm/assets/comps.xml#L60-L63). 

[...]

PR #271, opened 2016-05-25
Update content unit counts tests

Ichimonji10:
When one reads an RPM repository, the `content_unit_counts` attribute
lists how many units of each type are present in the repository.  Pulp
2.9 introduces support for the `package_langpacks` content type.  Update
a test that verifies the values in a repository's `content_unit_counts`
attribute.

The tests originally walked through each of the content types one at a
time and verified only specific content types. This works, but it means
that new content types can be added and the test will not notice. Make
the test more strict.

Related to: https://github.com/PulpQE/pulp-smash/issues/233

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -35,6 +35,7 @@ import inspect
 from itertools import product
 
 import unittest2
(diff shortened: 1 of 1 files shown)


Ichimonji10:

Ichimonji10:
Related to #233.

elyezer:
:+1:

PR #272, opened 2016-05-25
Fix test for the delete_orphans API call

Ichimonji10:
According to the Pulp API documentation for the POST
`/pulp/api/v2/content/actions/delete_orphans/` call, one should specify
the orphans to delete with a list of dicts, like so:

    [{"content_type_id": "erratum", "unit_id": "abc123"}]

The test for this path currently passes a bare dict:

    {"content_type_id": "erratum", "unit_id": "abc123"}

Fix the test.


Ichimonji10:

Thanks, @jortel.

elyezer:
:+1:

PR #273, opened 2016-05-25
Refactor an RPM publish test case

Ichimonji10:
`PublishTestCase` in module
`pulp_smash.tests.rpm.api_v2.test_sync_publish` creates a pair of
distributors and publishes to them, among other things. One of the tests
in that test case is to fetch the published RPM files and verify their
sanity.

The test should run in its entirety regardless of whether the RPM files
can be fetched. Move these actions into a test method, thus helping to
isolate this action. As a side effect, this slightly simplifies the test
case.


Ichimonji10:

Ichimonji10:
I did this while working on something else entirely, and figured it could be pulled out into its own PR.

elyezer:
:+1:

PR #274, opened 2016-05-25
Check for langpacks in published repo's comps.xml

Ichimonji10:
This pull request has two commits. The first refactors the existing tests for the `comps.xml` file. This second adds a new test case with additional tests for the `comps.xml` file.

The second commit fixes https://github.com/PulpQE/pulp-smash/issues/233 and is related to https://pulp.plan.io/issues/1367.


Ichimonji10:

I've marked this pull request as "do not merge" because, when run against a 2.9 system, the test fails. That's because there's no skipping logic in the test, and *that's* because I'm unsure of which issue to point at. Should I reference https://pulp.plan.io/issues/1367?

Ichimonji10:
Aha. This test is broken. I'm referencing a repository in which we upload a single RPM, not one in which we sync in the fixtures repository. Dumb. I'll fix this.

elyezer:
:+1:

PR #278, opened 2016-05-30
added test_force_full

midnightercz:
Simple test for force_full feature
https://pulp.plan.io/issues/1158

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_force_full.py
+++ b/pulp_smash/tests/rpm/api_v2/test_force_full.py
@@ -1,0 +1,206 @@
+# coding=utf-8
+"""Test pulp ability to run publish from scratch with 'force_full' parameter.
+
+Following steps are executed in order to test correct functionality of
+repository created with valid feed.
+
+1. Create repository1 with valid feed, run sync
+2. Create repository2 add distributor to it
+3. Copy 3 units from repository1 to repository2
+3. Publish repository2
+3. Copy 3 units from repository1 to repository2
+3. Publish repository2
+3. Publish repository2 with force_full = true
+
+4. Assert number of processed units from first publish == 3
+4. Assert number of processed units from second publish == 3
+4. Assert number of processed units from third publish == 6
+"""
+from __future__ import unicode_literals
+
+import time
+
+from pulp_smash import api, utils, selectors
+from pulp_smash.compat import urljoin
+from pulp_smash.constants import REPOSITORY_PATH, RPM_FEED_URL
+from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
+from pulp_smash.tests.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+_PUBLISH_DIR = 'pulp/repos/'
+
+
+class WaitForTasksTimeoutError(Exception):
+    """Raised when wait_for_task method waited too long."""
+
+    pass
+
+
+def wait_for_tasks(client, cfg, tasks):
+    """Wait until all tasks finish or raise exception."""
+    task_ids = set()
+    tasks_search = '/pulp/api/v2/tasks/search/'
+    search_criteria = {'filters': {'state': {'$ne': 'running'}}}
+    polls = 0
+    for task in tasks:
+        for spawned in task['spawned_tasks']:
+            task_ids.add(spawned['task_id'])
+    results = {}
+    while task_ids:
+        search_criteria['filters']['task_id'] = {'$in': list(task_ids)}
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/rpm/api_v2/test_force_full.py
```diff
@@ -0,0 +1,200 @@
+# coding=utf-8
+"""Test re-publish repository after unassociating content.
```
Ichimonji10: This module-wide docstring doesn't seem to match [Pulp #1158](https://pulp.plan.io/issues/1158) or the test case below. Am I missing something?

pulp_smash/tests/rpm/api_v2/test_force_full.py
```diff
@@ -0,0 +1,200 @@
...
+        client = api.Client(cls.cfg)
+
+        # Create repo 1
+        repo1, _ = cls.setup_repo(client)
+        cls.resources.add(repo1['_href'])  # mark for deletion
+
+        # Sync repo 1
+        cls.responses['sync'] = utils.sync_repo(cls.cfg, repo1['_href'])
+        report = cls.responses['sync'].json()
+        next(api.poll_spawned_tasks(cls.cfg, report))
```
Ichimonji10: By default, the `api.Client` class treats all HTTP 202 responses from the server as call reports, and polls all spawned tasks to completion. The `utils.sync_repo` method uses the `api.Client` class, and therefore also waits for spawned tasks to complete. These two lines of code have no effect except to poll already-completed tasks, and the same is true in several other places.

I'll take care of this in a follow-up commit if these lines are left in. No big deal. But it may set your mind at ease to know that task polling is taken care of automatically.

pulp_smash/tests/rpm/api_v2/test_force_full.py
```diff
@@ -0,0 +1,200 @@
...
+        wait_for_tasks(client, cls.cfg, [cls.responses['repo2_publish2']])
+
+        # publish with force_full
+        cls.responses['repo2_publish_full'] = client.post(
+            urljoin(repo2['_href'], 'actions/publish/'),
+            {'id': distributor2['id'], 'force_full': True},
+        ).json()
+
+    def test_initial_publish(self):
+        """Verify the HTTP status code of each publish."""
```
Ichimonji10: The docstrings on these methods don't match what the methods do. I'll re-write these docstrings to reflect the actual assertions, and if possible, I'll add an explanation of why these assertions are used. This process is error-prone, though, and the end result is more likely to be correct if the docstrings are correct.

Ichimonji10:
This test fails when run against Pulp 2.9, like so:

```
$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_force_full
E..
ERROR: test_force_full_publish (pulp_smash.tests.rpm.api_v2.test_force_full.ForceFullTestCase)
Verify the HTTP status code of each publish.
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/rpm/api_v2/test_force_full.py", line 195, in test_force_full_publish
    if step['step_type'] == 'rpms':
TypeError: string indices must be integers

Ran 3 tests in 40.261s

FAILED (errors=1)
```

Any idea what's up with that?

midnightercz:
I would say those are copy-paste error, however cp errors is more fitting term here. So I updated the doc strings, now they match the real purposes of methods/module.

pulp_smash/tests/rpm/api_v2/test_force_full.py
midnightercz: I wasn't sure about that. That's why I rather added poll. Also, once you will be able to run tests on your local deployment, you can run into mysterious assertError when, initial publish result correctly assert for 3 published rpms, but then following incremental asserts for 3 but hit 6. This situation isn't occur every run. That's also the reason why I thought poll for running tasks is needed to avoid race conditions

Ichimonji10: Oooh, that's weird. I don't know why that error would crop up.

For what it's worth, the existing task polling logic recursively waits only on the tasks mentioned by a specific task report. It doesn't fetch a global list of all asynchronous tasks on the server and wait on them.

Ichimonji10:
> copy-paste error

I figured that was the case.

> I updated the doc strings

Great! Thank you.

Ichimonji10:
Merged as d94c9cea9082823d3e18cdcc5e65b23c36a7f2f7, with a follow-up commit of 9543517c44bceac2404b1826fab1b0d6333d5518. Notice the two referenced bugs, https://pulp.plan.io/issues/1965 and https://pulp.plan.io/issues/1966. Thank you, @midnightercz.

PR #283, opened 2016-06-01
Increase hard-coded timeout by order of magnitude

Ichimonji10:
Pulp Smash has a task polling function, `poll_task`, that waits for a
task to complete and then polls each task spawned by the current task.
This function includes a hard-coded time-out. This is intended to let
Pulp Smash continue testing, even if Pulp produces a never-ending task.

The time-out is currently set to 3+ minutes. [1] This is generally
reasonable, but frequently, one or several tasks will take more than
this long. This is problematic: Pulp Smash is not in the business of
enforcing performance metrics, and it should not fail simply because of
an annoying level of slowness.

Change the time-out to 30+ minutes. This is an enormously long period of
time for any single task, and it should produce very few false-positive
test errors.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -354,7 +354,10 @@ def poll_task(server_config, href):
     :raises pulp_smash.exceptions.TaskTimedOutError: If a task takes too
         long to complete.
     """
-    poll_limit = 36  # 36 * 5s == 180s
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@seandst, you may appreciate this.

elyezer:
:+1:

Issue #269, opened 2016-05-24
Perform Forge-style puppet syncs where content units already exist in Pulp


jeremycline:
This is from https://pulp.plan.io/issues/1937

The problem boiled down to the importer recognizing the content was already in Pulp, but failing to associate it with the repository.

The test steps would be as follows:

1. upload or sync a module to repo1
2. perform a sync on repo2 against a forge-style repository (one with a modules.json file) that includes the module from 1.
3. Assert the module from 1. is in the published repo2.

To do this in pulp-admin, it could be something as simple as:

1. pulp-admin puppet repo create --repo-id=repo1 --feed=http://forge.puppetlabs.com --queries=torssh
2. pulp-admin puppet repo sync run --repo-id=repo1
3. pulp-admin puppet repo create --repo-id=repo2 --feed=http://forge.puppetlabs.com --queries=torssh
4. pulp-admin puppet repo sync run --repo-id=repo2

Ichimonji10:
Thanks for the clear explanation and the steps for reproducing. This looks easy to implement as an automated test case...

danuzclaudes:
@Ichimonji10 Thank you for the comments!

While trying to extract the content-unit-counts for each repository, I found out a very subtle detail of this issue. The very first repository, at a clean state of db, would have counts as:
```
Content Unit Counts:  
  Puppet Module: 2
\n\n\n
```

However, for the rest of repositories, they would only have counts as:
```
Content Unit Counts: 
\n\n\n
```

And **even all those repositories are all deleted, the following new repos would still have no counts**, until an execution of `pclean`.

In that case. this pull request has been revised to react to the above output.

Ichimonji10:
Fixed by #293.

PR #293, opened 2016-06-07
Fix up Pulp Issue #1937

danuzclaudes:
This commit fixes the issue: https://github.com/PulpQE/pulp-smash/issues/269.

It creates two repositories with the same feed content. By synchronizing both repos, check if the 2nd repo would have content; otherwise, the sync will fail.

Test output:
```
[vagrant@dev pulp-smash]$ python -m unittest pulp_smash.tests.puppet.cli
.test_sync_from_puppet_skip_duplicate_content.SyncFromPuppetSkipDuplicateContentTestCase
F
FAIL: test_sync_puppet_forge (pulp_smash.tests.puppet.cli
.test_sync_from_puppet_skip_duplicate_content.SyncFromPuppetSkipDuplicateContentTestCase)
Create two Puppet repositories and trigger their syncs.
  File "pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py", line 66, in test_sync_puppet_forge
    self.assertIsNot(first_repo_counts, 0, 'Counts should be non-zero.')
AssertionError: unexpectedly identical: 0 : Counts should be non-zero.

Ran 1 test in 13.514s

FAILED (failures=1)
```

The test fails due to either of the two cases:
1. At a clean state of the database, the 1st repository would have non-zero content counts, while the 2nd has no content.
2. Otherwise, neither the 1st nor the 2nd repository would have content.

[1] https://pulp.plan.io/issues/1937

### DIFF
--- a/pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
+++ b/pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
@@ -1,0 +1,102 @@
+# coding=utf-8
+"""Tests that to verify Puppet sync on duplicate contents."""
+from __future__ import unicode_literals
+
+import unittest2
+
+from pulp_smash import cli, config, utils
+from pulp_smash.constants import PUPPET_FEED, PUPPET_QUERY
+from pulp_smash.tests.puppet.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+
+class SyncFromPuppetSkipDuplicateContentTestCase(unittest2.TestCase):
+    """Test whether the importer recognizes the content already in Pulp.
+
+    According to `Pulp #1937`_, Pulp sometimes fails to sync units into a
+    repository if the unit already existed.
+
+    How do we test this? The test steps would be as follows:
+
+    1. upload or sync a module to repo1
+    2. perform a sync on repo2 against a forge-style repository
+       (one with a modules.json file) that includes the module from 1.
+    3. Assert the module from 1. is in the published repo2.
+       (i.e., check if the they have identical counts of content units)
+
+    Please see reference from: https://github.com/PulpQE/pulp-smash/issues/269
+
+    The end result is a test case that syncs an existing Puppet modules
+    into two repositories and check whether the 2nd would have same amount
+    of content units with the 1st.
+
+    .. _Pulp #1937: https://pulp.plan.io/issues/1937
+
+    """
+
+    def test_sync_puppet_forge(self):
+        """Create two Puppet repositories and trigger their syncs."""
+        cfg = config.get_config()
+        utils.pulp_admin_login(cfg)
+
+        # Create two different repositories and schedule them for syncs.
+        repo_ids = [utils.uuid4() for _ in range(2)]
+        client = cli.Client(cfg)
+
+        for repo_id in repo_ids:
+            # Create the repository.
+            cmd = (
+                'pulp-admin puppet repo create '
+                '--repo-id {} --feed {} --queries {}'
+            ).format(repo_id, PUPPET_FEED, PUPPET_QUERY)
+            client.run(cmd.split())
+
+            # Delete the repository.
+            cmd = 'pulp-admin puppet repo delete --repo-id {}'.format(repo_id)
+            self.addCleanup(client.run, cmd.split())
+
+            # Sync the repository.
+            cmd = (
+                'pulp-admin puppet repo sync run --repo-id {}'
+            ).format(repo_id)
+            client.run(cmd.split())
+
+        # Verify that all the repositories have an identical non-zero
+        # number of content units present with the 1st one
+        first_repo_counts = get_content_counts_by_repo_id(repo_ids[0], cfg)
+        self.assertIsNot(first_repo_counts, 0, 'Counts should be non-zero.')
+        for repo_id in repo_ids:
+            # Compute current repository's content unit counts
+            if repo_id is not repo_ids[0]:
+                counts = get_content_counts_by_repo_id(repo_id, cfg)
+                self.assertIsNot(counts, 0, 'Counts should be non-zero.')
+                self.assertEqual(
+                    first_repo_counts,
+                    counts,
+                    'The numbers of content units are not equal.'
+                )
+
+
+def get_content_counts_by_repo_id(repo_id, server_config):
+    """Tell how many puppet modules are in a repository.
+
+    :param repo_id: A Puppet repository ID.
+    :param  pulp_smash.config.ServerConfig server_config: Information
+        about the Pulp server being targeted.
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_sync_puppet_forge(self):
+        """Create two Puppet repositories and trigger their syncs."""
+        cfg = config.get_config()
+        utils.pulp_admin_login(cfg)
+
+        # Create two different repositories and schedule them for syncs.
+        repo_id1 = utils.uuid4()
+        repo_id2 = utils.uuid4()
+        if repo_id2 == repo_id1:
+            repo_id2 = utils.uuid4()
```
Ichimonji10: The chance of producing duplicate values is approximately one in 340,000,000,000,000,000,000,000,000,000,000,000,000. No need to check for duplicates.

What happens if this test needs a third repo? Should a variable named `repo_id3` be created? Naaah. Use a list to store this information.

pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
```diff
@@ -0,0 +1,70 @@
...
+            # Sync the repository.
+            cmd = (
+                'pulp-admin puppet repo sync run --repo-id {}'
+            ).format(repo_id)
+            client.run(cmd.split())
+
+        # Delete the two repos.
+        for repo_id in (repo_id1, repo_id2):
+            cmd = 'pulp-admin puppet repo delete --repo-id {}'.format(repo_id)
+            self.addCleanup(client.run, cmd.split())
```
Ichimonji10: What happens if an error occurs while executing `pulp-admin puppet repo sync`? In that case, these `addCleanup` functions will never execute. `addCleanup` should run immediately after each resource is created.

Ichimonji10:
Please add some assertions verifying Pulp's behaviour. Just because a sync succeeded doesn't mean that the desired content units are actually present in a repository. A good check would be to run `pulp-admin puppet repo list` (or some similar command) against each repository, and ensuring that both repositories have an identical non-zero number of content units present.

Generate documentation with `make docs-html` and run `git diff`. You should see some additional changes in `docs/` that should be added.

pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
```diff
@@ -0,0 +1,70 @@
...
+    Please see reference from: https://github.com/PulpQE/pulp-smash/issues/269
+
+    To do this in pulp-admin, it could be something as simple as:
+
+    1. pulp-admin puppet repo create --repo-id=repo1
+        --feed=http://forge.puppetlabs.com --queries=torssh
+    2. pulp-admin puppet repo sync run --repo-id=repo1
+    3. pulp-admin puppet repo create --repo-id=repo2
+        --feed=http://forge.puppetlabs.com --queries=torssh
+    4. pulp-admin puppet repo sync run --repo-id=repo2
```
Ichimonji10: Heh. I appreciate thorough documentation. Even so, this is a bit verbose. We want to tell *what*, not *how*. We want to be able to do things like change the `--queries=torssh` in our test without having to also update this docstring.

pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
danuzclaudes: @Ichimonji10: so a simple `repos = list(); repos.add(utils.uuid4()); repos.add(utils.uuid4())` would be fine?

Ichimonji10: Yes, that's fine. A more concise way to write that is:

```python
repo_ids = [utils.uuid4() for _ in range(2)]
```

danuzclaudes:
@Ichimonji10 Thank you for the comments!

While trying to extract the content-unit-counts for each repository, I found out a very subtle detail of this issue. The very first repository, at a clean state of db, would have count of 2 as:
```
Id:                   repo1
Content Unit Counts:  
  Puppet Module: 1
```

However, for the rest of repositories, they would only have counts as:
```
Id:                   repo2
Content Unit Counts:
```

And **even all those repositories have been deleted, the following new repos would still have no counts**, until an execution of `pclean`.

In that case. the code has been revised to react to the above output; it will verify that counts are non-zero and identical.

Ichimonji10:
> While trying to extract the content-unit-counts for each repository, I found out [...]

:+1: I found the same thing.

pulp_smash/tests/puppet/cli/test_sync_from_puppet_skip_duplicate_content.py
```diff
@@ -0,0 +1,102 @@
...
+        :param repo_id: the repository's id with which to get content counts
+        :param  client: the convenience object to execute commands
+        :return:    the number of content units of repository
+        """
+        cmd = (
+            'pulp-admin puppet repo list --repo-id {} '
+            '--fields content_unit_counts'
+        ).format(repo_id)
+        completed_proc = client.run(cmd.split())
+        output = getattr(completed_proc, 'stdout').split('Puppet Module:')
```
Ichimonji10: What happens if there's more than one type of content unit in this repository? For example:

```python
>>> mytext = (
...   'foo bar biz baz\n'
...   'Puppet Module: 5\n'
...   'one two three four'
... )
>>> mytext.split('Puppet Module:')
['foo bar biz baz\n', ' 5\none two three four']
```

In this case, the call to `int(output[1].strip())` will fail. A more robust solution is to find the one line of text containing the keyword 'Puppet Module:', and to parse that one line. Also, this logic seems like something that other test cases might want to make use of. A standalone function will be more re-usable.

Use a function like the following instead:

```python
def _get_n_puppet_modules(server_config, repo_id):
    """Tell how many puppet modules are in a repository.

        Pulp server being targeted.
    :param repo_id: A Puppet repository ID.
    :returns: The number of puppet modules in a repository, as an ``int``.
    """
    keyword = 'Puppet Module:'
    completed_proc = cli.Client(server_config).run((
        'pulp-admin puppet repo list --repo-id {} --fields content_unit_counts'
    ).format(repo_id).split())
    lines = [
        line for line in completed_proc.stdout.splitlines() if keyword in line
    ]
    assert len(lines) < 2
    if len(lines) == 0:
        return 0
    else:
        return int(lines[1].split(keyword)[1].strip())
```

Ichimonji10:
Merged, with a follow-up commit of 56f47781507a7a12f6a7746c57175337d8965ebc. Thank you, @danuzclaudes!

elyezer:
:tada:

Issue #255, opened 2016-05-17
Langpacks copied via pulp-admin on command line


preethit:
You can copy langpacks between repositories. You define it with pulp-admin such as:

    pulp-admin rpm repo create --repo-id repo_1
    pulp-admin rpm repo uploads langpacks -i gorilla-%s -n gorilla-en -i zebra-%s -n zebra-docs --repo-id repo_1

You can verify it's added to the repo with:

    pulp-admin rpm repo list --repo-id  repo_1

You can verify it's defined correctly with:

    pulp-admin rpm repo content langpacks --repo-id repo_1

Create your second repo

    pulp-admin rpm repo create --repo-id=repo_2

Copy the contents of first repo to the second repo

    pulp-admin rpm repo copy langpacks --from-repo-id=repo_1 --to-repo-id=repo_2

Verify  that repo_2 contains newly copied package langpacks:

     pulp-admin repo list --repo-id repo_2

The feature story: https://pulp.plan.io/issues/1367

Pulp's pulp-admin langpacks documentation: FIXME

PR #297, opened 2016-06-08
Ensure langpacks can be copied with CLI

Ichimonji10:
As of Pulp 2.9, it is possible to use the `pulp-admin rpm repo copy
langpacks` command to copy langpacks from one repository to another. Add
a test case for this feature.

Fix: https://github.com/PulpQE/pulp-smash/issues/255


Ichimonji10:
Tested against Pulp 2.8.5 and master (2.9) with `python -m unittest2 pulp_smash.tests.rpm.cli.test_copy_units`.

Issue #270, opened 2016-05-25
Test Langpacks Can be Removed


preethit:
Create repo a repo

    $pulp-admin rpm repo create --repo-id=repo_1

Upload langpacks

    $ pulp-admin rpm repo uploads langpacks -i hyphen -n hyphen-%s --repo-id repo_1

List the repo
```
$ pulp-admin repo list --repo-id repo_1

+----------------------------------------------------------------------+
                              Repositories
+----------------------------------------------------------------------+

Id:                   repo_1
Display Name:         None
Description:          None
Content Unit Counts:
  Package Langpacks: 1
```

Remove langpacks

```
$ pulp-admin rpm repo remove langpacks --repo-id repo_1 --str-eq repo_id=repo_1
$ pulp-admin repo list --repo-id repo_1

+----------------------------------------------------------------------+
                              Repositories
+----------------------------------------------------------------------+

Id:                   repo_1
Display Name:         None
Description:          None
Content Unit Counts:
  Package Langpacks: 0
```

Ichimonji10:
Fixed by #298.

PR #298, opened 2016-06-08
Fix up Pulp Smash #270

danuzclaudes:
This commit fixes the issue: https://github.com/PulpQE/pulp-smash/issues/270.

It creates a new repository, uploads langpacks into it and finally remove all
langpacks from it.

Test output:

```
[vagrant@dev pulp-smash]$ python -m unittest pulp_smash.tests.rpm.cli.test_langpacks.
UploadAndRemoveLangpacksTestCase
..
Ran 2 tests in 5.546s

OK
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -84,6 +84,7 @@ developers, not a gospel.
     api/pulp_smash.tests.rpm.api_v2.utils
     api/pulp_smash.tests.rpm.cli
     api/pulp_smash.tests.rpm.cli.test_copy_units
+    api/pulp_smash.tests.rpm.cli.test_langpacks
     api/pulp_smash.tests.rpm.cli.test_search
     api/pulp_smash.tests.rpm.utils
     api/pulp_smash.utils
--- a/pulp_smash/tests/rpm/cli/test_langpacks.py
+++ b/pulp_smash/tests/rpm/cli/test_langpacks.py
@@ -1,0 +1,93 @@
+# coding=utf-8
+"""Tests that perform langpacks upload and removal."""
+from __future__ import unicode_literals
+
+import unittest2
+
+from pulp_smash import cli, config, utils
+from pulp_smash.tests.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+
+def _count_langpacks(server_config, repo_id):
+    """Tell how many package langpacks are in a repository.
+
+    :param server_config: pulp_smash.config.ServerConfig server_config:
+        Information about the Pulp server being targeted.
+    :param repo_id: A RPM repository ID.
+    :returns: The number of package langpacks in a repository, as an ``int``.
+    """
+    keyword = 'Package Langpacks:'
+    completed_proc = cli.Client(server_config).run((
+        'pulp-admin repo list --repo-id {} '
+        '--fields content_unit_counts'
+    ).format(repo_id).split())
+    lines = [
+        line for line in completed_proc.stdout.splitlines()
+        if keyword in line
+    ]
+    assert len(lines) in (0, 1)
+    if len(lines) == 0:
+        return 0
+    else:
+        return int(lines[0].split(keyword)[1].strip())
+
+
+class UploadAndRemoveLangpacksTestCase(unittest2.TestCase):
+    """Upload langpacks and remove it through pulp-rpm.
+
+    This test targets `Pulp Smash #270`_. The test steps would be as follows:
+
+    1. upload langpacks to repo1
+    2. check whether the langpacks are really uploaded
+    3. remove all langpacks from repo1
+    4. check whether the langpacks are actually removed
+
+    .. _Pulp Smash #270: https://github.com/PulpQE/pulp-smash/issues/270
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a repository."""
+        cls.cfg = config.get_config()
+        cls.client = cli.Client(cls.cfg)
+        cls.repo_id = utils.uuid4()
+        cls.client.run(
+            'pulp-admin rpm repo create --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+    def test_01_upload_langpacks(self):
+        """Upload langpacks through pulp-rpm."""
+        # SubTest to upload langpacks into the repository
+        cmd = (
+            'pulp-admin rpm repo uploads langpacks -i hyphen '
+            '-n hyphen-%s --repo-id {}'
+        ).format(self.repo_id)
+        self.client.run(cmd.split())
+        package_counts = _count_langpacks(
+            self.cfg,
+            self.repo_id
+        )
+        self.assertGreater(package_counts, 0, 'Langpacks are not uploaded')
+
+    def test_02_remove_langpacks(self):
+        """Remove langpacks through pulp-rpm."""
+        # SubTest to remove all langpacks from the repository
+        cmd = (
+            'pulp-admin rpm repo remove langpacks --repo-id {0} '
+            '--str-eq repo_id={0}'
+        ).format(self.repo_id)
+        self.client.run(cmd.split())
+        package_counts = _count_langpacks(
+            self.cfg,
+            self.repo_id
+        )
+        self.assertEqual(package_counts, 0, 'Langpacks are not removed')
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository created by :meth:`setUpClass`."""
+        cls.client.run(
+            'pulp-admin rpm repo delete --repo-id {}'
+            .format(cls.repo_id).split()
+        )
--- a/docs/api/pulp_smash.tests.rpm.cli.test_langpacks.rst
+++ b/docs/api/pulp_smash.tests.rpm.cli.test_langpacks.rst
@@ -1,0 +1,6 @@
+`pulp_smash.tests.rpm.cli.test_langpacks`
+=========================================
+
+Location: :doc:`/index` → :doc:`/api` → :doc:`/api/pulp_smash.tests.rpm.cli.test_langpacks`
+
+.. automodule:: pulp_smash.tests.rpm.cli.test_langpacks


pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository created by :meth:`setUpClass`."""
+        cls.client.run(
+            'pulp-admin rpm repo delete --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+
+def get_langpacks_counts_by_repo_id(repo_id, server_config):
```
elyezer: Please move this before the class definition, is better to have a code used defined before because when reading the module you see this first and then goes to the class. This is not something that will break the code but will make it easier to understand.

docs/api.rst
```diff
@@ -83,6 +83,7 @@ developers, not a gospel.
     api/pulp_smash.tests.rpm.cli
     api/pulp_smash.tests.rpm.cli.test_copy_units
     api/pulp_smash.tests.rpm.cli.test_search
+    api/pulp_smash.tests.rpm.cli.test_upload_and_remove_langpacks
```
elyezer: I am wondering if we should call the  module `test_langpacks` and the inside the module have the `UploadAndRemoveTestCase`. I am thinking about this because we may want to do more operation with langpacks and having all that inside the same module sounds better to me.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+    This test case targets `Pulp Smash #270`_. How do we test this?
+    The test steps would be as follows:
+
+    1. upload langpacks to repo1
+    2. check whether the langpacks are really uploaded
+    3. remove all langpacks from repo1
+    4. check whether the langpacks are actually removed
+
+    .. _Pulp Smash #270: https://github.com/PulpQE/pulp-smash/issues/270
+
```
elyezer: This blank line is not needed.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+import unittest2
+
+from pulp_smash import cli, config, utils
+from pulp_smash.tests.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+
+class UploadAndRemoveLangpacksTestCase(unittest2.TestCase):
+    """Upload langpacks and remove it through pulp-rpm.
+
+    This test case targets `Pulp Smash #270`_. How do we test this?
```
elyezer: What about: `This test case targets `Pulp Smash #270`_. The test steps would be as follows:` ?

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+        ).format(self.repo_id)
+        self.client.run(cmd.split())
+        package_counts = get_langpacks_counts_by_repo_id(
+            self.repo_id,
+            self.cfg
+        )
+        self.assertEqual(package_counts, 0, 'Langpacks are not removed')
+
+    @classmethod
+    def tearDownClass(cls):
```
elyezer: There is already some machinery to remove created repos (at least for API), what about using it if that is the case?

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+        cmd = (
+            'pulp-admin rpm repo uploads langpacks -i hyphen '
+            '-n hyphen-%s --repo-id {}'
+        ).format(self.repo_id)
+        self.client.run(cmd.split())
+        package_counts = get_langpacks_counts_by_repo_id(
+            self.repo_id,
+            self.cfg
+        )
+        self.assertGreater(package_counts, 0, 'Langpacks are not uploaded')
```
elyezer: We try to have one assert per test method, let's see if @Ichimonji10 have any suggestion here.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+    :param server_config: pulp_smash.config.ServerConfig server_config:
+        Information about the Pulp server being targeted.
+    :returns: The number of package langpacks in a repository, as an ``int``.
+    """
+    keyword = 'Package Langpacks:'
+    completed_proc = cli.Client(server_config).run((
+        'pulp-admin repo list --repo-id {} '
+        '--fields content_unit_counts'
+    ).format(repo_id).split())
+    lines = [
```
elyezer: I remember seen this kind of parsing in other places, seems that we need a global parser that will get the output and then return a easy to inspect object like a dictionary. This could be in place already, but if it is not then we should start planning to have that ASAP.

docs/api.rst
Ichimonji10: :+1:

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
Ichimonji10: :+1:

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
Ichimonji10: It's nice to have assertions isolated from each other, so that when one assertion fails, the other still run. We can do this either by putting one assert per test method, or by using the `subTest` context manager.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
Ichimonji10: There are three different PRs right now that are all adding similar logic. A globally available parser would be fantastic. See #293 and #297. There are some useful comment on this topic in #297.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
Ichimonji10: Unfortunately, there is no common tear down machinery for the CLI. Thus far, I've been handling this by using base classes to minimize the amount of pain. :/

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
danuzclaudes: I could help on refactoring out this global parser...but now 100% sure how to deal with the 3 pull-requests. Is it right that, if doing so, I have to send out another PR, get it merged, and all the 3 PRs should revise to call that parser correspondingly?

Ichimonji10: IMO, we should merge these three PRs and refactor them after. It'll be more expedient.

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
Ichimonji10: :+1:

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
elyezer: seems that we have a plan, +1 on refactoring later

pulp_smash/tests/rpm/cli/test_upload_and_remove_langpacks.py
elyezer: :+1:

pulp_smash/tests/rpm/cli/test_langpacks.py
```diff
@@ -0,0 +1,92 @@
...
+        """Create a repository."""
+        cls.cfg = config.get_config()
+        cls.client = cli.Client(cls.cfg)
+        cls.repo_id = utils.uuid4()
+        cls.client.run(
+            'pulp-admin rpm repo create --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+    def test_langpacks_upload_remove(self):
```
Ichimonji10: Nice work. This PR is looking really good! I have just one more request: split this test method in two. unittest-compatible test runners execute test methods one at a time, in alphanumeric order. As a result, we can create a pair of methods named e.g. `test_01_uploads_langpacks` and `test_02_remove_langpacks`, and know that they will be executed in the right order.

One thing to keep in mind when doing this is that the test should still be valid even if one or more test methods fails. What happens if `test_01_...` fails? Will `test_02_...` still work correctly?

Ichimonji10: As an example of this, see http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_orphan_remove.html#pulp_smash.tests.rpm.api_v2.test_orphan_remove.OrphansTestCase

danuzclaudes: @Ichimonji10, from the `OrphansTestCase`, it looks like each `test_0x_...` would have `SubTest` wherever the test methods might fail, right?

Ichimonji10: A `test_*` method can fail, and all the other test methods will still run. A `subTest` context manager is only needed in a `test_*` method when multiple independent assertions are present.

Ichimonji10: https://docs.python.org/3.5/library/unittest.html#distinguishing-test-iterations-using-subtests

Ichimonji10: In this PR, only two assertions are present. If `test_langpacks_upload_remove` is split into two methods, then one assertion will end up in each of those two methods. As a result, `subTest` shouldn't be needed.

danuzclaudes: OK..i see it. If they are still in one test method then SubTest is required...but now since they will be organized into two separate methods, just assertion would be fine. And unittest can continue all test methods even one might fail unexpectedly.

danuzclaudes: @Ichimonji10 But what if test_01_upload failed? And since no langpacks were uploaded to the repository, then test_02 would always fail, right?

Ichimonji10: > since they will be organized into two separate methods, just single assertion would be fine. And unittest can continue all test methods even one might fail unexpectedly.

Right.

> since no langpacks were uploaded to the repository, then test_02 would always fail, right?

Great question. :smile: What *does* happen when langpacks are removed from a repo with no langpacks?

Ichimonji10: Answer:

```
[root@mgmt6 ~]# pulp-admin rpm repo remove langpacks --repo-id ichi10 --str-eq repo_id=ichi10
This command may be exited via ctrl+c without affecting the request.


[\]
Running...

Nothing found that matches the given criteria.

[root@mgmt6 ~]# echo $?
0
```

danuzclaudes: Yeah...after I run `python -m unittest ....UploadAndRemoveLangpacksTestCase.test_02_remove_langpacks`, I realize that, since there will be no exceptions raised by the CLI u just mentioned, the `assert 0==0` always holds. Thanks.

danuzclaudes:
Updated the code and test outcome re-published. Thanks!

Ichimonji10:
Merged, with a follow-up commit of a1d4f80186a42f31191b7405021058bef4f8c50d. Thank you, @danuzclaudes!

elyezer:
:tada:

Issue #141, opened 2016-02-29
Test pulp-admin content source


preethit:
http://docs.pulpproject.org/user-guide/content-sources.html

$ pulp-admin content sources
Usage: pulp-admin [SUB_SECTION, ..] COMMAND
Description: manage content sources

Available Commands:
  list    - list sources
  refresh - refresh sources

https://pulp.plan.io/issues/1692

elyezer:
PR #305 is merged. This can be closed.

PR #305, opened 2016-06-13
Fix up Pulp Smash #141

danuzclaudes:
This commit fixes the issue: #141 and pulp story: https://pulp.plan.io/issues/1692

For now, a content source is manually added into the directory: `/etc/pulp/content/sources/conf.d/`, and the test are written with the assumption that the content source file is present.
    
It checks if there exists a content source, refresh either all sources or a specified source, and delete a specified source. **Else, the test case of deleting a specific source will be skipped.**
    
This commit has also reated a new module: `pulp_smash.tests.platform.cli`.
Test output:
    
    ```
    [vgrant@dev pulp-smash]$ python -m unittest pulp_smash.tests.platform.cli.test_content_sources
    .s.
    ---------------------------------------------------------------------
    Ran 3 tests in 2.629s
    OK (skipped=1)
    ```

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -12,6 +12,13 @@ CALL_REPORT_KEYS = frozenset(('error', 'result', 'spawned_tasks'))
     http://docs.pulpproject.org/en/latest/dev-guide/conventions/sync-v-async.html#call-report
 """
 
+CONTENT_SOURCE_ID = 'content-world'
+"""See: `Content Sources`_.
+
+.. _Content Sources:
+    http://pulp.readthedocs.io/en/latest/user-guide/content-sources.html
+"""
+
 CONTENT_UNITS_PATH = '/pulp/api/v2/content/units/'
 """See: `Search for Units`_.
 
--- a/pulp_smash/tests/platform/cli/test_content_sources.py
+++ b/pulp_smash/tests/platform/cli/test_content_sources.py
@@ -1,0 +1,96 @@
+# coding=utf-8
+"""Tests for list/refresh/delete content sources."""
+from __future__ import unicode_literals
+
+import unittest2
+from packaging.version import Version
+
+from pulp_smash import cli, config
+from pulp_smash.constants import CONTENT_SOURCE_ID
+
+
+def _get_content_source_ids(server_config):
+    """Get the id list of all content sources, or empty list.
+
+    :param server_config: Information about the Pulp server being targeted.
+    :type server_config: pulp_smash.config.ServerConfig server_config.
+    :returns: A list of content source IDs, where each ID is a string.
+    """
+    keyword = 'Source Id:'
+    completed_proc = cli.Client(server_config).run(
+        'pulp-admin content sources list'.split()
+    )
+    lines = [
+        line for line in completed_proc.stdout.splitlines()
+        if keyword in line
+    ]
+    return [line.split(keyword)[1].strip() for line in lines]
+
+
+class RefreshAndDeleteContentSourcesTestCase(unittest2.TestCase):
+    """Test whether pulp-admin client can refresh and delete content source.
+
+    This test case targets Pulp #1692`_ and `Pulp Smash #141`_. The
+    `content sources`_ documentation describes the CLI syntax. The
+    test steps are as follows:
+
+    1. Create configuration file of a content source.
+    2. Check whether the content sources list is empty.
+    3. Refresh all content sources. Verify no errors are reported.
+    4. Refresh a specified content source. Verify no errors are reported.
+    5. Remove the specified content source. Verify that the source is
+        actually deleted.
+
+    For now, a content source is manually added into the directory:
+    `/etc/pulp/content/sources/conf.d/`, and the test are written with
+    the assumption that the content source file is present.
+
+    .. _Pulp #1692:  https://pulp.plan.io/issues/1692
+    .. _Pulp Smash #141: https://github.com/PulpQE/pulp-smash/issues/141
+    .. _content sources:
+        http://pulp.readthedocs.io/en/latest/user-guide/content-sources.html
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Verify if content resources exist."""
+        cls.cfg = config.get_config()
+        cls.client = cli.Client(cls.cfg)
+        cls.source_id = CONTENT_SOURCE_ID
+        if cls.cfg.version < Version('2.8.3'):
+            raise unittest2.SkipTest('This test requires at least 2.8.3')
+
+    def test_01_refresh_all_sources(self):
+        """Refresh all content sources."""
+        completed_proc = self.client.run(
+            'pulp-admin content sources refresh'.split()
+        )
+        self.check_error_existing(completed_proc)
+
+    def test_02_refresh_specific_source(self):
+        """Refresh a specific content sources."""
+        if len(_get_content_source_ids(self.cfg)) == 0:
+            raise unittest2.SkipTest(
+                'This test requires a content source defined in: '
+                '`/etc/pulp/content/sources/conf.d`'
+            )
+        completed_proc = self.client.run((
+            'pulp-admin content sources refresh '
+            '--source-id {}'.format(self.source_id)
+        ).split())
+        self.check_error_existing(completed_proc)
+
+    def test_03_delete_content_sources(self):
+        """Delete a specific content source."""
+        self.client.run((
+            'pulp-admin content catalog delete -s {}'.format(self.source_id)
+        ).split())
+        # Verify that the specified content source does not exist.
+        self.assertNotIn(self.source_id, _get_content_source_ids(self.cfg))
+
+    def check_error_existing(self, completed_proc):
+        """Verify no errors were reported."""
+        phrase = 'Task Failed'
+        for stream in ('stdout', 'stderr'):
+            with self.subTest(stream=stream):
+                self.assertNotIn(phrase, getattr(completed_proc, stream))
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -44,6 +44,8 @@ developers, not a gospel.
     api/pulp_smash.tests.platform.api_v2.test_repository
     api/pulp_smash.tests.platform.api_v2.test_search
     api/pulp_smash.tests.platform.api_v2.test_user
+    api/pulp_smash.tests.platform.cli
+    api/pulp_smash.tests.platform.cli.test_content_sources
     api/pulp_smash.tests.puppet
     api/pulp_smash.tests.puppet.api_v2
     api/pulp_smash.tests.puppet.api_v2.test_crud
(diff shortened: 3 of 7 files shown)


pulp_smash/constants.py
```diff
@@ -12,6 +12,13 @@
     http://pulp.readthedocs.io/en/latest/dev-guide/conventions/sync-v-async.html#call-report
 """
 
+CONTENT_SOURCE_ID = 'content-world'
```
elyezer: We should be creating random IDs using uuid calls. This id was an example on documentation.

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    keyword = 'Source Id:'
+    completed_proc = cli.Client(server_config).run(
+        'pulp-admin content sources list'.split()
+    )
+    lines = [
+        line for line in completed_proc.stdout.splitlines()
+        if keyword in line
+    ]
+    assert len(lines) >= 0
```
Ichimonji10: `lines` must be at least zero elements long. This guarantee is provided by the Python language itself. This assertion can be dropped.

pulp_smash/constants.py
elyezer: Or am I missing something?

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+    if len(lines) == 0:
+        return None
+    return [line.split(keyword)[1].strip() for line in lines]
+
+
+class RefreshAndDeleteContentSourcesTestCase(unittest2.TestCase):
```
elyezer: There should be a base class to inherit from, which provides the cfg attribute, check pulp_smash.utils for more informaiton.

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+    .. _Pulp #1692:  https://pulp.plan.io/issues/1692
+    .. _Pulp Smash #141: https://github.com/PulpQE/pulp-smash/issues/141
+    .. _content sources:
+        http://pulp.readthedocs.io/en/latest/user-guide/content-sources.html
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Verify if content resources exist."""
+        cls.cfg = config.get_config()
```
elyezer: When inheriting from the base class, call super setUpClass here.

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+    completed_proc = cli.Client(server_config).run(
+        'pulp-admin content sources list'.split()
+    )
+    lines = [
+        line for line in completed_proc.stdout.splitlines()
+        if keyword in line
+    ]
+    assert len(lines) >= 0
+    if len(lines) == 0:
+        return None
```
Ichimonji10: Imagine the Pulp server has content sources with IDs 15, 323 and 11. In this case, this function returns `[15, 323, 11]`.

Imagine the Pulp server has content sources with IDs 15 and 323. In this case, this function returns `[15, 323]`.

Imagine the Pulp server has a content source with an ID of 15. In this case, the function returns `[15]`.

Imagine the Pulp server has no content sources. Shouldn't the function return `[]`?

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+    return [line.split(keyword)[1].strip() for line in lines]
```
Ichimonji10: Are content sources integers? If so, cast them here. If not, note so in the docstring.

pulp_smash/constants.py
elyezer: I checked my local Pulp 2.9 and when running the following:

```
[vagrant@dev ~]$ ls /etc/pulp/content/sources/conf.d/
```

I get no output, which means that you probably will need to set a content source. Since I don't know much about this feature, it may require to create some fixture in order to have this information.

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+        cls.cfg = config.get_config()
+        if cls.cfg.version < Version('2.9'):
+            raise unittest2.SkipTest('This test requires Pulp 2.9 or greater.')
+        cls.client = cli.Client(cls.cfg)
+        cls.source_id = CONTENT_SOURCE_ID
+        cls.source_ids = _get_content_source_ids(cls.cfg)
+        if cls.source_ids is None:
+            raise unittest2.SkipTest(
+                'This test requires a content source defined '
+                'in `/etc/pulp/content/sources/conf.d`'
```
Ichimonji10: Can this test case add a content source?

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+from pulp_smash import cli, config
+from pulp_smash.constants import CONTENT_SOURCE_ID
+
+
+def _get_content_source_ids(server_config):
+    """Get the id list of all content sources, or None if the list is empty.
+
+    :param server_config: pulp_smash.config.ServerConfig server_config:
+        Information about the Pulp server being targeted.
+    :return: None if the list of content sources is empty, or the IDs of all
```
Ichimonji10: How about `:returns: A list of content source IDs, where each ID is a string.`

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+from packaging.version import Version
+
+from pulp_smash import cli, config
+from pulp_smash.constants import CONTENT_SOURCE_ID
+
+
+def _get_content_source_ids(server_config):
+    """Get the id list of all content sources, or None if the list is empty.
+
+    :param server_config: pulp_smash.config.ServerConfig server_config:
```
Ichimonji10: This line has an extra `:`. Please read http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain, and please verify the output of `make docs-html`.

pulp_smash/constants.py
Ichimonji10: Content sources may be named anything at all. During run-time, we could make a content source with an ID of `foo`, or of `0b935025-573d-42df-b8ff-19e0686c6d52`, or of anything else. Given that, is there a reason to define a single global content source ID?

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Verify if content resources exist."""
+        cls.cfg = config.get_config()
+        if cls.cfg.version < Version('2.9'):
+            raise unittest2.SkipTest('This test requires Pulp 2.9 or greater.')
+        cls.client = cli.Client(cls.cfg)
+        cls.source_id = CONTENT_SOURCE_ID
+        cls.source_ids = _get_content_source_ids(cls.cfg)
```
Ichimonji10: `cls.source_ids` is only referenced in this one method. Is there a reason to make it a class attribute?

pulp_smash/tests/platform/cli/test_content_sources.py
Ichimonji10: The base class is only present for API tests, IIRC. :cry: We could certainly make one - this is a common enough need.

pulp_smash/constants.py
danuzclaudes: @elyezer, actually there's no content source for a fresh Pulp, in `/etc/pulp/content/sources/conf.d/`

@Ichimonji10, as for creating content source, I understand what you mean...but currently I can't find an existing module related to content source, therefore I use a constant as workaround.

pulp_smash/tests/platform/cli/test_content_sources.py
danuzclaudes: Actually the content source was manually added into folder. As soon as there's a `content.conf` in the folder `/etc/pulp/content/sources/conf.d/`, Pulp could automatically read in that file...

Also, from the docs, they assume an existing content source in Pulp...and they provide no pulp-admin CLI tool of adding it...

Ichimonji10: For now, manually add a content source into that directory, and write the test with the assumption that the content source file is present.

Creating a content source file on the fly is troublesome due to the `cli.Client`'s inability to make use of pipes. It'd be sweet if we could fix that shortcoming. See #89.

danuzclaudes: :+1: will read thru that tutorial in detail

pulp_smash/tests/platform/cli/test_content_sources.py
```diff
@@ -0,0 +1,96 @@
...
+    .. _Pulp Smash #141: https://github.com/PulpQE/pulp-smash/issues/141
+    .. _content sources:
+        http://pulp.readthedocs.io/en/latest/user-guide/content-sources.html
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Verify if content resources exist."""
+        cls.cfg = config.get_config()
+        if cls.cfg.version < Version('2.9'):
```
elyezer: What do you have on you Pulp Smash settings.json file?

Ichimonji10:
```python
>>> Version('2.9') < Version('2.9b1')
False
```

Ichimonji10: Alpha comes before beta, comes before rc, comes before a release.

danuzclaudes: Gotcha!

danuzclaudes:
Thank you guys for your advice! I have updated this PR with most of them, though there remain several structural issues for future works, such as "Base class for CLI tests" and "Creating a content source file on the fly".

Ichimonji10:
> Base class for CLI tests

IMO, we can do that in a separate PR. There are plenty of other modules that can benefit from this.

> Creating a content source file on the fly

This one is important, and I'm not sure how to solve it.

danuzclaudes:
> There are plenty of other modules that can benefit from this.

I expect to finish this milestone soon...beyond that I am willing to help on this...as well as other possible issues.

danuzclaudes:
Hi all, one update on this PR: Skip `test_02_refresh_specific_source` if there's no content source in the Pulp instance. `test_01_refresh_all_sources` and `test_03_delete_content_sources` will raise no exception, whether or not the content source exists.

elyezer:
Will work on this PR today

elyezer:
Merged as 60d93c098bf4764df865422e168379b8402fec75 with a follow up commit of 84ce1a884ceed98549be6d8db4af33855b51c270. Tests are passing against Pulp 2.8.6 and 2.9

danuzclaudes:
@elyezer, Thank you for your review and improvement. I have always saved either your or @Ichimonji10's newer version as a **template** for my next python script. And this really forged my python coding to be more elegant and advanced.

Ichimonji10:
I'm glad to hear it, @danuzclaudes.

Issue #300, opened 2016-06-10
Test that rpms can be published to custom directory


preethit:
https://pulp.plan.io/issues/1976

Test that a separate directory can be specified where packages can be published.

Context
There was decision made in the past which led to this split (repodata and packages in separate directory) and that's why we have to keep this structure compatible. Although we have some repositories where repodata are in the same directory as packages. I don't know historical details, but as I mentioned above, fedora repositories also use this.

Acceptance Criteria
- attribute has to be optional
- attribute has to be in format that is allowed by mkdir
- if attribute is not set, packages have to be sync to relative_path of repository

Ichimonji10:
Can I get more information about this desired test case? I've read through the following:

* https://pulp.plan.io/issues/1976
* https://github.com/pulp/pulp_rpm/pull/897

However, I'm unclear on:

* Is the new `packages_directory` yum exporter configuration option only available via the API? Or is there a new CLI option too?
* What is the intended effect of this new option? A contrast between what's placed on disk without using this option and with using this option would be great.

@midnightercz, you might be able to shed some light on this.

midnightercz:
@Ichimonji10 sorry for late response. I think @bmbouter forced me to do also CLI option for that. 
When you set this option, your rpms should end in repo/<packages_directory> instead of /repo directory

Ichimonji10:
I'm adding this issue to the July 15th milestone because there's been a bit of a ruckus surrounding https://pulp.plan.io/issues/1976, and having a test for that feature request will help prevent additional issues.

PR #312, opened 2016-06-20
Added basic tests for 'packages_directory' feature.

midnightercz:
Module: pulp_smash.tests.rpm.api_v2.test_packages.directory
TestCase: PackagesDirectoryTestCase
 - tests if default location of published packages
 - tests custom location of published packages

Fix: https://github.com/PulpQE/pulp-smash/issues/300

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_packages_directory.py
+++ b/pulp_smash/tests/rpm/api_v2/test_packages_directory.py
@@ -1,0 +1,152 @@
+# coding=utf-8
+"""Test re-publish repository after unassociating content.
+
+Following steps are executed in order to test correct functionality of
+repository created with valid feed.
+
+1. Create repository foo with valid feed, run sync, add distributor to it and
+   publish over http and https.
+2. Pick a unit X and and assert it is accessible.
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This test passes, but it doesn't actually test whether the `packages_directory` feature works or not. The problem lies with this snippet of code:

```python
packages_locations = get_locations(self.responses['publish2_xml'])
for location in packages_locations:
    self.assertEqual(os.path.dirname(location), 'Packages')
```

This bit of code assumes that several `packages_locations` have been found. However, due to the way `get_locations` is implemented, `packages_locations` is an empty list.

I believe the fix is to use different namespaces in the XPath expressions. I'll try it out tomorrow, and if it works, will use the fix in a follow-up commit.

Ichimonji10:
I've taken a good, close look at the code, and there's a few things that need to be fixed up. Given how much time I've spent looking at the code, and how much time I've spent toying around with new solutions, I'll work on coming up with additional commits that satisfy my concerns.

Ichimonji10:
Superseded by #331.

Issue #311, opened 2016-06-17
Follow-up issues with Repo creation/copy/publish (1/2)


danuzclaudes:
This is the 1st follow-up issue after https://pulp.plan.io/issues/1684#change-10578 and Pulp-Smash Issue https://github.com/PulpQE/pulp-smash/issues/190.

As @mhrivnak has pointed out thru his [comments](https://pulp.plan.io/issues/1684#change-10578) as the 5th one, a new test case will be setup as following:

1. create repo1 and sync from upstream.
2. find an RPM that has two different versions in the repo.
3. create a new repo2, and copy the older of the rpm version into it.
4. setup that repo2 on a host, and install the packages.
5. copy a newer version into the repo2 and re-publish

---
Finally, yum update on the client, and see what happens. But this will be left to future work since currently pulp-smash does not have utility to setup a client vm.

Ichimonji10:
> But this will be left to future work since currently pulp-smash does not have utility to setup a client vm.

Pulp Smash will never, ever have the ability to set up a client VM.

This test can already be implemented in Pulp Smash, including installing a package on a client.

Ichimonji10:
See: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:test-agent-demo

Ichimonji10:
Can you update the name of this issue, please? When browsing through the list of issues, it'd be nice to see something like "Test whether Pulp can do foo," instead of "this issue is a follow up to another issue."

danuzclaudes:
gotcha!

Ichimonji10:
AFAIK, Pulp doesn't have any way of issuing commands to clients. My assumption is that the Pulp Smash system would execute commands on `server2` via SSH, using the `pulp_smash.cli.Client` class.

elyezer:
AFAIK, so far we have tests that involves Pulp components either on the same machine or in multiple (even though the later is not running at the moment).

We also need ability to have machines consuming Pulp content which we can call as being clients and so far there is no machinery in place for that. We can take advantage of the Pulp Smash settings and maybe expect some `client-` entries which would create a static environment or, on the other hand, we can have a entry for a client provisioning system and create VMs there which can build a dynamic environment and we can expect a clean client every time.

The first approach the responsibility to provide some clients is on the person which is running Smash and on the second approach the responsibility is to Smash create its own VMs, but the person which is running need to provide the system to get the machines provisioned.

Ichimonji10:
[...]

PR #314, opened 2016-06-22
Fix up Pulp Smash #311

danuzclaudes:
This commit fixes the issue: #311.

The test steps are as following:

1. create and sync repo1 from upstream.
2. find an RPM that has two different versions in the repo1 (walrus).
3. create a new repo2, and copy the older of the rpm version (e.g., walrus-0.71) into it.
4. Also install all dependency packages of the walrus RPM (whale, shark, stork).
5. setup that repo2 on the same host of Pulp server, and install the packages:
    `sudo yum-config-manager --add-repo {url_of_repo2_after_publish}`
    `sudo yum install --nogpgcheck walrus`.

    (The repo2 will be automatically enabled after publishing in this case.)
6. copy a newer version into the repo2 and re-publish
7. yum update on the client (`sudo yum update walrus`).

Test Output:

```
[vagrant@dev pulp-smash]$ clear; python -m unittest pulp_smash.tests.rpm.cli.test_copy_units.
CopyAndPublishTwoVersionsRepoTestCase

.s
Ran 2 tests in 71.745s

OK (skipped=1)
[vagrant@dev pulp-smash]$ 
```

### DIFF
--- a/pulp_smash/tests/rpm/cli/test_copy_units.py
+++ b/pulp_smash/tests/rpm/cli/test_copy_units.py
@@ -8,11 +8,117 @@ import subprocess
 import unittest2
 
 from pulp_smash import cli, config, constants, selectors, utils
+from pulp_smash.compat import urljoin
 from pulp_smash.tests.rpm.utils import set_up_module
 from pulp_smash.tests.rpm.cli.utils import _count_langpacks
 
 _REPO_ID = None
 """The ID of the repository created by ``setUpModule``."""
+
+
+def _get_rpm_names_versions(server_config, repo_id):
+    """Get a dict of repo's RPMs with names as keys, mapping to version lists.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server being targeted.
+    :param repo_id: A RPM repository ID.
+    :returns: The names of all modules in a repository, with a list of all
+        versions mapped to each module, as an ``dict``.
+    """
+    keyword = 'Filename:'
+    completed_proc = cli.Client(server_config).run(
+        'pulp-admin rpm repo content rpm --repo-id {}'.format(repo_id).split()
+    )
+    filenames = [
+        line for line in completed_proc.stdout.splitlines() if keyword in line
+    ]
+    assert len(filenames) > 0
+    rpm_dict = {}
+    for file_name_str in filenames:
+        # Example of a filename string: 'Filename: walrus-0.71-1.noarch.rpm'.
+        name_version = file_name_str.split('-')
+        rpm_dict.setdefault(
+            name_version[0].split(keyword)[1].strip(),
+            []
+        ).append(name_version[1])
+    return rpm_dict
+
+
+def _copy_repo(server_config, source, destination, rpm_name, rpm_version=None):
+    """Copy specific module name and version from repo1 to repo2.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server being targeted.
+    :param source: The source module to copy from.
+    :param destination: The destination module to copy to.
+    :param rpm_name: The name of target module.
+    :param rpm_version: The version of target module.
+    :returns: A :class: `pulp_smash.cli.CompletedProcess`.
+    """
+    return cli.Client(server_config).run(
+        ' '.join([
+            'pulp-admin rpm repo copy rpm --from-repo-id {}'.format(source),
+            '--to-repo-id {}'.format(destination),
+            '--str-eq=name={}'.format(rpm_name)
+            if rpm_name is not None else '',
+            '--str-eq=version={}'.format(rpm_version)
+            if rpm_version is not None else '',
+        ]).split()
+    )
+
+
+def _get_rpm_dependencies(server_config, repo_id, rpm_name):
+    """Get a list of all required packages for a given RPM in repository.
+
+    :param pulp_smash.config.ServerConfig server_config: Information about the
+        Pulp server being targeted.
+    :param repo_id: The ID of repository which contains the module.
+    :param rpm_name: The name of module to query with. It is assumed that
+        there's no cycle in the dependecy relations; that is, a RPM cannot
+        point itself as required package.
+    """
+    keyword = 'Requires:'
+    res = []
+    stack = [rpm_name]
+    while len(stack) > 0:
+        top = stack.pop()
+        if top is not rpm_name:
+            res.append(top)
+        # Query parent modules for the current RPM `top`.
+        completed_proc = cli.Client(server_config).run(
+            'pulp-admin rpm repo content rpm --repo-id {0} --str-eq name={1} '
+            '--fields=requires'.format(repo_id, top).split()
+        )
+        # Depth-First-Search for parent's dependencies.
+        # E.g., 'walrus-0.71' <- 'whale' <- 'shark, stork' <- ''.
+        # But 'walrus-5.21' <- '', i.e., it has no parent package.
+        parent_modules = [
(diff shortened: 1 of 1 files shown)


danuzclaudes:
Hi folks, I have automated test steps 1~5 so far, as mentioned in above commit announcement. I'd appreciate it if you could review on these steps first. Then, I'll add on further test case to execute yum update on the client.

danuzclaudes:
Hi all, I have managed to add yum install/update functionality to the automation script. Sorry for changing the code rapidly. I think it's all set for your review. Thanks!

Ichimonji10:
> Sorry for changing the code rapidly.

No worries. You communicated your intent by appropriately setting the PR title.

elyezer:
This is requiring a rebase and conflict solving

danuzclaudes:
> This is requiring a rebase and conflict solving

Have fixed that. Thank you!

pulp_smash/tests/rpm/cli/test_copy_units.py
```diff
@@ -181,3 +270,131 @@ def test_copy_langpacks(self):
...
+        super(CopyAndPublishTwoVersionsRepoTestCase, cls).setUpClass()
+        # Retrieve all modules with multiple versions in the repo1.
+        cls.rpm_dict = {
+            key: value for (key, value)
+            in _get_rpm_names_versions(cls.cfg, _REPO_ID).items()
+            if len(value) > 1
+        }
+        assert len(cls.rpm_dict) > 0
+        # Choose the first module with multiple versions.
+        cls.rpm_name = cls.rpm_dict.keys()[0]
```
Ichimonji10: Under Python 3, `keys()` returns a view object, which causes this line of code blows up. Please fix.

Ichimonji10:
This test fails when run against Pulp 2.8. (I'm unsure about the state of 2.9. See https://github.com/PulpQE/pulp-smash/pull/314#discussion_r69014167.) Can you update this PR so that the tests added by it pass?

pulp_smash/tests/rpm/cli/test_copy_units.py
```diff
@@ -181,3 +270,131 @@ def test_copy_langpacks(self):
...
+        self._copy_and_publish(self.rpm_name, rpm_version)
+        # Setup the repo2 on the same host.
+        cli.Client(self.cfg).run(
+            'sudo yum-config-manager --add-repo {0}{1}'
+            .format('https://dev/pulp/repos/', self.repo_id).split()
+        )
+        cli.Client(self.cfg).run('sudo yum repolist enabled'.split())
+        # Execute `yum install` to deploy the RPM on the host.
+        completed_proc = cli.Client(self.cfg).run(
+            'sudo yum install -y --nogpgcheck {}'.format(self.rpm_name).split()
```
danuzclaudes: @Ichimonji10: Sorry again for the failure. After trying to reproduce it for multiple times, I finally found out that, the `yum install` of a RPM 'walrus-0.71' may **fail sometimes, while in other times, it will succeed**. And the output is due to this:
```
CompletedProcess(args=[u'sudo', u'yum', u'update', u'walrus'], returncode=1,
stdout=u'Last metadata expiration check: 0:00:05 ago on Wed Jun 29 21:22:24 2016.
No match for argument: walrus\n',
stderr=u"Yum command has been deprecated, redirecting to '/usr/bin/dnf update walrus'.\
See 'man dnf' and 'man yum2dnf' for more information.
To transfer transaction metadata from yum to DNF, run:
'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'

Failed to synchronize cache for repo 'dev_pulp_repos_repo2_', disabling.
Package walrus not installed, cannot update it.
Error: No packages marked for upgrade.\n")
```

As u can see, is it sort of related with **"synchronize cache"** for repository that is already existing? So first, I'll add try..catch module to make my script more robust. And then I suppose to execute a `yum clean` in order to get rid of this "sometimes works; sometime fails" problem.

Ichimonji10: > Sorry again for the failure :(

It's totally OK. I'm happy that you're putting effort into learning how to write solid tests. Thank *you*.

Thanks for the info about why the test failure occurs. Knowing that this happens on your end will really help others (me? Elyezer?) come up with a good solution.

danuzclaudes: From [this article](https://access.redhat.com/solutions/203603), it mentions 3 possible causes:
  
    1. Corrupted yum cache.
    2. Network related issues.
    3. Presence of customized or other media repositories.

So I'll suppose the 2nd one is unlikely in my case. As for the 1st cause, I'll use `setUp` method to do `yum clean` before every test. And in case any exceptions thrown by `yum install`, I will catch that to `yum clean` + `yum install` to see if this will be working.

As for the 3rd cause of similar packages, I think that the yum.repos.d is assumed to be clean after a fresh publishing; so I can also ignore it for now.

danuzclaudes: I have finally managed to reproduce this in a single run of this test case in the format of bash scripts. Please note the error on this line:

> Status code: 404 for https://dev/pulp/repos/838dd32a-b2e5-4c3f-a622-47baa4dbe214/


To be frank, I have *totally no idea* how this uuid-ish repository name would occur, since this repo is published as `repo2`, as you can see from this line:

> added from: https://dev/pulp/repos/repo2/

And there's no doubt that the yum install cannot find its mirror...Even though I have executed both `yum clean all` and `pclean` before the yum install, I think there's some underlying bug behind it? Please let me if this will be a true bug, or my mistake that I have not noticed. Thanks.

pulp_smash/tests/rpm/cli/test_copy_units.py
```diff
@@ -181,3 +270,131 @@ def test_copy_langpacks(self):
...
+            with self.subTest():
+                self.assertIn(rpm_version, getattr(completed_proc, 'stdout'))
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repositories and clean up orphans."""
+        super(CopyAndPublishTwoVersionsRepoTestCase, cls).tearDownClass()
+        # Delete the published repositories.
+        cli.Client(cls.cfg).run(
+            'sudo rm /etc/yum.repos.d/dev_pulp_repos_{}.repo'
```
danuzclaudes: > 'sudo rm /etc/yum.repos.d/dev_pulp_repos_{}.repo'

Instead of this, it should be sudo rm `/etc/yum.repos.d/dev_pulp_repos_{}`**_**`.repo`. lol..Please let me execute the test case multiple times again to verify that.

danuzclaudes:
@elyezer: I have cleaned the code fyi. Currently, the test will fail "occasionally", instead of "sometimes", due to 'yum install' on missing mirror issue. Please let me know the test result on your machine. 

One problem on repeated testing is that, if the repository is not cleaned in `yum.repos.d`, another publishing and yum install will subscribe from that legacy repository, **which will easily corrupt yum**. I think this is a lesson that I've learned from last week. In that case, I have elevated [this line](https://github.com/PulpQE/pulp-smash/pull/314/files#diff-a673d553dd971cf0abea01396bd63ff1R423) before `yum remove` to avoid that.

Thank you very much for your help and for your review!

elyezer:
Merged as 307e19f57de1e41de0e8f72cbb9c3866538d670d with two follow up commits 152603d9f215f0c7401f4be4ebddee811fdc14b0 which refactor a util function to follow the public attr name convention and b18c222e81b2363f95aadcce551b80cc587b0645 which improves the tests added by this PR.

PR #324, opened 2016-06-29
Work around Pulp issue 2042

Ichimonji10:
According to Pulp issue 2042, the `updateinfo.xml` file is missing
checksums on Pulp 2.9. Work around this issue. See:
https://pulp.plan.io/issues/2042


Ichimonji10:
Tested with the following against Pulp 2.8 and 2.9:

    python -m unittest2 pulp-smash.tests.rpm.api_v2.test_upload_publish.UploadErratumTestCase

This fixes one test failure for each of the Pulp 2.9 test matrix items. That's four failures in total.

Issue #319, opened 2016-06-28
Test that pulp can upload package_environments


preethit:
https://pulp.plan.io/issues/1003
Test package environments can be uploaded

```
$ pulp-admin rpm repo uploads environment

Command: environment
Description: creates a new package environment

Available Arguments:

  --bg                 - if specified, the client process will end immediately
                         (the task will continue to run on the server)
  --repo-id            - (required) unique identifier; only alphanumeric, ., -,
                         and _ allowed
  -v                   - display extra information about the upload process
  --environment-id, -i - (required) id of the package environment
  --name, -n           - (required) name of the package environment
  --description, -d    - (required) description of the package environment
  --display-order      - display order for the package environment. Defaults to
                         1024
  --group, -g          - package group IDs to include in the package
                         environment; multiple may be indicated by specifying
                         the argument multiple times
The following options are required but were not specified:
  --repo-id
  --environment-id
  --name
  --description
```

PR #335, opened 2016-07-18
Fix Pulp Smash #319

danuzclaudes:
This PR targets on #319 and https://pulp.plan.io/issues/1003.
It uploads some environments data into a new repo and verifies that the uploading succeeds.

Test output:

```
[vagrant@dev pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.cli.test_environments
.
Ran 1 test in 21.093s

OK
```

### DIFF
--- a/pulp_smash/tests/rpm/cli/test_environments.py
+++ b/pulp_smash/tests/rpm/cli/test_environments.py
@@ -1,0 +1,76 @@
+# coding=utf-8
+"""Tests for RPM package environments."""
+from __future__ import unicode_literals
+
+import subprocess
+import unittest2
+
+from pulp_smash import cli, config, utils
+from pulp_smash.constants import RPM_FEED_URL
+
+
+class UploadPackageEnvTestCase(unittest2.TestCase):
+    """Test whether Pulp can upload package environments into a repository.
+
+    This test case covers `Pulp #1003`_ and the corresponding Pulp Smash
+    issue `Pulp Smash #319`_. The following test steps are based on official
+    `Pulp RPM Recipes`_.
+
+    1. Create and sync a repository.
+    2. Upload the environmrnt into this repo and check if there's any error.
+
+    .. _Pulp #1003: https://pulp.plan.io/issues/1003
+    .. _Pulp Smash #319: https://github.com/PulpQE/pulp-smash/issues/319
+    .. _Pulp RPM Recipes: http://docs.pulpproject.org/plugins/pulp_rpm/
+            user-guide/recipes.html#create-your-own-package-environment
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create and sync a repository."""
+        cls.cfg = config.get_config()
+        cls.repo_id = utils.uuid4()
+        cls.client = cli.Client(cls.cfg)
+        cls.client.run(
+            'pulp-admin rpm repo create --repo-id {} --feed {}'
+            .format(cls.repo_id, RPM_FEED_URL).split()
+        )
+        try:
+            cls.client.run(
+                'pulp-admin rpm repo sync run --repo-id {}'
+                .format(cls.repo_id).split()
+            )
+        except subprocess.CalledProcessError:
+            cls.client.run(
+                'pulp-admin rpm repo delete --repo-id {}'
+                .format(cls.repo_id).split()
+            )
+            raise
+
+    @classmethod
+    def tearDownClass(cls):
+        """Destroy the repository."""
+        cls.client.run(
+            'pulp-admin rpm repo delete --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+    def test_upload_environment(self):
+        """Test if package environments can be uploaded."""
+        rpm_env_name = 'Pulp Test Packages'
+        rpm_env_desc = 'A package environment of Pulp tests.'
+        proc_upload = self.client.machine.session().run(
+            'pulp-admin rpm repo uploads environment '
+            '--repo-id {0} --environment-id {1} '
+            '--name "{2}" --description "{3}"'
+            .format(self.repo_id, utils.uuid4(), rpm_env_name, rpm_env_desc)
+        )
+        with self.subTest(comment='verify upload environments stdout'):
+            self.assertNotIn('Task Failed', proc_upload[1])
+        proc_content = self.client.run(
+            'pulp-admin rpm repo content environment --repo-id {}'
+            .format(self.repo_id).split()
+        )
+        for expected in (rpm_env_name, rpm_env_desc):
+            with self.subTest():
+                self.assertIn(expected, proc_content.stdout)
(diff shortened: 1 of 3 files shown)


pulp_smash/constants.py
```diff
@@ -248,6 +248,11 @@
 )
 """The absolute path to :data:`pulp_smash.constants.RPM` in the filesystem."""
 
+RPM_ENV_DESC = 'A package environment of Pulp tests.'
+
+RPM_ENV_NAME = 'Pulp Test Packages'
+"""The name and description of Package Environments."""
```
elyezer: Split this docstring and add the description part to the RPM_ENV_DESC and leave the name information here.

elyezer: By the way there is no need to create a constant for the name and description they can go inside the test itself.

pulp_smash/tests/rpm/cli/test_environments.py
```diff
@@ -0,0 +1,70 @@
...
+    RPM_ENV_DESC,
+    RPM_ENV_NAME,
+    RPM_FEED_URL,
+)
+
+
+class UploadPackageEnvTestCase(unittest2.TestCase):
+    """Test whether Pulp can upload package environments into a repository.
+
+    This test case covers `Pulp # 1003`_ and the corresponding Pulp Smash
```
elyezer: Missing the links for the Pulp redmine issue and the pulp smash issue.

elyezer: Would be great to also link to the documentation http://docs.pulpproject.org/plugins/pulp_rpm/user-guide/recipes.html#create-your-own-package-environment

pulp_smash/tests/rpm/cli/test_environments.py
```diff
@@ -0,0 +1,70 @@
...
+            '--name "{2}" --description "{3}"'
+            .format(self.repo_id, utils.uuid4(), RPM_ENV_NAME, RPM_ENV_DESC)
+        )
+        with self.subTest(comment='verify upload environments stdout'):
+            self.assertNotIn('Task Failed', proc_upload[1])
+        proc_content = cli.Client(self.cfg).run(
+            'pulp-admin rpm repo content environment --repo-id {}'
+            .format(self.repo_id).split()
+        )
+        self.assertNotEqual('\x1b[0m', proc_content.stdout)
```
elyezer: A better assertion would be checking the values passed are the ones returned by the list command:

Below is a sample output from the docs:

```
$ pulp-admin rpm repo content environment --repo-id repo_1 --match id=test-env

Description: test-env
Group Ids:   pulp_dotted_name_packages, pulp_test_packages
Id:          test-env
Name:        test-env
Options:
```

A better assertion would be checking if the name and description match what you've passed during creation.

pulp_smash/tests/rpm/cli/test_environments.py
```diff
@@ -0,0 +1,70 @@
...
+    def tearDownClass(cls):
+        """Destroy the repository."""
+        cli.Client(cls.cfg).run(
+            'pulp-admin rpm repo delete --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+    def test_upload_environment(self):
+        """Test if package environments can be uploaded."""
+        proc_upload = cli.Client(self.cfg).machine.session().run(
```
elyezer: Can environments be deleted? If yes then a cleanUp should be added in order to remove the created environment after the test.

pulp_smash/tests/rpm/cli/test_environments.py
```diff
@@ -0,0 +1,70 @@
...
+        cls.client.run(
+            'pulp-admin rpm repo create --repo-id {} --feed {}'
+            .format(cls.repo_id, RPM_FEED_URL).split()
+        )
+        try:
+            cls.client.run(
+                'pulp-admin rpm repo sync run --repo-id {}'
+                .format(cls.repo_id).split()
+            )
+        except subprocess.CalledProcessError:
```
elyezer: A better approach would be:

```python
try:
    # ...
finally:
    cls.client.run(
        'pulp-admin rpm repo delete --repo-id {}'
        .format(cls.repo_id).split()
    )

Comment (elyezer) [inline pulp_smash/tests/rpm/cli/test_environments.py]: This way you don't need to worry about the exception and to reraise it later.

File: pulp_smash/tests/rpm/cli/test_environments.py [hunk 6]
```diff
@@ -0,0 +1,70 @@
...
+            cls.client.run(
+                'pulp-admin rpm repo delete --repo-id {}'
+                .format(cls.repo_id).split()
+            )
+            raise
+
+    @classmethod
+    def tearDownClass(cls):
+        """Destroy the repository."""
+        cli.Client(cls.cfg).run(
```
elyezer: Why not use the `cls.client` and create a new client instead?

pulp_smash/tests/rpm/cli/test_environments.py
elyezer: Why not use `self.client` and create a new client instead?

pulp_smash/tests/rpm/cli/test_environments.py
```diff
@@ -0,0 +1,70 @@
...
+        """Test if package environments can be uploaded."""
+        proc_upload = cli.Client(self.cfg).machine.session().run(
+            'pulp-admin rpm repo uploads environment '
+            '--repo-id {0} --environment-id {1} '
+            '--name "{2}" --description "{3}"'
+            .format(self.repo_id, utils.uuid4(), RPM_ENV_NAME, RPM_ENV_DESC)
+        )
+        with self.subTest(comment='verify upload environments stdout'):
+            self.assertNotIn('Task Failed', proc_upload[1])
+        proc_content = cli.Client(self.cfg).run(
```
elyezer: Why not use `self.client` and create a new client instead?

pulp_smash/constants.py
Ichimonji10: As an aside, are there uniqueness constraints on names?

pulp_smash/tests/rpm/cli/test_environments.py
elyezer: Never mind, the delete should happen later on tearDownClass and not here.

pulp_smash/tests/rpm/cli/test_environments.py
danuzclaudes: > Can environments be deleted

IMO, the package environments are associated with a repository...since the uploading requires a specific repo-id. In that case, when deleting the repository in tearDownClass step, the package envs should also be deleted? At least it can no longer be looked up through `pulp-admin rpm repo content environment --repo-id <repo-id>`

elyezer:
Merged as 739040518abd153e4deed086361279707056f699 with a follow up commit da2e3ec1b82037aa492462082b02e639d3996de8

Issue #336, opened 2016-07-21
Test uploading DRPMs


BrnoPCmaniak:
Tests for Uploading DRPMs from [story 1806](https://pulp.plan.io/issues/1806).

API
===

Steps:

1. Create a repository
2. Upload DRPM

Assert:

1. DRPM is in content retrieval of the repo.
2. DRPM has a correct name

CLI
===

## Test Upload
Steps:

1. Create repository
2. Upload DRPM

Assert:

output of `pulp-admin rpm repo content drpm` matches:

```
Arch:         noarch
Checksum:     a498a1d973f6e8550cf2fb863a2640b92dc4ba3ed5fcb26e3b123ce04bd39723
Checksumtype: sha256
Epoch:        0
Filename:     drpms/yum-3.2.29-20.fc16_from_el6_3.4.3-8.fc16.noarch.drpm
Release:      8.fc16
Version:      3.4.3


```

## Test skip existing

Steps:
1. Create Repository
2. Upload DRPM
3. Upload DRPM again with flag `--skip-existing`

Assert:
`No files eligible for upload` in output of step 3

I'm trying to write a PR myself.

elyezer:
@BrnoPCmaniak great.

A little tip, currently on https://github.com/PulpQE/pulp-fixtures we don't have any DRPM fixtures, if you need some binary files to upload the fixtures is the place.

Ichimonji10:
https://github.com/PulpQE/pulp-fixtures/issues/11

Ichimonji10:
Fixed by https://github.com/PulpQE/pulp-smash/pull/337

PR #337, opened 2016-07-21
Issue #336 - Test uploading DRPMs

BrnoPCmaniak:
Added:
 * pulp_smash.constans - DRPM, DRPM_URL
 * pulp_smash.tests.rpm.api_v2.test_upload_publish.UploadDrpmTestCase - API
 * pulp_smash.tests.rpm.cli.test_upload - Client
 * Added docs for pulp_smash.tests.rpm.cli.test_upload

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -56,6 +56,15 @@ DRPM_FEED_URL = (
 )
 """The URL to an DRPM repository."""
 
+DRPM = 'yum-3.4.3-10.fc16_3.4.3-11.fc16.noarch.drpm'
+"""The name of an DRPM file. See :data:`pulp_smash.constants.DRPM_URL`."""
+
+DRPM_URL = urljoin(urljoin(DRPM_FEED_URL, 'drpms/'), DRPM)
+"""The URL to an DRPM file.
+
+Built from :data:`DRPM_FEED_URL`, '/drpms/' and :data:`DRPM`.
+"""
+
 ERROR_KEYS = frozenset((
     '_href',
     'error',
--- a/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
@@ -13,7 +13,12 @@ from __future__ import unicode_literals
 
 from itertools import product
 
+import os
+
+import unittest2
+
 from pulp_smash import api, config, selectors, utils
+from pulp_smash.utils import upload_import_unit
 from pulp_smash.compat import urljoin
 from pulp_smash.constants import (
     CALL_REPORT_KEYS,
@@ -22,6 +27,8 @@ from pulp_smash.constants import (
     RPM,
     RPM_FEED_URL,
     RPM_URL,
+    DRPM_URL,
+    DRPM
 )
 from pulp_smash.tests.rpm.utils import gen_erratum
 from pulp_smash.tests.rpm.api_v2.utils import (
@@ -147,6 +154,61 @@ def _get_pkglist(update):
                 checksum.attrib['type'], checksum.text
             ]
     return out_pkglist
+
+
+class UploadDrpmTestCase(utils.BaseAPITestCase):
+    """Test whether one can upload and associate DRPMs.
+
+    Steps:
+
+    1. Create a RPM repo.
+    2. Upload DRPM to repo.
+    3. Check if DRPM has been uploaded.
+
+
+    This test case targets
+    `Pulp Smash #336 <https://github.com/PulpQE/pulp-smash/issues/336>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Upload DRPM to a repo."""
+        super(UploadDrpmTestCase, cls).setUpClass()
+
+        if selectors.bug_is_untestable(1806, cls.cfg.version):
+            raise unittest2.SkipTest('https://pulp.plan.io/issues/1806')
+
+        # init repo and client
+        client = api.Client(cls.cfg, api.json_handler)
+        repo = client.post(REPOSITORY_PATH, gen_repo())
+        cls.resources.add(repo['_href'])
+        client.response_handler = api.safe_handler
+
+        # Download DRPM
+        cls.drpm = utils.http_get(DRPM_URL)
+
+        upload_import_unit(cls.cfg, cls.drpm, 'drpm', repo['_href'])
+
+        # Search for drpm to determine if is present
+        body = {'criteria': {}}
+        cls.repo_units = client.post(
+            urljoin(repo['_href'], 'search/units/'),
+            body
+        )
+
+    def test_status_code_repo_units(self):
+        """Verify the HTTP status code for repo units response."""
+        self.assertEqual(self.repo_units.status_code, 200)
+
+    def test_drpm_uploaded_successfully(self):
+        """Test if DRPM has been uploaded successfully."""
+        self.assertEqual(len(self.repo_units.json()), 1)
+
+    def test_drpm_file_name_is_correct(self):
+        """Test if DRPM extracted correct metadata for creating filename."""
+        fname = self.repo_units.json()[0]['metadata']['filename']
+        expected_filename = os.path.join('drpms/', DRPM)
+        self.assertEqual(fname, expected_filename)
 
 
 class UploadRpmTestCase(utils.BaseAPITestCase):
--- a/pulp_smash/tests/rpm/cli/test_upload.py
+++ b/pulp_smash/tests/rpm/cli/test_upload.py
@@ -1,0 +1,109 @@
+# coding=utf-8
+"""Tests upload of units via client."""
+from __future__ import unicode_literals
+
+import os
+
+import unittest2
+
+from pulp_smash import cli, config, utils, selectors
+from pulp_smash.constants import DRPM_URL, DRPM
+from pulp_smash.tests.rpm.utils import set_up_module
+
+
+def setUpModule():  # pylint:disable=invalid-name
+    """Execute ``pulp-admin login`` on the target Pulp system."""
+    set_up_module()
+    utils.pulp_admin_login(config.get_config())
+
+
+class UploadDrpmTestCase(unittest2.TestCase):
+    """Test whether one can upload DRPMs."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a repository and download temporary drpm."""
+        super(UploadDrpmTestCase, cls).setUpClass()
+
+        cls.cfg = config.get_config()
+
+        if selectors.bug_is_untestable(1806, cls.cfg.version):
+            raise unittest2.SkipTest('https://pulp.plan.io/issues/1806')
+
+        cls.repo_id = utils.uuid4()
+        cls.client = cli.Client(cls.cfg)
+
+        cls.client.run(
+            'pulp-admin rpm repo create --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+
+        completed_process = cls.client.run(
+            'mktemp --directory'.split()
+        )
+        cls.temp_folder = completed_process.stdout.strip()
+        cls.temp_drpm = os.path.join(cls.temp_folder, DRPM)
+
+        cls.client.run(
+            'curl -o {} {}'
+            .format(cls.temp_drpm, DRPM_URL).split()
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the repository and drpm created by :meth:`setUpClass`."""
+        super(UploadDrpmTestCase, cls).tearDownClass()
+
+        cls.client.run(
+            'pulp-admin rpm repo delete --repo-id {}'
+            .format(cls.repo_id).split()
+        )
+        cls.client.run('pulp-admin orphan remove --all'.split())
+        cls.client.run('rm -rf {}'.format(cls.temp_folder).split())
+
+    def test_upload_and_output(self):
+        """Test upload of DRPM.
+
+        Steps
+
+        1. Upload DRPM
+        2. Check output of ``pulp-admin rpm repo content drpm``
+
+        This test case targets
+        `Pulp Smash #336 <https://github.com/PulpQE/pulp-smash/issues/336>`_
+        """
+        self.client.run(
+            'pulp-admin rpm repo uploads drpm --repo-id {} -f {}'
+            .format(self.repo_id, self.temp_drpm).split()
+        )
+        completed_process = self.client.run(
+            'pulp-admin rpm repo content drpm --repo-id {} --fields filename'
+            .format(self.repo_id).split()
+        )
+
+        filename = completed_process.stdout.split('Filename:')[1].strip()
+
+        self.assertEqual(filename, 'drpms/' + DRPM)
+
+    def test_upload_skip_existing(self):
+        """Test duplicit upload of DRPM with flag ``--skip-existing``.
+
+        Steps
+
+        1. Upload DRPM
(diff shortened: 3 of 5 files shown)


Ichimonji10:
On a related note, DRPMs are not present in pulp fixtures. Thus, it's appropriate that the existing constants be used and expanded. See https://github.com/PulpQE/pulp-fixtures/issues/11.

Ichimonji10:
@BrnoPCmaniak You can avoid virtually all Travis errors by executing `make all` locally, preferentially under both Python 2 and 3. Doing this is often faster than pushing a branch to github and waiting for Travis.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +149,78 @@ def _get_pkglist(update):
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Upload DRPM to a repo.
+
+        Steps:
+
+        1. Create a RPM repo.
+        2. Upload DRPM to repo.
+        3. Check if DRPM has been"uploaded.
```
Ichimonji10: `s/"//`

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +149,78 @@ def _get_pkglist(update):
...
+
+        https://github.com/PulpQE/pulp-smash/issues/336
```
Ichimonji10: Writing something like:

    This test case targets `Pulp Smash #336`_.

...will save future readers a bit of time. Doing this means that the user can know something about the linked-to information without having to context-switch and move to a web browser.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +149,78 @@ def _get_pkglist(update):
...
+        # Remove upload task
+        cls.responses['free'] = client.delete(
+            cls.responses['malloc'].json()['_href'],
+        )
+
+        # Search for drpm to determine if is present
+        body = {'criteria': {}}
+        cls.responses['repo units'] = client.post(urljoin(repo['_href'], 'search/units/'), body)
+
+    def test_status_code(self):
```
Ichimonji10: I think there's already a test case which verifies the status codes returned by each of these actions.

Please drop this test method entirely, and use the `pulp_smash.utils.upload_import_unit` function instead.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +149,78 @@ def _get_pkglist(update):
...
+                self.assertEqual(self.responses[step].status_code, code)
+
+    def test_drpm_uploaded_successfully(self):
+        """Test if DRPM has been uploaded successfully."""
+        self.assertEqual(len(self.responses['repo units'].json()), 1)
+
+    def test_drpm_file_name_is_correct(self):
+        """Test if DRPM extracted correctly metadata for createing filename."""
+        fname = self.responses['repo units'].json()[0]['metadata']['filename']
+        self.assertEqual(fname, 'drpms/yum-3.4.3-10.fc16_3.4.3-11.fc16.noarch.drpm')
```
Ichimonji10: Can we use a constant here instead of hard-coding this string?

pulp_smash/constants.py
```diff
@@ -56,6 +56,12 @@
 )
 """The URL to an DRPM repository."""
 
+DRPM = 'yum-3.4.3-10.fc16_3.4.3-11.fc16.noarch.drpm'
+"""The name of an DRPM file. See :data:`pulp_smash.constants.DRPM_URL`."""
+
+DRPM_URL = urljoin(urljoin(DRPM_FEED_URL, 'drpms/'), DRPM)
+"""The URL to an RPM file. Built from :data:`DRPM_FEED_URL`, '/drpms/' and :data:`RPM`."""
```
Ichimonji10: The URL to an **RPM** file. Built from `DRPM_FEED_URL`, `/drpms/` and **RPM**.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,119 @@
...
+from pulp_smash.tests.rpm.utils import set_up_module
+
+
+def setUpModule():  # pylint:disable=invalid-name
+    """Execute ``pulp-admin login`` on the target Pulp system."""
+    set_up_module()
+    utils.pulp_admin_login(config.get_config())
+
+
+class UploadBaseTestCase(unittest2.TestCase):
```
Ichimonji10: Is an abstract base class needed, given that there's exactly one child of this class? Let's add complexity when it's needed. Doing so makes the code easier to understand and maintain in its current state, and it makes the code easier to refactor in the future.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,119 @@
...
+
+        cli.Client(cls.cfg).run(
+            'curl -o {} {}'
+            .format(cls.temp_drpm, DRPM_URL).split()
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the drpm created by :meth:`setUpClass`."""
+        cli.Client(cls.cfg).run('rm -rf {}'.format(cls.temp_folder).split())
```
Ichimonji10: This doesn't call super. See what I mean about parent classes adding complexity? :wink:

pulp_smash/tests/rpm/cli/test_upload.py
BrnoPCmaniak: I thought that maybe somebody would want to add RPM upload test in the future.

Ichimonji10: > in the future.

:wink:

BrnoPCmaniak:
Ready for second round.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +150,61 @@ def _get_pkglist(update):
...
+
+
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Upload DRPM to a repo.
+
+        This test case targets `Pulp Smash #336`.
+        https://github.com/PulpQE/pulp-smash/issues/336
```
Ichimonji10: Here's one way to write a link in reStructuredText:

    `my link <http://example.com>`_

Here's another way:

    `my link`_

    .. _my link: http://example.com

See [A reStructuredText Primer](http://docutils.sourceforge.net/docs/user/rst/quickstart.html).

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +150,61 @@ def _get_pkglist(update):
...
+        self.assertEqual(self.repo_units.status_code, code)
+
+    def test_drpm_uploaded_successfully(self):
+        """Test if DRPM has been uploaded successfully."""
+        self.assertEqual(len(self.repo_units.json()), 1)
+
+    def test_drpm_file_name_is_correct(self):
+        """Test if DRPM extracted correct metadata for creating filename."""
+        fname = self.repo_units.json()[0]['metadata']['filename']
+        excepted_filename = 'drpms/yum-3.4.3-10.fc16_3.4.3-11.fc16.noarch.drpm'
```
Ichimonji10: Magic string. :frowning:

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,114 @@
...
+                           'Arch:         noarch\n'
+                           'Checksum:     e348519427b367a13f1898f3fc02235acb7'
+                           '56cfdb834b3ca391416216b423a9e\n'
+                           'Checksumtype: sha256\n'
+                           'Epoch:        0\n'
+                           'Filename:     drpms/yum-3.4.3-10.fc16_3.4.3-11.fc'
+                           '16.noarch.drpm\n'
+                           'Release:      11.fc16\n'
+                           'Version:      3.4.3\n'
+                           '\n\n')
```
Ichimonji10: Don't you hate how pulp-fixtures hard-wraps text?

The `--fields` option can make output more predictable. It also helps to eliminate issues due to silly things like whitespace differences. This should be taken care of before merging, either in this commit or in a reviewer's follow-up commit.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Steps
+
+        1. Upload DRPM
+        2. Upload again with flag and check if it wasn't uploaded.
+
+        https://github.com/PulpQE/pulp-smash/issues/336
+        """
+        cli.Client(self.cfg).run(
+            'pulp-admin rpm repo uploads drpm --repo-id {} -f {}'
```
Ichimonji10: What does `-f` do? I'd recommend expanding this to the full option name, either in this commit or in the reviewer's follow-up commit.

pulp_smash/tests/rpm/cli/test_upload.py
Ichimonji10: Also, magic strings. :frowning:

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -22,7 +23,7 @@
     RPM,
     RPM_FEED_URL,
     RPM_URL,
-)
+    DRPM_URL)
```
Ichimonji10: This should be updated to the canonical style of:

```
from flizbang import (
    bar,
    baz,
    biz,
    foo,
)
```

Can happen in follow up commit.

Ichimonji10:
At this point, the PR looks pretty good. @BrnoPCmaniak, don't feel obliged to handle the current comments. If you do - great! But they can also be handled by the reviewer.

BrnoPCmaniak:
I fixed that too so now should be everything good.

Ichimonji10:
Related: https://pulp.plan.io/issues/1806

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +154,59 @@ def _get_pkglist(update):
...
+        body = {'criteria': {}}
+        cls.repo_units = client.post(
+            urljoin(repo['_href'], 'search/units/'),
+            body
+        )
+
+    def test_status_code_repo_units(self):
+        """Verify the HTTP status code for repo units response."""
+        code = 200
+        self.assertEqual(self.repo_units.status_code, code)
```
elyezer: Avoid single use variables. Drop code variable and use 200 on the line below.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,105 @@
...
+    """Test whether one can upload DRPMs."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a repository and download temporary drpm."""
+        super(UploadDrpmTestCase, cls).setUpClass()
+
+        cls.cfg = config.get_config()
+        cls.repo_id = utils.uuid4()
+        cli.Client(cls.cfg).run(
```
elyezer: Instead of instantiating a cli.Client every time, create a class client and then reuse it for every run call.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,105 @@
...
+        cli.Client(self.cfg).run(
+            'pulp-admin rpm repo uploads drpm --repo-id {} -f {}'
+            .format(self.repo_id, self.temp_drpm).split()
+        )
+        completed_process = cli.Client(self.cfg).run(
+            'pulp-admin rpm repo content drpm --repo-id {} --fields filename'
+            .format(self.repo_id).split()
+        )
+
+        excepted_output = (
```
elyezer: s/excepted/expected?

That is what you are expecting right?

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +154,59 @@ def _get_pkglist(update):
...
+
+    def test_drpm_uploaded_successfully(self):
+        """Test if DRPM has been uploaded successfully."""
+        self.assertEqual(len(self.repo_units.json()), 1)
+
+    def test_drpm_file_name_is_correct(self):
+        """Test if DRPM extracted correct metadata for creating filename."""
+        fname = self.repo_units.json()[0]['metadata']['filename']
+        excepted_filename = os.path.join('drpms/', DRPM)
```
elyezer: s/excepted/expected?

pulp_smash/tests/rpm/cli/test_upload.py
Ichimonji10:
```python
client = cli.Client(cls.cfg)
client.run(...)
client.run(...)
```

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -149,6 +154,59 @@ def _get_pkglist(update):
...
+
+    This test case targets
+    `Pulp Smash #336 <https://github.com/PulpQE/pulp-smash/issues/336>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Upload DRPM to a repo."""
+        super(UploadDrpmTestCase, cls).setUpClass()
+
```
Ichimonji10: This test fails when run against Pulp systems, regardless of whether they've been installed from source, the nightly RPMs, or from official RPMs. Something like this will conditionally skip the test:

```python
if selectors.bug_is_untestable(1806, cls.cfg.version):
    raise unittest2.SkipTest('https://pulp.plan.io/issues/1806')
```

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -0,0 +1,105 @@
...
+            '\x1b[0mFilename:     %s' % os.path.join('drpms/', DRPM)
+        )
+
+        self.assertEqual(completed_process.stdout.strip(), excepted_output)
```
Ichimonji10: FYI, this is a more resilient approach:

```python
filename = completed_process.stdout.split('Filename:')[1].strip()
self.assertEqual(filename, 'drpms/' + DRPM)
```

The `[0m` bit of code is an ANSI color code, and is one of those things that could reasonably appear or disappear at any time.

BrnoPCmaniak:
Comments fixed

Ichimonji10:
@BrnoPCmaniak, this looks good.

Can you verify whether https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:pr/337 still passes? Or, can you email me info for accessing a system with the necessary commits applied?

BrnoPCmaniak:
@Ichimonji10  Yes when I comment out the skipping part it works. 

```
(pulp-smash)[vagrant@dev pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_upload_publish
................
Ran 16 tests in 57.616s

OK
(pulp-smash)[vagrant@dev pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.cli.test_upload
.
Ran 1 test in 5.276s

OK
```

Also the `make all` works.

Ichimonji10:
Rebased and merged as 705dee0e6371966ad33024836d3e7fe1b7093979, with a follow-up commit of 7cbb369f8b89222b71d8eb81e1012e2b8b5dcabb. Thank you, @BrnoPCmaniak. That was easy. :smile:

PR #346, opened 2016-07-26
Convert documentation build warnings into errors

Ichimonji10:
The `docs-html` make target uses `sphinx-build` to generate
documentation. Add the `-W` option, which turns warnings into errors.
This will help keep syntactic mistakes, like broken links and
incorrectly formatted code blocks, out of the documentation.


Ichimonji10:
Travis executes the `all` target during each build. This commit will thus cause Travis to (finally!) warn us about syntactic documentation issues.

PR #348, opened 2016-08-02
Changes the expected message when no-op publish occurs.

dkliban:
After Pulp was modified to support using docker rsync distributor as a predistributor
for docker web distributor, Pulp skips publishing for two different reasons. As the
result of this change, the task status for skipped publishes has changed. This commit
adjusts pulp-smash tests to expect the new string.


elyezer:
Added a follow up commit 2572fc67c7a2dbca97da03986bccf47fbf511c6c to proper select the summary message depending on the Pulp version.

Ichimonji10:
:+1:

Issue #349, opened 2016-08-04
test_download_policies tests failing


seandst:
Recent smash runs in jenkies have been failing in the test_download_policies modules with tracebacks like this:
```
Traceback (most recent call last):
  File "/home/jenkins/workspace/pulp-smash-runner/pulp-smash/pulp_smash/tests/rpm/api_v2/test_download_policies.py", line 99, in setUpClass
    cls.rpm = client.get(path)
  File "/home/jenkins/workspace/pulp-smash-runner/pulp-smash/pulp_smash/api.py", line 259, in get
    return self.request('GET', url, **kwargs)
  File "/home/jenkins/workspace/pulp-smash-runner/pulp-smash/pulp_smash/api.py", line 319, in request
    requests.request(method, **request_kwargs),
  File "/home/jenkins/workspace/pulp-smash-runner/pulp-smash/pulp_smash/api.py", line 108, in safe_handler
    response.raise_for_status()
  File "/home/jenkins/workspace/pulp-smash-runner/venv/lib/python2.7/site-packages/requests/models.py", line 844, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
[...]

PR #351, opened 2016-08-09
Skip tests failing due to Pulp issue 2144

Ichimonji10:
According to Pulp issue 2144, Nectar throwing tracebacks due to an
unbound method being called. As a result, tests in module
`pulp_smash.tests.rpm.api_v2.test_download_policies` fail. Skip these
tests until the issue is fixed.

Fixes https://github.com/PulpQE/pulp-smash/issues/349

See https://pulp.plan.io/issues/2144


Ichimonji10:
I've manually tested this change.

PR #354, opened 2016-08-11
Skip ForceSyncTestCase as appropriate

### LINKED ISSUE #353: --force-full not supported on sync until 2.10
2.9 added the ability to --force-full when publishing a repo in https://pulp.plan.io/issues/1158, but the corresponding ability to --force-full when *syncing* a repo arrives in 2.10, with the completion of https://pulp.plan.io/issues/1982.

pulp_smash.tests.rpm.cli.test_sync.ForceSyncTestCase.test_force_sync is currently failing against 2.9 because this feature is not supported on that release.

Ichimonji10:
`ForceSyncTestCase` tests whether it's possible for a user to force Pulp
to perform a full sync. This feature is detailed in Pulp issue 1982. [1]
This test case should only run on Pulp 2.10.0 and above, as outlined by
that issue. Conditionally skip the test case so that this is so.

Fix #353

[1] https://pulp.plan.io/issues/1982

### DIFF
--- a/pulp_smash/tests/rpm/cli/test_sync.py
+++ b/pulp_smash/tests/rpm/cli/test_sync.py
@@ -6,7 +6,7 @@ import random
 
 import unittest2
 
-from pulp_smash import cli, config, utils
+from pulp_smash import cli, config, selectors, utils
 from pulp_smash.constants import RPM_FEED_URL
 from pulp_smash.tests.rpm.utils import set_up_module
 from pulp_smash.utils import is_root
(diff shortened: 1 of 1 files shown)


Ichimonji10:
While implementing this change, I realized that there's a few less-critical issues with `ForceSyncTestCase`. I'll make a follow-up PR with those less-critical changes.

Ichimonji10:
I've manually tested this change, and it's OK. In addition, this change is fairly high priority. I'm putting on the BDFL hat and merging this. :smile:

elyezer:
:+1:

PR #355, opened 2016-08-11
Refactor ForceSyncTestCase

Ichimonji10:
Do at least the following:

* Make the test case docstring reference correct Pulp and Pulp Smash
  issues.
* Make the test case docstring correctly describe what the test case
  does.
* Greatly simplify the test case implementation. Formerly, the test case
  defined `setUpClass` and `tearDownClass` methods, among others,
  despite having just one `test*` method and inheriting directly from
  `unittest2.TestCase`. Now, the test case has a single `test*` method
  and a simple classmethod.

### DIFF
--- a/pulp_smash/tests/rpm/cli/test_sync.py
+++ b/pulp_smash/tests/rpm/cli/test_sync.py
@@ -105,57 +105,57 @@ class RemovedContentTestCase(unittest2.TestCase):
 
 
 class ForceSyncTestCase(unittest2.TestCase):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I can confirm that this change is still valid. I've run these refactored tests against a pair of Pulp 2.9 and 2.10 systems running on Fedora 23 and provisioned this morning, and they still pass.

We're short on manpower right now. Merging.

elyezer:
:+1:

PR #357, opened 2016-08-15
Rename modules containing new signature tests

Ichimonji10:
Rename two modules from:

    pulp_smash/tests/rpm/api_v2/test_signature_checks_for_uploads.py
    pulp_smash/tests/rpm/api_v2/test_packages_signature.py

To:

    pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_uploads.py
    pulp_smash/tests/rpm/api_v2/test_signatures_saved_for_packages.py

These two modules contain the newly-added package signature tests. This
change is intended to:

* Ensure that modules clearly describe the tests they contain.
* Provide an obvious naming scheme for additional signature tests. For
  example, it's feasible that we might also add a module named
  `test_signatures_checked_for_syncs` and
  `test_signatures_checked_for_copies`.

These modules were recently added. See:

* 81c509f "Add tests for signature check on packages and RPM repos."
* f663905 "Update newly added signature tests"
* 7779cea "Test signature checks when uploading packages"


elyezer:
Much better! Thank you

PR #359, opened 2016-08-15
Drop unnecessary usage of sudo

Ichimonji10:
`sudo` is not guaranteed to be present on a system, and even if it is,
there may be restrictions on its use. For example, consider this
interaction with an RHEL 7.2 system I recently laid my hands on:

    $ ssh $hostname sudo -u apache pulp-manage-db
    sudo: sorry, you must have a tty to run sudo

In this case, it is better to use `runuser` than `sudo`.


elyezer:
`runuser` is new to me, interesting.

Ichimonji10:
Aye. I didn't know about it either, but `sudo`'s man page referenced it. It seems to be a core utility. It's available on the RHEL 7.2 system mentioned, it's available on an F23 VM I have, and it's provided by Arch's "base" package group.

Issue #264, opened 2016-05-20
 Test rsync distributor can be used to publish RPM


preethit:
https://pulp.plan.io/issues/1759
The rsync_distributor will afford users of RPM's yum_distributor to publish repositories published by the yum_distributor to any server that supports key based ssh communication. The rsync_distributor will limit the units published by only publishing units associated with the repository before the date of the last publish of the yum_distributor.

The order of publishes for Docker will be different. The rsync_distributor will first publish content to any server that supports key based ssh communication. Then the web_distributor will publish the metadata about the content to some other web servers.

The rsync_distributor [for docker] will have an optional configuration option to 'update_web_distributor'. When set to True, the publish will update the 'web_distributor' associated with the same repository. The update will set the web_distributor's 'end_date' to the start time of the rsync_distributor's publish. This will limit which units are published by the web_distributor as described in #1887. This will prevent Pulp from publishing docker metadata for units that have not yet been published by the rsync_distributor.

1) Rsync content needed for the publish
2) Rewrite symlinks to be relative symlinks
3) Deduplicate the remote filesystem (make sure that content is present once and only symlinks live in published directories)
4) Ensure paths to published content on remote system are in proper order

Ichimonji10:
I'm assigning this to the July 29th milestone. I think this is issue is important in part because it's a brand new feature that's likely to land in Pulp 2.9 and in part because of RCM's involvement in testing this feature. (See #320.)

I'm not assigning this task to the current milestone of July 15th. This feature has an extensive description and conversation log on pulp.plan.io, and from my reading of the issue, it will take a fair amount of time to fully understand the feature and write a correct test. I could likely get this done by the 15th, but I want to get some other tasks done between now and then.

Ichimonji10:
The developers have shared their test plan with us. See: https://pulp.plan.io/issues/1759#note-20

Ichimonji10:
This is a little note to myself. I'm dumping it here for convenience.

        # TODO: Rename test case to FastForwardTestCase. Add these steps.
        #
        # 4. Remove all files from the target directory. Publish with the rpm
        #    rsync distributor, and verify that no files are in the target
        #    directory. (This tests Pulp's fast-forward logic.)
        # 5. Perform a full publish with the rpm rsync distributor. Verify that
        #    the correct files are in in the target directory.

Ichimonji10:
This is a note to self, listing some additional steps to be added to `RemoteUnitsPathTestCase`:

[...]

PR #360, opened 2016-08-16
Add basic tests for the rsync rpm distributor

Ichimonji10:
Implement two tests for the new RPM rsync distributor.

* `PublishTestCase` ensure that the RPM rsync distributor works in the
  most basic sense. It doesn't do anything fancy like testing
  fast-forward logic or perform multiple publishes.
* `PublishBeforeYumDistTestCase` ensures that publishing with the RPM
  rsync distributor fails if the yum distributor hasn't yet been used
  for a publish.

See: https://github.com/PulpQE/pulp-smash/issues/264

See: https://pulp.plan.io/issues/1759

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -1,0 +1,255 @@
+# coding=utf-8
+"""Test the functionality of the RPM rsync distributor.
+
+The RPM rsync distributor lets one publish content units to a directory via
+rsync+ssh. A typical usage of the RPM rsync distributor is as follows:
+
+1. Create RPM repository with yum and RPM rsync distributors.
+2. Upload some content units to the repository.
+3. Publish the repository with the yum distributor.
+4. Publish the repository with the RPM rsync distributor.
+
+The RPM rsync distributor may not be used by itself. One cannot create an RPM
+repository with just an RPM rsync distributor; and one cannot publish a
+repository with the RPM rsync distributor without first publishing with a yum
+distributor.
+
+For more information on the RPM rsync distributor, see `Pulp #1759`_.
+
+.. _Pulp #1759: https://pulp.plan.io/issues/1759
+"""
+from __future__ import unicode_literals
+
+import unittest2
+
+from pulp_smash import api, cli, config, exceptions, selectors, utils
+from pulp_smash.compat import urljoin, urlparse
+from pulp_smash.constants import REPOSITORY_PATH, RPM_FEED_URL
+from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
+from pulp_smash.tests.rpm.utils import set_up_module
+
+
+def setUpModule():  # pylint:disable=invalid-name
+    """Conditionally skip the tests in this module.
+
+    Skip the tests in this module if:
+
+    * The RPM plugin is not installed on the target Pulp server.
+    * `Pulp #1759`_ is not implemented on the target Pulp server.
+
+    .. _Pulp #1759: https://pulp.plan.io/issues/1759
+    """
+    set_up_module()
+    if selectors.bug_is_untestable(1759, config.get_config().version):
+        raise unittest2.SkipTest('https://pulp.plan.io/issues/1759')
+
+
+def _make_user(cfg):
+    """A generator to create a user account on the target system.
+
+    Yield a username and private key. The corresponding public key is made the
+    one and only key in the user's ``authorized_keys`` file.
+
+    The username and private key are yielded one-by-one rather than as a pair
+    because the user creation and key creation steps are executed serially.
+    Should the latter fail, the calling function will still be able to delete
+    the created user.
+
+    The user is given a home directory. When deleting this user, make sure to
+    pass ``--remove`` to ``userdel``. Otherwise, the home directory will be
+    left in place.
+    """
+    client = cli.Client(cfg)
+    sudo = '' if utils.is_root(cfg) else 'sudo '
+
+    # According to useradd(8), usernames may be up to 32 characters long. But
+    # long names break the rsync publish process: (SNIP == username)
+    #
+    #     unix_listener:
+    #     "/tmp/rsync_distributor-[SNIP]@example.com:22.64tcAiD8em417CiN"
+    #     too long for Unix domain socket
+    #
+    username = utils.uuid4()[:12]
+    cmd = 'useradd --create-home {0}'
+    client.run((sudo + cmd.format(username)).split())
+    yield username
+
+    cmd = 'runuser -u {0} -- ssh-keygen -N . -f /home/{0}/.ssh/mykey'
+    cmd = (sudo + cmd.format(username)).split()
+    cmd[cmd.index('-N') + 1] = ''
+    client.run(cmd)
+    cmd = 'cp /home/{0}/.ssh/mykey.pub /home/{0}/.ssh/authorized_keys'
+    client.run((sudo + cmd.format(username)).split())
+    cmd = 'cat /home/{0}/.ssh/mykey'
+    private_key = client.run((sudo + cmd.format(username)).split()).stdout
+    yield private_key
+
+
+class _MakeUserMixin(object):  # pylint:disable=too-few-public-methods
+    """A mixin providing the ability to create a temporary user.
+
+    This mixin requires that the ``unittest.TestCase`` class from the standard
+    library is a parent class.
+    """
+
+    def make_user(self, cfg):
+        """Create a user account with a home directory and an SSH keypair.
+
+        In addition, schedule the user for deletion with ``self.addCleanup``.
+
+        :param pulp_smash.config.ServerConfig cfg: Information about the server
+            being targeted.
+        :returns: A ``(username, private_key)`` tuple.
+        """
+        sudo = '' if utils.is_root(cfg) else 'sudo '
+        creator = _make_user(cfg)
+        username = next(creator)
+        self.addCleanup(
+            cli.Client(cfg).run,
+            (sudo + 'userdel --remove {}').format(username).split()
+        )
+        private_key = next(creator)
+        return (username, private_key)
+
+    def write_private_key(self, cfg, private_key):
+        """Write the given private key to a file on disk.
+
+        Ensure that the file is owned by user "apache" and has permissions of
+        ``600``. In addition, schedule the key for deletion with
+        ``self.addCleanup``.
+
+        :param pulp_smash.config.ServerConfig cfg: Information about the server
+            being targeted.
+        :returns: The path to the private key on disk, as a string.
+        """
+        sudo = '' if utils.is_root(cfg) else 'sudo '
+        client = cli.Client(cfg)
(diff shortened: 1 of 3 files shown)


Ichimonji10:
These tests are *not* suitable for merging yet. They're here for ease of reference.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+    because the user creation and key creation steps are executed serially.
+    Should the latter fail, the calling function will still be able to delete
+    the created user.
+
+    The user is given a home directory. When deleting this user, make sure to
+    pass ``--remove`` to ``userdel``. Otherwise, the home directory will be
+    left in place.
+    """
+    client = cli.Client(cfg)
+    sudo = '' if utils.is_root(cfg) else 'sudo '
```
elyezer: I've seen this logic on every place SSH is used, we should think on a way to make it generic or an util function like `sudo = utils.sudo(cfg)` or maybe some argument to `SSHClient.run`, something like `client.run([...], sudo=True)`. Just leaving some thoughts for later improvement.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+
+    username = utils.uuid4()[:32]  # See useradd(8)
+    cmd = 'useradd --create-home {0}'
+    client.run((sudo + cmd.format(username)).split())
+    yield username
```
elyezer: :100:

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+    client.run(cmd)
+    cmd = 'cp /home/{0}/.ssh/mykey.pub /home/{0}/.ssh/authorized_keys'
+    client.run((sudo + cmd.format(username)).split())
+    cmd = 'cat /home/{0}/.ssh/mykey.pub'
+    private_key = client.run((sudo + cmd.format(username)).split()).stdout
+    yield private_key
+
+
+class _MakeUserMixin(object):  # pylint:disable=too-few-public-methods
+    """A mixin providing the ability to create a temporary.
```
elyezer: s/a temporary/a temporary user?

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+        """Create a user and schedule it for deletion.
+
+        :param pulp_smash.config.ServerConfig cfg: Information about the server
+            being targeted.
+        :returns: A ``(username, private_key)`` tuple.
+        """
+        sudo = '' if utils.is_root(cfg) else 'sudo '
+        creator = _make_user(cfg)
+        username = next(creator)
+        self.addCleanup(
```
elyezer: We can ensure the parent being an `unittest.TestCase` by checking if `hasattr(self, 'addCleanup')` raising an exception. This is not required but may be useful if we need the temporary user logic on other places.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+        utils.sync_repo(self.cfg, repo_href)
+
+        # Publish the repo, and verify its files are present.
+        api_client.post(urljoin(repo_href, 'actions/publish/'), {
+            'id': body['distributors'][1]['distributor_id']
+        })
+
+        cmd = (sudo + 'ls -1 /home/{}'.format(ssh_user))
+        dirs = set(cli_client.run(cmd.split()).stdout.strip().split('\n'))
+        self.assertEqual(dirs, {'content'})
```
elyezer: set literals :heart:

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,228 @@
...
+
+        cmd = (sudo + 'ls -1 /home/{}/content'.format(ssh_user))
+        dirs = set(cli_client.run(cmd.split()).stdout.strip().split('\n'))
+        self.assertEqual(dirs, {'origin'})
+
+        cmd = (sudo + 'ls -1 /home/{}/content/origin'.format(ssh_user))
+        dirs = set(cli_client.run(cmd.split()).stdout.strip().split('\n'))
+        self.assertEqual(dirs, {'rpm'})
+
+        # TODO: Finish steps 2, 3 and 4.
```
elyezer: Will continue the review tomorrow. Will delete this comment when doing so.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
Ichimonji10: Agreed. It's annoying to keep writing this over and over again.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
Ichimonji10: I'm normally wary of adding in these kinds of checks. But it makes more sense here. If `addCleanup` isn't available, I'd rather know before a user is created.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
Ichimonji10: This logic came in very handy while I was working out bugs in the SSH key generation logic, below. Pop quiz: what's wrong with this?

```python
cmd = 'runuser -u alice -- ssh-keygen -N '' -f /home/alice/.ssh/mykey'.split()
client.run(cmd)
```

elyezer: Usage of `--`? That is the only thing that seems to not work. Also the plumbum may be using `--` already when executing the command.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
elyezer: Not sure if linters will be able to catch this, need to test.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
Ichimonji10: Hehe. The `--` is actually necessary, as it tells `runuser` to stop processing arguments.

No, the real issue here is that `.split()` produces the following list:

```python
['runuser', '-u', 'alice', '--', 'ssh-keygen', '-N', '-f', '/home/alice/.ssh/mykey']
```

Plumbum then passes `''` as the *literal* argument to `-N`, thus instructing `ssh-keygen` to make a key with the passphrase `''`.

The above translates to the following bash command:

```bash
runuser -u alice -- ssh-keygen -N "''" -f /home/alice/.ssh/mykey
```

Ichimonji10:
PR updated. When run against 2.9, the tests skip:

```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_rsync_distributor
s
Ran 0 tests in 1.721s

OK (skipped=1)
```

Against 2.10, they pass:

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_rsync_distributor
..
Ran 2 tests in 145.139s

OK
```

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
elyezer: Ahh I see your point. I have faced that on automation-tools some time ago...

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,253 @@
+# coding utf-8
```
elyezer: coding utf-8 == coding=utf-8?

Ichimonji10: Whoops!

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,253 @@
...
+        username = next(creator)
+        self.addCleanup(
+            cli.Client(cfg).run,
+            (sudo + 'userdel --remove {}').format(username).split()
+        )
+        private_key = next(creator)
+        return (username, private_key)
+
+    def write_private_key(self, cfg, private_key):
+        """Write the given private key to a publicly-readable file on disk.
```
Ichimonji10: This isn't right. I'll fix it.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,253 @@
...
+        client.machine.session().run(
+            "echo '{}' > {}".format(private_key, ssh_identity_file)
+        )
+        client.run(['chmod', '600', ssh_identity_file])
+        client.run((sudo + 'chown apache ' + ssh_identity_file).split())
+        return ssh_identity_file
+
+
+class PublishBeforeYumDistTestCase(_MakeUserMixin, utils.BaseAPITestCase):
+    """Publish a repository the rsync distributor before the yum distributor.
```
elyezer: Publish a repository with rsync ...?

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,255 @@
...
+        ssh_user, priv_key = self.make_user(self.cfg)
+        ssh_identity_file = self.write_private_key(self.cfg, priv_key)
+
+        # Create a repository with an importer and pair of distributors, and
+        # sync it.
+        api_client = api.Client(self.cfg, api.json_handler)
+        body = gen_repo()
+        body['importer_config']['feed'] = RPM_FEED_URL
+        body['distributors'] = [gen_distributor()]
+        body['distributors'].append({
```
elyezer: This same log was used in the previous test case, we may consider creating a helper function if this is required for any other test case.

Ichimonji10: Yes, I agree. There's some definite repetition going on here.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -0,0 +1,255 @@
...
+        username = next(creator)
+        self.addCleanup(
+            cli.Client(cfg).run,
+            (sudo + 'userdel --remove {}').format(username).split()
+        )
+        private_key = next(creator)
+        return (username, private_key)
+
+    def write_private_key(self, cfg, private_key):
+        """Write the given private key to a file on disk.
```
elyezer: Much better

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
elyezer: Let's have this in mind for future improvements. The PR looks good and we can refactor this later.

Issue #325, opened 2016-06-30
Test that multiple checksums for packages in the erratum are handled correctly


goosemania:
In Pulp 2.9+ multiple checksums for packages in the erratum should be published correctly.
If the checksum of the [distributor checksum_type](http://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/yum-plugins.html#yum-distributor) is available for the package in the erratum pkglist then such checksum will appear in the published updateinfo.xml, otherwise package will be listed with the best(longest) checksum available in the updateinfo.xml.

Example 1:
- distributor checksum_type is 'sha256'
- in the erratum pkglist 'md5' and 'sha256' are available for each package

=> 'sha256' will be used in the updateinfo.xml.

Example 2:
- distributor checksum_type is 'sha1'
- in the erratum pkglist 'md5' and 'sha256' are available for each package 

=> 'sha256' will be used in the updateinfo.xml.

In Pulp 2.10+ a new distributor option `updateinfo_checksum_type` is introduced. If it is specified, such checksum type will be used for packages in the erratum pkglist. Otherwise publish task will fail.

Example 3:
- updateinfo_checksum_type is 'md5'
- distributor checksum_type is 'sha256'
- in the erratum pkglist 'md5' and 'sha256' are available for each package

=> 'md5' will be used in the updateinfo.xml. 

Example 4:
- updateinfo_checksum_type is 'sha1'
- distributor checksum_type is 'sha256'
- in the erratum pkglist 'md5' and 'sha256' are available for each package

=> publish task failed

Redmine issue for handling multiple checksums: https://pulp.plan.io/issues/2020
Redmine story to support `updateinfo_checksum_type`: https://pulp.plan.io/issues/1878
Redmine issue to change behavior in case cheksum of the requested type is unavailable: https://pulp.plan.io/issues/2042

PR #361, opened 2016-08-17
Update tests for erratum checksums

Ichimonji10:
Errata may list one or more checksums for each of the packages they
reference. However, the errata published by pulp list at most one
checksum for each package they reference. As a result, Pulp has to be
smart about which checksum to place in the errata it generates. This
logic is outlined in Pulp issue 2042.

Update the checksum-checking logic in `UploadErratumTestCase`.

See: https://pulp.plan.io/issues/2042

See: https://github.com/PulpQE/pulp-smash/issues/325


Ichimonji10:
@elyezer @smyers

Ichimonji10:
Verified against two-day-old Pulp 2.9.3 and 2.10.0 systems.

PR #362, opened 2016-08-17
Make reset_pulp function compatible with RHEL 6

Ichimonji10:
The `reset_pulp` function executes `runuser -u apache pulp-manage-db`.
This doesn't work on RHEL 6, because RHEL 6 ships with an old version of
runuser that doesn't support `-u`. Work around this limitation.

For exact details, see runuser(1).

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -100,10 +100,20 @@ def reset_pulp(server_config):
         service.stop()
 
     # Reset the database and nuke accumulated files.
+    #
+    # Why use `runuser` instead of `sudo`? Because some systems are configured
+    # to refuse to execute `sudo` unless a tty is present (The author has
+    # encountered this on at least one RHEL 7.2 system.)
+    #
+    # Why not use runuser's `-u` flag? Because RHEL 6 ships an old version of
+    # runuser that doesn't support the flag, and RHEL 6 is a supported Pulp
+    # platform.
     client = cli.Client(server_config)
     prefix = '' if is_root(server_config) else 'sudo '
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I've tested this on a pair of Fedora 23 systems. Can someone test this on RHEL 6?

@elyezer @seandst

pulp_smash/utils.py
```diff
@@ -100,10 +100,20 @@ def reset_pulp(server_config):
...
+    #
+    # Why not use runuser's `-u` flag? Because RHEL 6 ships an old version of
+    # runuser that doesn't support the flag, and RHEL 6 is a supported Pulp
+    # platform.
     client = cli.Client(server_config)
     prefix = '' if is_root(server_config) else 'sudo '
     client.run('mongo pulp_database --eval db.dropDatabase()'.split())
-    client.run((prefix + 'runuser -u apache pulp-manage-db').split())
+    client.run((
+        prefix + 'runuser --shell /usr/bin/sh apache --command pulp-manage-db'
```
elyezer: On RHEL6:

```console
$ sudo runuser --shell /usr/bin/sh apache --command pulp-manage-db
runuser: /usr/bin/sh: No such file or directory
```

seandst: /usr/bin/sh is correct for newer distros, but they provide backward compatibility with a symlink of /bin to /usr/bin, so /bin/sh is the easy compromise if not the correct choice.

Ichimonji10: Will update.

Ichimonji10:
Updated.

PR #363, opened 2016-08-17
Fix a typo on CLI test_sync

elyezer:
Make sure to use sudo when the user is not root.


pulp_smash/tests/rpm/cli/test_sync.py
```diff
@@ -149,7 +149,7 @@ def test_force_sync(self):
         # Delete a random RPM
         rpms = self._list_rpms(cfg)
         client.run('{} rm -rf {}'.format(
-            'sudo' if is_root(cfg) else '',
+            'sudo' if not is_root(cfg) else '',
```
Ichimonji10: /me is embarrased

PR #367, opened 2016-08-22
Refactor and expand SELinux logic

Ichimonji10:
The first commit refactors existing SELinux management logic. It moves the logic out of a `test*` module and into a `utils` module.

The second commit uses the just-refactored SELinux management logic to make a failing test pass.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_export.py
+++ b/pulp_smash/tests/rpm/api_v2/test_export.py
@@ -30,37 +30,12 @@ from pulp_smash.constants import (
     RPM_SHA256_CHECKSUM,
 )
 from pulp_smash.tests.rpm.api_v2.utils import (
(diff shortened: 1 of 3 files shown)


Ichimonji10:
This test passes both when run against Fedora 23 and Fedora 24 Pulp 2.10.0 systems. Here are representative results for one of the systems:

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_{rsync_distributor,export}
..s..s.s......
Ran 13 tests in 637.380s

OK (skipped=3)
```

Ichimonji10:
I made a one-line change to a docstring. I'm going to wait for Travis to finish re-checking this PR just to be sure nothing breaks, but I don't think it needs a re-review.

PR #370, opened 2016-08-25
Un-pin dependency on 'flake8<3'

Ichimonji10:
Allow the current version of flake8 to be installed, instead of some
version before version 3. See:

* b5309f6bab0a99c8a5f5dba70a14a4c5fe19dd16
* https://github.com/zheller/flake8-quotes/issues/29


Ichimonji10:
`make lint-flake8` runs ~20% faster on my system after this change.

PR #371, opened 2016-08-25
Update assertion for PublishBeforeYumDistTestCase

Ichimonji10:
`PublishBeforeYumDistTestCase` creates an RPM repository with both a yum
distributor and an RPM rsync distributor, and it publishes with the
latter before the former. Currently, the test case asserts that the
publish fails. However, according to Pulp issue 2187, this should result
in the publish skipping. Update the test case accordingly.

See: https://pulp.plan.io/issues/2187

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -24,7 +24,7 @@ import os
 import unittest2
 from requests.exceptions import HTTPError
 
-from pulp_smash import api, cli, config, exceptions, selectors, utils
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Test results when run against a Pulp 2.10.1 system:

```
(pulp-smash3) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_rsync_distributor.PublishBeforeYumDistTestCase
.
Ran 1 test in 31.017s

OK
```

Ichimonji10:
This change fixes a failure for each item in our automation matrix.

PR #372, opened 2016-08-25
Drop a spurious reference to `self`

Ichimonji10:
Method `maybe_disable_selinux` accepts a `cfg` parameter. Using this
parameter instead of the instance attribute on the current object lets
it be used in a wider variety of situations, i.e. any class at all
inheriting from `unittest.TestCase`.


elyezer:
Merged as 4e02eed367199797b702734f9e41017cd967a029. Needed to rebase since I merged Dennis PR before this.

PR #374, opened 2016-08-26
Fixes rsync distributor test in situations where SELinux is Disabled

dkliban:
This commit adds a check for the state of SELinux in the rsynx distributor
test. On systems where SELinux is disabled, running chcon produces a non-zero
exit code.


elyezer:
Merged with a follow up commit of 265dd27defa1a431e592747c2cca26d1df02d6b6

PR #375, opened 2016-08-29
Refactor and expand RPM rsync distributor tests

Ichimonji10:
Both commits operate on module `pulp_smash.tests.rpm.api_v2.test_rsync_distributor`.

The first commit refactors the module. When applied, the commit will:

* Make method `make_repo` more powerful. Formerly, callers could only customize the `remote` section of the generated config. Now, callers can customize the parent `distributor_config` section.
* Make method `verify_files_to_dir` more generic, and move it to class `_RsyncDistUtilsMixin`. This change makes the method available to future test cases.

The second commit adds `RemoteUnitsPathTestCase`, which exercises the `remote_units_path` option.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -35,6 +35,15 @@ from pulp_smash.tests.rpm.api_v2.utils import (
 from pulp_smash.tests.rpm.utils import set_up_module
 
 
+def _split_path(path):
+    """Split a filesystem path into all of its component pieces."""
+    head, tail = os.path.split(path)
+    if head == '':
+        return (tail,)
+    else:
+        return _split_path(head) + (tail,)
+
+
 def setUpModule():  # pylint:disable=invalid-name
     """Conditionally skip the tests in this module.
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This pull request passes when run against a Pulp 2.10.1 system provisioned this morning.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -211,6 +221,37 @@ def verify_publish_is_skip(self, cfg, call_report):
...
+            'content/units',
+        )
+        for segment in _split_path(remote_units_path):
+            cmd = sudo + ('ls', '-1', path)
+            files = set(cli_client.run(cmd).stdout.strip().split('\n'))
+            self.assertIn(segment, files)
+            path = os.path.join(path, segment)
+        cmd = sudo + ('find', path, '-name', '*.rpm')
+        files = cli_client.run(cmd).stdout.strip().split('\n')
+        self.assertEqual(len(files), 32, files)
```
elyezer: We are using this constant on some places and this is directly linked to the fixtures repo. We may create a constant file and use that instead eventually so it will help updating if for some reason RPMs are added or removed from the fixtures repo.

PR #380, opened 2016-08-31
Expand test for pulp-admin's --force-sync option

Ichimonji10:
Expand `pulp_smash.tests.rpm.cli.test_sync.ForceSyncTestCase`.
Currently, the test case will:

1. Create and sync a repo.
2. Remove an RPM. Verify one RPM is missing.
3. Sync the repo with `--force-full true`. Verify all RPMs are present.

This test procedure shows that the second sync restores the removed RPM,
but it doesn't show that the `--force-sync true` option was the cause of
that restoration. It could be that syncs *always* restore removed RPMs.
(This would indicate a different issue with the fast-forward logic.)
Update the test so that it will:

1. Create and sync a repo.
2. Remove an RPM. Verify one RPM is missing.
3. Sync the repo. Verify one RPM is missing.
4. Sync the repo with `--force-full true`. Verify all RPMs are present.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -298,11 +298,11 @@ class PublishBeforeYumDistTestCase(
         self.assertNotIn('content', dirs)
 
 
-class PublishTestCase(
+class ForceFullTestCase(
         _RsyncDistUtilsMixin,
         DisableSELinuxMixin,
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested against a pair of Pulp 2.10.0 and 2.10.1 systems.

```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.cli.test_sync
...
Ran 3 tests in 135.682s

OK
```

PR #381, opened 2016-09-01
Add script for profiling test runs

Ichimonji10:
This new script can be used like so:

    python -m cProfile -o run.prof scripts/run_functional_tests.py

PR #382, opened 2016-09-01
Use orphans to speed up RPM rsync distrib tests

Ichimonji10:
If a sync is performed and matching orphans are on disk, Pulp will
re-use them instead of syncing downloading the content units from a
remote system.

Update the test cases for the RPM rsync distributor. Drop the call to
delete orphans after each test case, and instead delete orphans when the
module has finished executing.


Ichimonji10:
Yes, I know it's a bit ridiculous to submit a pull request whose sole purpose is to make a module of tests run ten seconds faster. What can I say - I did this *before* submitting https://github.com/PulpQE/pulp-smash/pull/381, which added the profiling script.

By the way, some other methods for making the tests run faster, like increasing the polling frequency in the API's HTTP 202 handler function, make an even *smaller* difference. That surprised me. Seems like some of the biggest gains - at least for my system setup - would come from looking at the communication times.

Issue #347, opened 2016-07-31
As a user, units which don't pass the signature check are not imported 


ipanova:
1)create repo with feed and specify list of allowed keys( which are different from the signing keys of synced content).
2) sync repo
3) check that none of those packages were imported into the repo
4) check the counter of not imported packages is correct

1) create repo with feed and specify list of allowed keys( which is the same as the signing keys of synced content).
2) sync repo
3) check that packages were imported into the repo

1)create repo with feed and set require-signature to True. It is better to provide a feed where there are several types of packages like [here](https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/test_drpm_repo/) DRPMs + SRPMs
2) sync repo
3) check that none of those packages were imported into the repo because they are unsigned
4) check the counter of not importer packages is correct ( sum of drpms and srpms)
5) update require-signature to False
6) sync repo again
7) check that all packages were imported into the repo

Since there are 2 config parameters `--require-signature` and `--allowed-keys` you can test different scenarios where you:

1) require packages to be signed and provide a specific list of keys
2) require packages just to be signed but you don't care with which keys( i.e allowed-keys is empty)
3) you allow unsigned packages but those which are signed should be signed with specific keys that are in the list of allowed-keys

Since signature policy is not compatible with lazy sync, confirm that:

1) during repo creation you cannot have on_demand background policy and signature policy enabled
2) during repo update you cannot switch to on_demand background policy if signature policy is enabled. ("Signature policy enabled" means you have some keys specified in the allowed-keys or require-signature is set to True.)

Keys that are provided into allowed-keys should be 8 characters long. Confirm that providing keys that has different length will raise an error.

This functionally was implemented not just for sync but for copy and upload. so you will have here a bit of work..

Some copy scenarios:

1) create repoA with on_demand policy and sync it
2) create repoB with signature policy enabled
3) copy from repoA to repoB and confirm that no packages were imported into repoB

1) create repoA  and sync it( for example sync unsigned packages)
2) create repoB with signature policy enabled( require-signature set to True)
3) copy from repoA to repoB and confirm that no packages were imported into repoB

Upload is the most simple here.

1) create repo with signature policy enabled
2) upload a concrete file  or from a directory
3) confirm package was imported or not based on yours signature policy settings.

Ichimonji10:
Very, very good issue. Thank you, @ipanova.

Ichimonji10:
Related: https://pulp.plan.io/issues/1991

[...]

PR #383, opened 2016-09-02
Test key ID checks when syncing packages

Ichimonji10:
Add a new module,
`pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_syncs`. This
new module verifies whether RPM repository importers configured to
perform package signature ID checks do so when packages are synced in to
repositories.

The new test cases exercise importers that:

* allow unsigned packages and specify a valid key ID.
* allow unsigned packages and specify an invalid key ID.
* allow unsigned packages and specify no key ID.
* require signed packages and specify a valid key ID.
* require signed packages and specify an invalid key ID.
* require signed packages and specify no key ID.

See: https://github.com/PulpQE/pulp-smash/issues/347

See: 7779cea52adeb009b280190fd4050c6abebdb2af

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -72,6 +72,9 @@ This URL can be used as the "feed" property of a Pulp Docker registry.
 DRPM_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'drpm-unsigned/')
 """The URL to a DRPM repository."""
 
+DRPM_UNSIGNED_FEED_COUNT = 4
+"""The number of packages available at :data:`DRPM_UNSIGNED_FEED_URL`."""
+
 DRPM_UNSIGNED_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'drpm-unsigned/')
 """The URL to an unsigned DRPM repository."""
 
@@ -289,7 +292,7 @@ RPM_ERRATUM_URL = (
 """The URL to an JSON erratum file for an RPM repository."""
 
 RPM_FEED_COUNT = 32
-"""The number of RPMs available at :data:`RPM_FEED_URL`."""
+"""The number of packages available at :data:`RPM_FEED_URL`."""
 
 RPM_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'rpm/')
 """The URL to an RPM repository. See :data:`RPM_URL`."""
@@ -298,6 +301,9 @@ RPM_SHA256_CHECKSUM = (
     '4fe8d0e21ee6d56d420c396a02aeaeb59feb00b625811b6a2b4d8f0c1aad80ca'
 )
 """The sha256 checksum of :data:`pulp_smash.constants.RPM`."""
+
+RPM_UNSIGNED_FEED_COUNT = 32
+"""The number of packages available at :data:`RPM_UNSIGNED_FEED_URL`."""
 
 RPM_UNSIGNED_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'rpm-unsigned/')
 """The URL to an unsigned RPM repository. See :data:`RPM_URL`."""
@@ -334,8 +340,14 @@ metadata/rpm
 SRPM = 'test-srpm02-1.0-1.src.rpm'
 """The name of an SRPM file at :data:`pulp_smash.constants.SRPM_FEED_URL`."""
 
+SRPM_FEED_COUNT = 3
+"""The number of packages available at :data:`SRPM_FEED_URL`."""
+
 SRPM_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'srpm/')
 """The URL to an SRPM repository."""
+
+SRPM_UNSIGNED_FEED_COUNT = 3
+"""The number of packages available at :data:`SRPM_UNSIGNED_FEED_COUNT`."""
 
 SRPM_UNSIGNED_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'srpm-unsigned/')
 """The URL to an unsigned SRPM repository."""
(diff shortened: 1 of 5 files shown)


Ichimonji10:
```
(pulp-smash2) [ichimonji10@beech:pulp-smash]$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_signatures_checked_f
or_syncs
.....................
Ran 21 tests in 640.281s

OK
```

pulp_smash/constants.py
```diff
@@ -289,7 +292,7 @@
 """The URL to an JSON erratum file for an RPM repository."""
 
 RPM_FEED_COUNT = 32
-"""The number of RPMs available at :data:`RPM_FEED_URL`."""
+"""The number of packages available at :data:`RPM_FEED_URL`."""
```
elyezer: :+1:

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
```diff
@@ -0,0 +1,454 @@
+# coding=utf-8
+"""Tests for repository signature checks when syncing packages.
+
+This module mimics
+:mod:`pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_uploads`, except
+that packages are synced in to Pulp instead of being uploaded.
+"""
+# NOTE to test authors: It's important to remove all content units after each
```
elyezer: :+1:

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
```diff
@@ -0,0 +1,454 @@
...
+    """
+
+    def test_signed_packages(self):
+        """Import signed RPMs, DRPMs and SRPMs into repositories.
+
+        Assert no packages are synced.
+        """
+        variants = (
+            (RPM_FEED_URL, 'rpm'),
+            (SRPM_FEED_URL, 'srpm'),
```
elyezer: Missing DRPM? The docstring mentions about DRPM

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
```diff
@@ -0,0 +1,454 @@
...
+
+    The importer should have the following pseudocode configuration:
+
+    .. code-block:: json
+
+        {"require_signature": false, "allowed_keys": ["invalid key id"]}
+    """
+
+    def test_signed_packages(self):
+        """Import signed RPMs, DRPMs and SRPMs into repositories.
```
elyezer: If `require_signature` is false, I think the contents should be synced even though an invalid key is provided. I may misunderstood the logic behind `require_signature` but it seems to be the switch when enabled to check if there is keys provided, is that correct?

elyezer:
What about explaining what should be the expected behavior on each TestCase class docstring? For example when using `require_signature` == False and `allowed_keys` == [valid_key] then pulp should only sync signed contents using the valid key and unsigned content, any content signed with a different key will be skipped.

I think that will make clear what should be the outcome of all the test methods of each class.

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
Ichimonji10: Ah. I need to fix the docstring. Yes, DRPMs should be a part of this test, and several others besides. However, we don't currently have signed DRPMs in pulp-fixtures. See: https://github.com/PulpQE/pulp-fixtures/issues/25

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
Ichimonji10: That's what I thought too, but the signature checking logic doesn't work that way. You're correct that if `require_signature` is false, then signatures aren't required. But there's some subtleties here.

If an importer config has `{'require_signature': False, 'allowed_keys': []}`, then any package with may be imported.

If an importer has `{'require_signature': False, 'allowed_keys': ['abcd1234']}`, then:

* Packages with no signatures are allowed.
* Packages with signature `abcd1234` are allowed.
* Packages with any other signature are disallowed.

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
Ichimonji10: I've dropped the reference to DRPMs, and added a reference to https://github.com/PulpQE/pulp-fixtures/issues/25

Ichimonji10:
I've made some updates to the docstrings in the module being added, `test_signatures_checked_for_syncs`, and in an existing signature-checking module, `test_signatues_checked_for_uploads`. Here's what I've done:

* Tests that should intuitively deal with signed DRPMs now more clearly indicate that they don't actually deal with signed DRPMs, and they reference https://github.com/PulpQE/pulp-fixtures/issues/25.
* Both the module-wide docstring and the test case with the `{'require_signature': False, 'allowed_keys': [valid key]}` importer state that Pulp's signature checking logic is subtle, and they suggest that the reader review the docstring in `test_signatures_checked_for_uploads`.
* The module-wide docstring in `test_signatures_checked_for_uploads` makes a special point of explaining what happens when an importer is set to `{'require_signature': False, 'allowed_keys': [valid key]}`.

pulp_smash/tests/rpm/api_v2/test_signatures_checked_for_syncs.py
```diff
@@ -0,0 +1,466 @@
...
+class AllowValidKeyTestCase(_BaseTestCase):
+    """Use an importer that allows unsigned packages and has a valid key ID.
+
+    The importer should have the following pseudocode configuration:
+
+    .. code-block:: json
+
+        {"require_signature": false, "allowed_keys": ["valid key id"]}
+
+    .. NOTE:: Pulp's signature checking logic is subtle. Please read
```
elyezer: :+1:

elyezer:
> I've not added explanations of how Pulp behaves into every test method because, IMO, the proper place for those in-depth explanations is in Pulp's documentation. I'm happy to point out un-intuitive cases to the reader, but want to avoid too much duplication. Thoughts?

Agreed. We should avoid duplication and pointing to the documentation is a good thing.

PR #387, opened 2016-09-08
Skip tests affected by Pulp issue 2242

Ichimonji10:
According to Pulp issue 2242, package signature ID checking is not done
when packages are synced into a repository. Skip the relevant tests.

See: https://pulp.plan.io/issues/2242


Ichimonji10:
This should fix several Jenkins failures.

PR #390, opened 2016-09-13
OSTree tests

Ichimonji10:
The first commit is a simple refactor.

The test added by the second commit is extremely simple. Definitely do check out the linked-to Pulp issues for a sense of why the test is so simple.


Ichimonji10:
Tested against Pulp 2.10 and 2.11:

```
$ python -m unittest2 pulp_smash.tests.ostree.api_v2.test_{sync,publish}
..............
Ran 14 tests in 52.513s

OK
```

PR #395, opened 2016-09-16
Drop RPM_SHA256_CHECKSUM constant

Ichimonji10:
As a replacement, add a get_sha256_checksum() function. The advantage of
calculating a checksum at runtime is that some details of the fixtures
used by the tests can change between test runs. For example, an RPM
fixture might gain or lose a PGP signature.

See: https://github.com/PulpQE/pulp-fixtures/pull/28

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -289,11 +289,6 @@ RPM_FEED_COUNT = 32
 RPM_FEED_URL = urljoin(PULP_FIXTURES_BASE_URL, 'rpm/')
 """The URL to an RPM repository. See :data:`RPM_URL`."""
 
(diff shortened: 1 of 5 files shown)


Ichimonji10:
```console
$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_download_policies
.................
Ran 17 tests in 325.297s

OK
$ python -m unittest2 pulp_smash.tests.rpm.api_v2.test_export
s......
Ran 6 tests in 304.338s

OK (skipped=1)
```

Ichimonji10:
Rebased, and sanity-checked with `python -m unittest pulp_smash.tests.rpm.api_v2.test_export`.

PR #399, opened 2016-09-20
Drop constant RPM_ABS_PATH

Ichimonji10:
The `RPM_ABS_PATH` constant defines the absolute path to the `RPM`
content unit on the Pulp filesystem.

For this constant to hold true, the `RPM` content unit must not change.
However, it has changed at least twice during the history of Pulp
Fixtures, as signatures have been added and removed. Also, its signature
might change in the future if the public/private keypair in Pulp
Fixtures changes for some reason, and the exact path to the content unit
is supposed to be an implementation detail.

Rewrite the one test case that uses this constant to use the `find` CLI
executable instead.

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -287,12 +287,6 @@ The erratum ID is defined by :data:`pulp_smash.constants.RPM_ERRATUM_ID`.
 RPM = 'bear-4.1-1.noarch.rpm'
 """The name of an RPM file. See :data:`pulp_smash.constants.RPM_URL`."""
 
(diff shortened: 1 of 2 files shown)


Ichimonji10:
This test fails. However, it doesn't fail because of an error introduced by this change. Rather, this test fails because of a bug in Pulp itself: Pulp ignores orders to re-download repositories. I'm going to file a separate bug and PR for that issue. In the meantime, you can verify that this change succeeds by checking out this PR *and* changing the download type from ``on_demand`` to ``immediate`` or something else. You'll see that the content unit is correctly found and used.

PR #400, opened 2016-09-20
Strengthen tests in FixFileCorruptionTestCase

Ichimonji10:
Among other things, `FixFileCorruptionTestCase` does the following:

1. Corrupt a content unit on disk.
2. Tell Pulp to download a repository with ``verify_all_units`` set to
   false.
3. Tell Pulp to download a repository with ``verify_all_units`` set to
   true.

One can make at least the following assertions about this:

1. The content unit's checksum changed as a result of step 1.
2. The content unit's checksum stayed the same as a result of step 2.
3. The content unit's checksum changed as a result of step 3.
4. The content unit's checksum is the same before step 1 and after step
   3.

Add the third assertion. Revamp the test names and docstrings for the
other three steps.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_download_policies.py
+++ b/pulp_smash/tests/rpm/api_v2/test_download_policies.py
@@ -302,17 +302,35 @@ class FixFileCorruptionTestCase(utils.BaseAPITestCase):
         )
 
     def test_corruption_occurred(self):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
As with https://github.com/PulpQE/pulp-smash/pull/399, the test affected by this PR fails, but not because of this PR. See: https://github.com/PulpQE/pulp-smash/pull/399#issuecomment-248409141

Issue #402, opened 2016-09-21
Test upload for srpm, comps.xml


preethit:
Taking a quick look at the existing tests and issues, seems like upload tests for srpm and comps.xml are missing.

  comps       - uploads comps.xml into a repository
  srpm        - uploads one or more SRPMs into a repository

Test case1
1. Create a repository A
2. Upload an srpm to the repository
3. Publish repository

Assert
1. The srpm can be uploaded
2. The repository with the srpm content can be published

Test case 2
1. Create a repository A
2. Upload a comps.xml to the repository
3. Publish repository

Assert
1. The comps.xml can be uploaded
2. The repository with the comps.xml content can be published

Ichimonji10:
We upload SRPMs in `pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_uploads`. That said, we can expand `pulp_smash.tests.rpm.api_v2.test_upload_publish` to better target this issue.

preethit:
I am looking at the available options for uploads.

 pulp-admin rpm repo uploads
Usage: pulp-admin [SUB_SECTION, ..] COMMAND
Description: upload modules into a repository

Available Commands:
  cancel      - cancels an outstanding upload request
  category    - creates a new package category
  comps       - uploads comps.xml into a repository
  environment - creates a new package environment
  erratum     - creates a new erratum
  group       - creates a new package group
  langpacks   - creates a new package langpacks
  list        - lists in-progress and paused uploads
  resume      - resume a paused upload request
  rpm         - uploads one or more RPMs into a repository
  srpm        - uploads one or more SRPMs into a repository

Also for the other contents

I see these
pulp_smash.tests.rpm.api_v2.test_upload_publish
with rpm & errata
pulp_smash.tests.rpm.cli.test_upload
with drpm

Ichimonji10:
Gotcha. Thanks!

PR #404, opened 2016-09-22
Add UploadSrpmTestCase

Ichimonji10:
This test case verifies that it's possible to upload an SRPM to a
repository.

See: https://github.com/PulpQE/pulp-smash/issues/402

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
@@ -24,6 +24,8 @@ from pulp_smash.constants import (
     RPM,
     RPM_FEED_URL,
     RPM_URL,
+    SRPM,
+    SRPM_UNSIGNED_URL,
 )
 from pulp_smash.tests.rpm.utils import gen_erratum
 from pulp_smash.tests.rpm.api_v2.utils import (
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This test case is basically a copy-paste job of `UploadDrpmTestCase`. I'm not happy about this. But the existing test case also has some definite issues, and the proper solution requires more work than I want to put in right now.

Ichimonji10:
@preethit FYI

Ichimonji10:
Tested against Pulp 2.10 and 2.11:

```console
$ python -m unittest pulp_smash.tests.rpm.api_v2.test_upload_publish.UploadSrpmTestCase
...
Ran 3 tests in 9.746s

OK
```

preethit:
@Ichimonji10 Awesome!

omaciel:
@Ichimonji10 I don't know what value should be used to import a debug rpm, but if you know the answer to this question, then you could basically have the same test for that content type too :)

omaciel:
@Ichimonji10 https://fedorahosted.org/spacewalk/wiki/DeltaRpmSupport

Ichimonji10:
> @Ichimonji10 I don't know what value should be used to import a debug rpm, but if you know the answer to this question, then you could basically have the same test for that content type too :)

@omaciel A quick look through http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_upload_publish.html shows that we have test cases that upload RPM, DRPM and errata files.

This new test case is basically a copy-paste of `UploadDrpmTestCase`. I'm not happy about the copy-pasta. But there's some other issues with `Upload{Drpm,Srpm}TestCase`, too, and I'd rather get this coverage in now and then both de-duplicate and refactor all together at a later point.

Ichimonji10:
It's also worth noting that the following modules do extensive work with RPMs, SRPMs and DRPMs:

* http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_signatures_saved_for_packages.html
* http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_copies.html
* http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_syncs.html
* http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.rpm.api_v2.test_signatures_checked_for_uploads.html

omaciel:
@Ichimonji10 thanks for the links. I actually wrote my comment while in a meeting, but I then searched for `drpm` and found the tests you mentioned, so... awesome! :)

elyezer:
:+1:

PR #408, opened 2016-09-27
Skip tests affected by Pulp issue 2277

Ichimonji10:
See: https://pulp.plan.io/issues/2277


Ichimonji10:
For each failing test in Jenkins, I did the following:

1. Verify that I can reproduce the issue on my Pulp 2.10 and 2.11 systems.
2. Verify that the issue disappears when SELinux is disabled.
3. Read the code and verify that it can reasonably be linked to issue 2277.

For each test that satisfied these conditions, I added in skipping logic. I could also have added in SELinux enable/disable logic, but that would have been a *lot* more work, all for a bug that should hopefully be fixed very soon.

PR #410, opened 2016-09-30
Pin dependency on pydocstyle==1.0.0

Ichimonji10:
pydocstyle 1.1.0 is not compatible with flake8.

See: https://github.com/PyCQA/pydocstyle/issues/207


Ichimonji10:
Thanks for the review.

Issue #384, opened 2016-09-02
Repos can be created and synced successfully with mirrorlist feed


preethit:
Test that repos can be created and synced with mirrorlist --feed
https://pulp.plan.io/issues/2224
https://pulp.plan.io/issues/175

1. pulp-admin -vv rpm repo create --repo-id mirror --feed "http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os"  --relative-url mirror

2. pulp-admin rpm repo sync run --repo-id=mirror

Assert that

1. Repo create is successful with the --relative-url 
2. Repo create is successful without --relative-url
3. Repo can be synced with relative-url
4. Repo can be synced without relative-url
5. Repo can be published with relative-url
7. Repo can be synced with lazy settings (on_demand,background)

Ichimonji10:
The given set of commands in this test case will sync a significant number of packages. For example, the command above will resolve to a page like http://mirror.fusioncloud.co/pub/centos/7.2.1511/os/x86_64/Packages/, which lists over 9,000 packages. It's unacceptable for an automated test to sync that many packages, especially from systems not hosted by us.

I'm not aware of any good technique for choosing which packages should be synced from a repository. The [yum importer documentation](http://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/yum-plugins.html#yum-importer) does list a `skip` option, but:

* `skip` is a blacklisting tool. Should the set of content types change or be expanded, we'd unknowingly start syncing additional content. A whitelisting tool is needed.
* `skip` is a coarse tool. It lets one select certain types of packages, like RPMs and SRPMs, but not individual packages.

I don't think this bug is suitable for automated testing, given Pulp's current capabilities.

omaciel:
@Ichimonji10 @preethit is it possible to further filter the provided `feed`? If not, can we generate our own fixture that would allow us to automate this? I don't think that this issue should be closed but further developed so that the use case can be automated imho.

mhrivnak:
Like @omaciel I also had the thought that we should have a fixture that provides a simple mirror list of a small repo.

Ichimonji10:
> can we generate our own fixture that would allow us to automate this?

The mirrorlist files hosted by centos contain full URLs including hostnames. If we want to do something similar, Pulp Fixtures can't generate that mirrorlist, as it's unaware of where files are being hosted.

[...]

PR #412, opened 2016-10-05
Add test cases for mirrorlists

Ichimonji10:
Add module `pulp_smash.tests.rpm.api_v2.test_mirrorlist`. This module
contains test cases that exercise Pulp's support for mirrorlist feed
URLs.

Fix: https://github.com/PulpQE/pulp-smash/issues/384


Ichimonji10:
Tested against Pulp 2.10.1 and 2.11 (master). Most of the tests are skipped due to various bugs. If the bug references are removed and SELinux is disabled, all tests pass.

PR #413, opened 2016-10-10
Add test to verify no change on a second sync

elyezer:
The test will sync a second time a repository that hasn't changed. Pulp must
report properly that no changes were made.


Ichimonji10:
Verified on Pulp 2.8-2.11.

PR #414, opened 2016-10-10
Update Docker CLI tests

Ichimonji10:
pulp-admin's output has changed when working with Docker tags has
changed. Update tests to properly screen-scrape its current output.

It doesn't seem appropriate to file a bug asking for the changes to be
reverted. Pulp Admin's output is intended to be human-friendly, not
machine parse-able. (Status codes should have been used for this
purpose.) And a close look at pulp-admin's output when copying images,
manifests or tags between repositories shows that there's already subtle
differences between the types of output. The same is true when searching
for images, manifests or tags in existing repositories.


Ichimonji10:
This fixes an intermittent failure across Pulp 2.8, 2.9, 2.10 and 2.11. I tested this change against all four versions of Pulp.

PR #415, opened 2016-10-11
Set repository type when creating Python repos

Ichimonji10:
Pulp repositories don't have a "type" per se. However, practical
experience shows that repositories should include only a closely related
set of content types. This is reflected in repository metadata. If a
repository's type is listed in its metadata, then front-ends (like
pulp-admin) can better display repository information to users.


Ichimonji10:
Tested with `python -m unittest discover pulp_smash.tests.python.api_v2` against Pulp 2.8-2.11. No test results changed.

Issue #416, opened 2016-10-12
Rename and skip python sync test for pulp_python version < 2.0


asmacdo:
Test to be renamed and skipped: 
https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/python/api_v2/test_sync_publish.py#L12

It should be renamed to `PulpToPulpSyncTestCase` because it syncs from a fixture that was generated by a publish from pulp_python 2.0. Pulp to pulp sync is a new feature in 2.0.

Ichimonji10:
What's the first version of Pulp that pulp_python 2.0 ships with?

asmacdo:
Probably 2.11, but we haven't officially made a decision yet.

Ichimonji10:
Thanks!

PR #418, opened 2016-10-12
Split Python repo sync test case in two

Ichimonji10:
Pulp is currently able to sync Python packages from PyPI. As of the
pulp_python plugin version 2.0, it should also be able to sync Python
packages from other Pulp systems. Update module
`pulp_smash.tests.python.api_v2.test_sync_publish` to test both of these
scenarios.

Fix: https://github.com/PulpQE/pulp-smash/issues/416


Ichimonji10:
Successfully tested against Pulp 2.8-2.11.

Issue #425, opened 2016-10-20
Automate issue https://pulp.plan.io/issues/1088


omaciel:
Please design an automated test case for the issue highlighted by https://pulp.plan.io/issues/1088

PR #428, opened 2016-10-26
Add a test for Pulp issue 1088

Ichimonji10:
See: https://pulp.plan.io/issues/1088

Fix: https://github.com/PulpQE/pulp-smash/issues/425

Thanks to @jortel for help with designing the test.


Ichimonji10:
Tested against Pulp 2.8-2.11. The issue described in https://pulp.plan.io/issues/1088 is present in all four versions of Pulp. The test passes against all four versions of Pulp, except for the last pair of assertions.

This test goes a bit above and beyond what's strictly necessary. In addition to verifying that incremental fast-forward publishes behave correctly, it also verifies that full publishes behave correctly.

PR #430, opened 2016-11-02
Added ssh_port option to config

midnightercz:
Allows user to set more flexible configuration, which can be
great for development instances of pulp

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -165,7 +165,8 @@ class Client(object):  # pylint:disable=too-few-public-methods
             # The SshMachine is a wrapper around the system's "ssh" binary.
             # Thus, it uses ~/.ssh/config, ~/.ssh/known_hosts, etc.
             self.machine = (  # pylint:disable=redefined-variable-type
-                plumbum.machines.SshMachine(hostname)
+                plumbum.machines.SshMachine(hostname,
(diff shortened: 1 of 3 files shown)


Ichimonji10:
The desired SSH port can be set in `~/.ssh/config`. See `ssh_config(5)`. I'd rather not set any OpenSSH options in Pulp:

* There's a large number of available OpenSSH options, and allowing them to be set in the Pulp Smash config opens the door to a great increase in the set of available options.
* Pulp Smash may include SSH customization options in the future. If so, however, I'd like for them to be able to deal with generic cases where, for example, dropbear SSH is used as the SSH  transp[ort.

Ichimonji10:
As an example of what `~/.ssh/config` can look like:

```
# libvirt hosts
Host 192.168.100.*
    User root
    IdentityFile ~/.ssh/libvirt
    IdentityFile ~/.ssh/libvirt-rhel6
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null

    ControlMaster auto
    ControlPersist 5m
    ControlPath ~/.ssh/controlmasters/%C
```

midnightercz:
From my point of view, it's better to have all configuration in one place rather then having multiple sources of configuration that can override each other. But I respect your standpoint

Issue #427, opened 2016-10-25
Test the process recycling feature


bmbouter:
I posted a quick test plan outline on the issue here:

https://pulp.plan.io/issues/2172#note-21

PR #432, opened 2016-11-02
Test Pulp's ability to recycle processes

Ichimonji10:
Test whether Pulp's `PULP_MAX_TASKS_PER_CHILD` configuration option
works as advertised.

See: https://pulp.plan.io/issues/2172#note-21

Fix: https://github.com/PulpQE/pulp-smash/issues/427

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -99,6 +99,7 @@ developers, not a gospel.
     api/pulp_smash.tests.rpm.cli.test_copy_units
     api/pulp_smash.tests.rpm.cli.test_environments
     api/pulp_smash.tests.rpm.cli.test_langpacks
+    api/pulp_smash.tests.rpm.cli.test_process_recycling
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Tested against Pulp 2.11 running on Fedora 24 and RHEL 6.

Ichimonji10:
Thanks! Credit must go to @bmbouter for the test plan outlined in https://pulp.plan.io/issues/2172#note-21. It made my life *much* easier.

By the way, I just noticed that while this test creates an RPM repository, it doesn't sync that repository. Syncing would give a better sanity check. I'll update this pull request tomorrow.

Ichimonji10:
I've added the `pulp-admin rpm repo sync run` lines of code. The test itself is valid. I did hit an assertion error, but if anything, that points to an issue with Pulp, not this test.

PR #433, opened 2016-11-04
Skip tests affected by Pulp issue 2387

Ichimonji10:
Also, add function `pulp_smash.tests.rpm.utils.os_is_rhel6`. Remove a
corresponding private function.

See: https://pulp.plan.io/issues/2387


Ichimonji10:
Tested against Pulp 2.8-2.11 on RHEL 6.

PR #434, opened 2016-11-11
Expand and clean up skips for Pulp issue 2387

Ichimonji10:
Add a new function, `check_issue_2387`, that checks whether the system
under test is affected by Pulp issue 2387. Use this function wherever
issue 2387 is referenced.

Add a check for issue 2387 into module
`pulp_smash.tests.rpm.api_v2.test_broker`. The tests in this module are
affected but not yet skipped.

See: https://pulp.plan.io/issues/2387


Ichimonji10:
This should fix several Jenkins issues.

[Pulp #2387](https://pulp.plan.io/issues/2387) has been set to MODIFIED just today, so this whole PR may be a moot point. Still. Tested by making `check_issue_2387` always return true and then running the affected modules.

PR #439, opened 2016-11-16
Enable pulp_manage_rsync boolean for rsync tests

Ichimonji10:
See:

* https://pulp.plan.io/issues/2199
* http://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/iso-rsync-distributor.html#configuration

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -92,13 +92,17 @@ def setUpModule():  # pylint:disable=invalid-name
     .. _Pulp #1759: https://pulp.plan.io/issues/1759
     """
     set_up_module()
-    if selectors.bug_is_untestable(1759, config.get_config().version):
+    cfg = config.get_config()
+    if selectors.bug_is_untestable(1759, cfg.version):
         raise unittest.SkipTest('https://pulp.plan.io/issues/1759')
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This should fix several test failures.

Ichimonji10:
Tested against Pulp 2.10.2 on RHEL 6, RHEL 7 and Fedora 24, and Pulp 2.11.1 on Fedora 24.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -92,13 +92,23 @@ def setUpModule():  # pylint:disable=invalid-name
...
+        cmd.insert(0, 'sudo')
+    cli.Client(cfg).run(cmd)
 
 
 def tearDownModule():  # pylint:disable=invalid-name
     """Delete orphan content units."""
-    api.Client(config.get_config()).delete(ORPHANS_PATH)
+    cfg = config.get_config()
+    api.Client(cfg).delete(ORPHANS_PATH)
+    cmd = 'semanage boolean --modify --off pulp_manage_rsync'.split()
```
elyezer: Would not worth getting this logic as a helper function? The helper function can receive a boolean indicating to turn on and off and the config object.

This is not a blocker.

Ichimonji10:
Good suggestion. Updated.

PR #445, opened 2016-12-01
Test Pulp's handling of packages in alternate locs

Ichimonji10:
An RPM repository may have packages in arbitrary locations. By default,
Pulp stores packages in the root of a repository, but this need not be
so. The metadata in the ``repodata`` directory can point to RPM files
wherever they be. Add a test for this feature.

See: https://pulp.plan.io/issues/2354

See: https://github.com/PulpQE/pulp-fixtures/pull/45


Ichimonji10:
Tested against a pair of RHEL 7.3 systems running Pulp 2.10.3 and 2.11.1, respectively.

jortel:
LGTM

Issue #444, opened 2016-11-30
Add test to check that all processes are running in the correct SELinux context


dkliban:
Pulp processes need to transition to proper SELinux contexts in order for them to function correctly. The context can be determined by passing -Z to the ps command.  The first column of the output is the SELinux context. 

For `httpd`, it should be `unconfined_u:system_r:httpd_t:s0`, as can be seen below:

```
# ps awfuxZ | grep wsgi
unconfined_u:system_r:streamer_t:s0 apache 14074 0.0  2.7 586200 52392 ?       S    Nov29   0:16 /usr/bin/python /usr/bin/pulp_streamer --pidfile=/var/run/pulp/pulp_streamer.pid --syslog --prefix=pulp_streamer --python /usr/share/pulp/wsgi/streamer.tac
unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 root 25180 0.0  0.0 103320 828 pts/3 S+ 15:14   0:00  |   \_ grep wsgi
unconfined_u:system_r:httpd_t:s0 apache  19563  0.0  3.2 1004888 62500 ?       Sl   Nov29   1:05  \_ (wsgi:pulp)    
unconfined_u:system_r:httpd_t:s0 apache  19565  0.0  0.9 514596 18172 ?        Sl   Nov29   0:02  \_ (wsgi:pulp-cont
unconfined_u:system_r:httpd_t:s0 apache  19566  0.0  2.9 841864 56812 ?        Sl   Nov29   0:17  \_ (wsgi:pulp_forg
unconfined_u:system_r:httpd_t:s0 apache  23558  0.0  0.9 514596 18196 ?        Sl   09:02   0:00  \_ (wsgi:pulp-cont
unconfined_u:system_r:httpd_t:s0 apache  23559  0.0  0.9 514596 18196 ?        Sl   09:02   0:00  \_ (wsgi:pulp-cont
unconfined_u:system_r:httpd_t:s0 apache  23560  0.1  3.2 1004888 62504 ?       Sl   09:02   0:24  \_ (wsgi:pulp)    
unconfined_u:system_r:httpd_t:s0 apache  23561  0.0  2.9 841860 56812 ?        Sl   09:02   0:07  \_ (wsgi:pulp_forg
unconfined_u:system_r:httpd_t:s0 apache  23642  0.1  3.2 1004892 62508 ?       Sl   09:02   0:25  \_ (wsgi:pulp)    
[...]

PR #449, opened 2016-12-06
Test Pulp process labels

Ichimonji10:
Add module `pulp_smash.tests.platform.cli.test_selinux`.

Fix: https://github.com/PulpQE/pulp-smash/issues/444

### DIFF
--- a/pulp_smash/tests/platform/cli/test_selinux.py
+++ b/pulp_smash/tests/platform/cli/test_selinux.py
@@ -1,0 +1,94 @@
+# coding=utf-8
+"""Tests to verify that Pulp has proper SELinux permissions."""
+import unittest
+from collections import namedtuple
+
+from pulp_smash import cli, config, utils
+
+
+CELERY_LABEL = ':system_r:celery_t:s0'
+"""A label commonly applied to celery processes.
+
+The "user" segment of the label is intentionally omitted, as it is known to
+vary. See `Pulp Smash #444 (comment)
+<https://github.com/PulpQE/pulp-smash/issues/444#issuecomment-265798957>`_.
+"""
+
+HTTPD_LABEL = ':system_r:httpd_t:s0'
+"""A label commonly applied to httpd processes.
+
+The "user" segment of the label is intentionally omitted, as it is known to
+vary. See `Pulp Smash #444 (comment)
+<https://github.com/PulpQE/pulp-smash/issues/444#issuecomment-265798957>`_.
+"""
+
+PS_FIELDS = ('label', 'args')
+"""The fields that ``ps`` should display. See ``man ps`` for details."""
+
+Process = namedtuple('Process', PS_FIELDS)
+"""A single line of output from ``ps``."""
+
+
+class ProcessLabelsTestCase(unittest.TestCase):
+    """Test that Pulp processes have correct SELinux labels.
+
+    This test case targets `Pulp Smash #444
+    <https://github.com/PulpQE/pulp-smash/issues/444>`_.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Get all of the processes running on the target Pulp system."""
+        cfg = config.get_config()
+        cmd = ['ps', '-A', '-w', '-w', '-o', ','.join(PS_FIELDS)]
+        if not utils.is_root(cfg):
+            cmd.insert(0, 'sudo')
+        cls.procs = [
+            Process(*line.split(maxsplit=1))
+            for line in cli.Client(cfg).run(cmd).stdout.splitlines()
+        ]
+
+    def _do_test(self, label, arg):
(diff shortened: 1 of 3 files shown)


Ichimonji10:
@dkliban, every test in this PR fails, and all the failures are alike. Here's a sample failure:

```
FAIL: test_celerybeat (pulp_smash.tests.platform.cli.test_selinux.ProcessLabelsTestCase)
Verify the label of the Pulp celerybeat process.
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/platform/cli/test_selinux.py", line 79, in test_celerybeat
    self._do_test('unconfined_u:system_r:celery_t:s0', 'celerybeat.pid')
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/platform/cli/test_selinux.py", line 45, in _do_test
    self.assertEqual(proc.label, label, procs)
AssertionError: 'system_u:system_r:celery_t:s0' != 'unconfined_u:system_r:celery_t:s0'
- system_u:system_r:celery_t:s0
? ^^^^ ^
+ unconfined_u:system_r:celery_t:s0
? ^^^^^^^^ ^
 : [Process(label='system_u:system_r:celery_t:s0', args='/usr/bin/python /usr/bin/celery beat --app=pulp.server.async.celery_instance.celery --scheduler=pulp.server.async.scheduler.Scheduler --workdir=/var/run/pulp/ -f /var/log/pulp/celery
beat.log -l INFO --detach --pidfile=/var/run/pulp/celerybeat.pid')]
```

In every case, the issue is that Pulp processes have a context starting with `unconfined_u`, not `system_u`. Can you give any insight into this?

pulp_smash/tests/platform/cli/test_selinux.py
```diff
@@ -0,0 +1,79 @@
+# coding=utf-8
+"""Tests to verify that Pulp and SELinux are interacting properly."""
```
elyezer: Maybe `Tests to verify that Pulp has the proper SELinux permissions`?

pulp_smash/tests/platform/cli/test_selinux.py
```diff
@@ -0,0 +1,79 @@
...
+import unittest
+from collections import namedtuple
+
+from pulp_smash import cli, config, utils
+
+
+PS_FIELDS = ('label', 'args')
+
+Process = namedtuple('Process', PS_FIELDS)
```
elyezer: namedtuple :heart:

pulp_smash/tests/platform/cli/test_selinux.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _do_test(self, label, arg):
+        """Assert that certain processes have a label of ``label``.
+
+        Select all processes that were invoked with an argument string
+        containing ``arg``. Assert that at least one such process exists, and
+        that all such processes have a label of ``label``.
+        """
+        procs = [proc for proc in self.procs if arg in proc.args]
+        self.assertGreater(len(procs), 0, arg)
```
elyezer: Thanks for checking that at least one process is running.

pulp_smash/tests/platform/cli/test_selinux.py
```diff
@@ -0,0 +1,79 @@
...
+        They should have a label of ``unconfined_u:system_r:httpd_t:s0``.
+        """
+        self._do_test('unconfined_u:system_r:httpd_t:s0', 'wsgi:pulp')
+
+    def test_resource_manager(self):
+        """Verify the labels of the ``resource_manager`` processes.
+
+        They should have a label of ``unconfined_u:system_r:celery_t:s0``.
+        """
+        self._do_test('unconfined_u:system_r:celery_t:s0', 'resource_manager')
```
elyezer: `unconfined_u:system_r:celery_t:s0` is being used on all tests, what about having a constant for this?

Ichimonji10: It's actually used in three of the four tests. :wink: But I'll add a constant or two.

elyezer: Ahh missed the one that does not use it. Anyway adding a constant will help maintenance, but it is not a big deal, we can decide to refactor later.

elyezer:
LGTM, I have not added the label because there's the Needs Work assigned and I am not sure if something else is missing.

Ichimonji10:
The issue is that the tests currently fail, due to *slightly* wrong SELinux labels. It's easy to update the test to make them pass, but I'd like to get some feedback from @dkliban on the right way to do that.

Ichimonji10:
@dkliban Thanks for the clarification!

@elyezer The tests now pass. Only minor modifications were needed. :smile:

PR #450, opened 2016-12-07
Select yum or dnf at runtime

Ichimonji10:
The first commit deduplicates the logic behind the `pulp_smash.utils.is_root` function. It's a small refactor.

The second commit adds `pulp_smash.cli.PackageManager` and replaces hard-coded references to yum.


Ichimonji10:
Tested against a matrix of four systems, running Fedora and RHEL 7.3, with Pulp 2.10 and 2.11 installed.

elyezer:
We can work later on increasing the coverage.

Issue #442, opened 2016-11-28
Test SELinux filesystem labels


elyezer:
Write tests to ensure SELinux labels are applied correctly. At [1] there are .fc files where the expected files/directories SELinux labels are expected.

[1] https://github.com/pulp/pulp/tree/master/server/selinux/server

PR #455, opened 2016-12-13
Test SELinux filesystem labels

Ichimonji10:
Add test case `FileLabelsTestCase` in module
`pulp_smash.tests.platform.cli.test_selinux`.

Fix: https://github.com/PulpQE/pulp-smash/issues/442

### DIFF
--- a/pulp_smash/tests/platform/cli/test_selinux.py
+++ b/pulp_smash/tests/platform/cli/test_selinux.py
@@ -1,9 +1,10 @@
 # coding=utf-8
 """Tests to verify that Pulp has proper SELinux permissions."""
+import re
 import unittest
 from collections import namedtuple
 
-from pulp_smash import cli, config, utils
+from pulp_smash import cli, config, selectors, utils
 
 
 CELERY_LABEL = ':system_r:celery_t:s0'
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@dkliban: This test fails like so:

```console
$ python -m unittest pulp_smash.tests.platform.cli.test_selinux.FileLabelsTestCase
..
FAIL: test_pulp_server_fc (pulp_smash.tests.platform.cli.test_selinux.FileLabelsTestCase) [('/var/lib/pulp', ':object_r:httpd_sys_rw_content_t:s0')]
Test files listed in ``pulp-server.fc``.
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/platform/cli/test_selinux.py", line 175, in test_pulp_server_fc
    self._do_test(file_, label, True)
  File "/home/ichimonji10/code/pulp-smash/pulp_smash/tests/platform/cli/test_selinux.py", line 147, in _do_test
    self.assertEqual(file_label, label, getfattr_file)
AssertionError: ':object_r:pulp_cert_t:s0' != ':object_r:httpd_sys_rw_content_t:s0'
- :object_r:pulp_cert_t:s0
+ :object_r:httpd_sys_rw_content_t:s0
 : ('var/lib/pulp/static/rsa_pub.key',)

Ran 3 tests in 1.532s

FAILED (failures=1)
```

Can you provide any input?

Ichimonji10:
@elyezer This test makes use of the `getfattr` executable. This executable is available out of the box on Fedora, but needs to be installed on RHEL. Can a step to install the `attr` package be added to pulp_packaging or some other appropriate script?

elyezer:
> @elyezer This test makes use of the getfattr executable. This executable is available out of the box on Fedora, but needs to be installed on RHEL. Can a step to install the attr package be added to pulp_packaging or some other appropriate script?

Definitely, I will handle that for you.

Ichimonji10:
By the way, I checked, and the file is called `attr` on Fedora 23, Fedora 24, RHEL 6.8 and RHEL 7.3. No platform-specific logic should be necessary.

Ichimonji10:
Rebased. I also made the minor change of moving a comment.

dkliban:
@Ichimonji10 The pulp_cert_t type makes sense. Without it, SELinux would not allow Pulp to use this file as a cert. However, we should probably have an explicit statement in our SELinux policy that says this file needs to have this security context type. I suspect running restorecon would change the security context type to httpd_sys_rw_content_t.

Ichimonji10:
@dkliban Running `restorecon /var/lib/pulp/static/rsa_pub.key` sets the file's context to `unconfined_u:object_r:pulp_cert_t:s0`. This context is expected by Pulp Smash and is listed here: https://github.com/pulp/pulp/tree/master/server/selinux/server

Ichimonji10:
Test now passes. See: https://pulp.plan.io/issues/2508

Issue #459, opened 2017-01-05
Test orphan remove reports the number of removed unit types


preethit:
Add tests to verify that orphan remove reports the number of units removed per unit type.
https://pulp.plan.io/issues/1268

For example

1. Create a repo with 3 rpms
2. Sync and publish the repo
3. Delete the repo
4. Assert that there are 3 rpms in the orphaned state
5. Run orphan remove
6. Assert that the report says that 3 rpm units were removed

Ichimonji10:
Note to test implementers: I think we have a test that does steps 1 through 5 already.

Ichimonji10:
Test ready at: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:orphan-counts?expand=1

PR #462, opened 2017-01-06
Make test case more resilient against errors

Ichimonji10:
Add a new method, `pulp_smash.selectors.skip_if`. This method is similar
to `@unittest.skipIf`, but more flexible: it lets one execute skipping
logic at runtime and is capable of accessing class and instance
attributes.

Why make this method available? It exists because of the structure of
the unittest module:

1. The `setUpClass` method lacks access to certain tools like the
   `addCleanup` method. As a result, it is desirable to move
   failure-prone code to other locations, such as test methods.
2. Some failure-prone code should control whether or not the test
   methods in a test case should execute, but this is difficult to do,
   because all test methods execute regardless of how many test methods
   execute.
3. The `skip_if` decorator lets one move failure-prone code into a
   regular test method *and* control whether subsequent test methods
   execute.

This decorator may make it easier to avoid
`pulp_smash.utils.BaseAPITestCase`.

### DIFF
--- a/pulp_smash/selectors.py
+++ b/pulp_smash/selectors.py
@@ -217,3 +217,45 @@ def require(version_string):
             return test_method(self, *args, **kwargs)
         return new_test_method
     return plain_decorator
+
+
+def skip_if(func, var_name, result):
+    """A decorator for optionally skipping a unittest test method.
+
+    This decorator checks to see if ``func(getattr(self, var_name))`` equals
+    ``result``. If so, a ``unittest.SkipTest`` exception is raised. Otherwise,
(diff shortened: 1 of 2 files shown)


Ichimonji10:
I wrote this up while tackling https://github.com/PulpQE/pulp-smash/issues/459.

Ichimonji10:
Tested against Pulp 2.11 and 2.12 on Fedora 24, RHEL 6.8 and RHEL 7.3.

elyezer:
It will give more flexibility for sure.

PR #464, opened 2017-01-09
Use sudo when running getfattr

elyezer:
Using sudo if not root use will prevent permission denied error when
trying to getfattr on /var/log/pulp.


Ichimonji10:
Tested against Fedora 24, RHEL 6.8 and RHEL 7.3 with Pulp 2.11 and 2.12. Works. Thank you!

PR #466, opened 2017-01-09
Drop BaseAPITestCase from test_login

Ichimonji10:
The `pulp_smash.utils.BaseAPITestCase` is fragile when dealing with
cleanup logic, and the addition of `pulp_smash.selectors.skip_if`
greatly reduces the need for such a method. Do not introduce newbies to
the class.


Ichimonji10:
As a bonus, the module is a bit shorter.

PR #467, opened 2017-01-09
Replace Service with ServiceManager

Ichimonji10:
The `Service` class represents a service (such as httpd) on a host. It
provides methods such as `start` and `stop`, and it abstracts away which
service manager is in use on the target host.

A drawback of the `Service` class is that it encourages serial code to
be written. If three services are to be restarted, it's easy to write
something like this:

    services = (Service(cfg, name) for name in ('foo', 'bar', 'biz'))
    for service in services:
        service.stop()
    for service in services:
        service.start()

The `ServiceManager` class represents the service manager (such as
systemctl) on a host. It's used like so:

    services = ('foo', 'bar', 'biz')
    svc_mgr = ServiceManager(cfg)
    svc_mgr.stop(services)
    svc_mgr.start(services)

Depending on the service manager used by the target system, `stop` and
`start` dispatch commands to the target system either in serial or in
parallel. This has the potential to improve runtime performance.

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -242,63 +242,42 @@ class Client(object):  # pylint:disable=too-few-public-methods
         return self.response_handler(completed_process)
 
 
-class Service(object):
-    """A service on a system.
+class ServiceManager(object):
+    """A service manager on a system.
 
-    Each instance of this class represents a single service running on a single
-    system. An example may help to clarify this idea:
+    Each instance of this class represents the service manager on a system. An
+    example may help to clarify this idea:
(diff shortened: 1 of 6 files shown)


Ichimonji10:
Tested with Fedora 24, RHEL 6.8 and RHEL 7.3 with Pulp 2.11 and  2.12. Empirical tests show that the execution time of `pulp_smash.utils.reset_pulp` drops by 40% or more for systemd-based distribution — on single-threaded VMs. Multi-threaded VMs will likely see even greater gains.

PR #468, opened 2017-01-09
Drop BaseAPITestCase from test_consumer

Ichimonji10:
The `pulp_smash.utils.BaseAPITestCase` is fragile when dealing with
cleanup logic, and the addition of `pulp_smash.selectors.skip_if`
greatly reduces the need for such a method. Drop it.


Ichimonji10:
Tested with Fedora 24, RHEL 6.8 and RHEL 7.3 with Pulp 2.11 and 2.12.

PR #469, opened 2017-01-09
Fix how command is built

Ichimonji10:
Module `pulp_smash.tests.platform.cli.test_selinux` contains CLI
command-building logic. Here's an example of a command that might be
built:

    getfattr --recursive --name=security.selinux /var/cache/pulp

Unfortunately, if `sudo` is required, the following command is built:

    sudo --recursive getfattr --name=security.selinux /var/cache/pulp

Fix this. Build the CLI command in strictly left-to-right order, in
order to ease human understanding of what's going on.

### DIFF
--- a/pulp_smash/tests/platform/cli/test_selinux.py
+++ b/pulp_smash/tests/platform/cli/test_selinux.py
@@ -41,9 +41,8 @@ class ProcessLabelsTestCase(unittest.TestCase):
     def setUpClass(cls):
         """Get all of the processes running on the target Pulp system."""
         cfg = config.get_config()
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against Fedora 24, RHEL 6.8 and RHEL 7.3 with Pulp 2.11 and 2.12. This should fix a test failure.

Ichimonji10:
I'm going to go ahead and merge this PR. It's a small, straightforward change that should immediately improve our test results, and if I'm wrong, we'll find out tomorrow.

elyezer:
Thank you very much! I've not considered when both sudo and recursive was set, was too anger for the quick fix.

PR #470, opened 2017-01-10
Skip some orphan tests for older versions of Pulp

Ichimonji10:
As of Pulp 2.12.0, Pulp reports how many content units were deleted when
a DELETE /pulp/api/v2/content/orphans/ API call is made. Only make
assertions about this fact when the Pulp system under test is Pulp
2.12.0 or newer.

See: https://pulp.plan.io/issues/1268

See: 0a063c063defaec4dd122974ca4f99716ab823b4


Ichimonji10:
This should fix some test failures for the Pulp 2.11 test matrix.

Ichimonji10:
Much like https://github.com/PulpQE/pulp-smash/pull/469, this PR is simple and should immediately fix several tests. Merging.

Issue #477, opened 2017-01-13
Creating a group no longer updates the existing group


omaciel:
Please see more information [here](https://pulp.plan.io/issues/2514)

PR #488, opened 2017-01-17
Add a test for the .../import_upload/ API call

Ichimonji10:
If an existing package group is uploaded to a repository, subsequent
uploads update the existing package group.

See: https://pulp.plan.io/issues/2514

Fix: https://github.com/PulpQE/pulp-smash/issues/477


Ichimonji10:

Issue #472, opened 2017-01-12
Take into account the new publish layout


goosemania:
With Pulp 2.12 the following changes to the publish layout will be introduced 
http://pulpproject.org/2016/11/14/yum-repo-layout-changes/

Those changes apply to RPMs and SRPMs only, DRPMs will stay where they are now in `drpms/`:
```
repo
├── drpms
├── Packages
└── repodata
````

So it is needed to identify and fix tests which rely on the file structure of published repos.

As for upgrade, it can verify for existing repositories that:
- if there was no re-publish => old structure is still in place
- if re-publish was an incremental one => there will be a mixture of the old and new structure, new recently added RPMs/SRPMs will be organized in a new structure.
- if re-publish was full => only new structure should be used.

Redmine story: https://pulp.plan.io/issues/1976

Ichimonji10:
AFAIK, all of the affected tests now function on both Pulp 2.11 and 2.12. I'm going to fix function `pulp_smash.tests.rpm.api_v2.utils.get_unit` so that it makes use of the metadata provided by the `repodata/` directory. That done, this issue can be closed.

PR #489, opened 2017-01-17
Change how package paths are calculated

Ichimonji10:
The paths to packages in published repositories have changed. In Pulp
2.11 and earlier, packages ended up in the root of published
repositories, alongside the "repodata" directory. In Pulp 2.12, packages
default to being placed elsewhere, and users have the option to
customize the exact path.

The end result of this change is that clients should consult the
`repodata/...primary.xml` file when finding packages. This commit is not
so elegant, and instead implements a stop-gap solution. This stop-gap
solution should fix many broken tests, but does not yet consult with the
XML files.

See: https://github.com/PulpQE/pulp-smash/issues/472


Ichimonji10:

Issue #473, opened 2017-01-12
Expand upload tests to assert all the required fields are populated


preethit:
https://pulp.plan.io/issues/2365

1. upload an RPM
2. query the REST API for that RPM
3. look at the "files" field and verify that it is populated

PR #491, opened 2017-01-18
Update UploadRpmTestCase

Ichimonji10:
Make at least the following changes:

* Make the test case tear-down procedure more robust, so that it
  executes in the face of a greater variety of failures.
* Fix an error related to the new publish layout in Pulp 2.12.
* Remove assertions that check for exact status codes. These assertions
  greatly expand the size and complexity of the test case, and
  experience indicates that such assertions rarely if ever help in
  finding issues.
* Expand the assertions for RPM metadata tracking.

Fix: https://github.com/PulpQE/pulp-smash/issues/473

See: https://pulp.plan.io/issues/2365

See: https://github.com/PulpQE/pulp-smash/issues/472


Ichimonji10:

This PR lets the relevant test case run faster. I've observed a drop in execution time from ~35-40 seconds down to ~20-25 seconds.

PR #498, opened 2017-01-20
Do not use labels to categorize pull requests

Ichimonji10:
Instead of labels, use GitHub's built-in "pull request reviews" feature.

See: https://help.github.com/articles/about-pull-request-reviews/

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -120,18 +120,10 @@ Join the #pulp IRC channel on `freenode`_ if you have further questions.
 Labels
 ~~~~~~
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
If accepted, the relevant labels need to be removed from this repository. We can do the same for Pulp Fixtures, too.

elyezer:
Already started using Github's review feature feature on another PR and I am doing the same here. I think it does not make sense having the labels since in addition to the labels we are already using the review feature status. Dropping the labels will drop some of the duplicate meaning and work we have for stating PR statuses.

Thank you for making sure our docs is updated.

PR #499, opened 2017-01-20
Drop unnecessary function

Ichimonji10:
Drop a function that is only used once and returns an unnecessarily
complicated data structure.


Ichimonji10:
This function was created some time ago, when I first encountered content unit unassociation. It seems inexperience bred complexity.

Ichimonji10:
I'm merging this now because it's extremely simple; because I can verify that it works; and because it opens up the ability to file a subsequent, much more important and complicated PR.

PR #500, opened 2017-01-20
Fix several tests that rely on package paths

Ichimonji10:
As of Pulp 2.12, the layout of published RPM repositories has changed.
Fix several tests, so that they work with both earlier and later
versions of Pulp.


Ichimonji10:
I'm just filing this PR  to ensure that Travis passes. It's high priority, though, and I'm going to merge it right away.

PR #509, opened 2017-01-24
Update export distributor package path

elyezer:
Make sure to use the new repository layout when asserting for RPMs
exported to a directory by the export distributor. Use the old layout
when testing Pulp < 2.12.


elyezer:
I will wait travis being green and will merge this. I am about to run automation again and want this to be included so I can finish the regression test result analysis.

PR #514, opened 2017-01-26
Make get_unit() use repodata/*.xml

Ichimonji10:
When an RPM repository is published, several XML files are placed in a
`repodata/` directory. These XML files describe various attributes of
the repository, including the relative paths to the RPM files. Make
function `pulp_smash.tests.rpm.api_v2.utils.get_unit` use these XML
files.

Fix: https://github.com/PulpQE/pulp-smash/issues/472


Ichimonji10:
Tested against a matrix of four systems, with one axis being Fedora 24 and Fedora 25, and the other axis being Pulp 2.11 and Pulp 2.12. All affected modules passed.

PR #516, opened 2017-01-27
Better manage pulp_manage_rsync SELinux policy

Ichimonji10:
Module `pulp_smash.tests.rpm.api_v2.test_rsync_distributor` manages the
`pulp_manage_rsync` SELinux policy if appropriate. As part of that task,
it executes `which semanage`. On some platforms, the executable is
installed in `/sbin`, which causes the check to fail when run as a
non-root user.

Fix this check to use `sudo` if appropriate.


Ichimonji10:
Tested on a matrix of four systems, with one axis being Pulp 2.11 and 2.12, and the other axis Fedora 24 and 25.

Issue #494, opened 2017-01-19
Test syncing wheel packages


preethit:
https://pulp.plan.io/issues/135

1. Create a python repo with feed to wheel packages
2. Sync the repo
Assert:  The sync completes and  wheel packages are synced successfully

This might need a new fixture repo

Issue #493, opened 2017-01-19
Test that pulp to pulp sync for python packages work


preethit:
https://pulp.plan.io/issues/140

1. Create  python repoA with feed
2. Sync repoA
3. Create repoB with --feed repoA
4. Run Sync
Assert: Sync is success

1. Create  python repoA with no feed
2. Upload content to the repo
3. Publish the repo
4. Create repoB with --feed repoA
5. Run Sync
Assert: Sync is success

Ichimonji10:
@dkliban clarified that, for this issue to be fully tested, we need to sync from one Pulp application to another completely independent Pulp application.

Issue #492, opened 2017-01-19
Test uploading wheel packages


preethit:
https://pulp.plan.io/issues/136

Test that python plugin support wheel packages

PR #531, opened 2017-01-31
Rewrite Python tests

Ichimonji10:
Rewrite module `pulp_smash.tests.python.api_v2.test_sync_publish`. Make
at least the following changes:

* Change which issues are queried when deciding whether or not to raise
  a `SkipTest` exception, and as a consequence, increase the amount of
  test code that's executed.
* Expand the tests, so that repo-to-repo syncs are performed.
* Expand the tests, so that content unit uploads are performed.

Fix: https://github.com/PulpQE/pulp-smash/issues/492

Fix: https://github.com/PulpQE/pulp-smash/issues/493

Fix: https://github.com/PulpQE/pulp-smash/issues/494

### DIFF
--- a/pulp_smash/tests/python/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/python/api_v2/test_sync_publish.py
@@ -1,36 +1,80 @@
 # coding=utf-8
 """Test the sync and publish API endpoints for Python repositories."""
+import inspect
 import unittest
-from urllib.parse import urljoin
-
-from packaging.version import Version
+from os.path import basename
+from urllib.parse import urljoin, urlparse
 
 from pulp_smash import api, config, constants, selectors, utils
-from pulp_smash.tests.python.api_v2.utils import gen_repo
+from pulp_smash.tests.python.api_v2.utils import (
+    gen_distributor,
+    gen_repo,
+    upload_import_unit,
+)
 from pulp_smash.tests.python.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
 
 
-class UtilsMixin(object):
-    """Methods for use by the test cases in this module.
+class BaseTestCase(unittest.TestCase):
+    """A base class for the test cases in this module.
 
-    Classes inheriting from this mixin must also inherit from
-    ``unittest.TestCase``.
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Tested against a matrix of six systems, with one axis being Pulp 2.11 and Pulp 2.12, and the other axis being Fedora 24, Fedora 25, and RHEL 7.3. The tests are entirely skipped on Pulp 2.11, due to the fact that support for the tested features isn't in place until 2.12.

Tests partially skip on Pulp 2.12 due to https://pulp.plan.io/issues/140.

pulp_smash/tests/python/api_v2/test_sync_publish.py
```diff
@@ -46,80 +90,92 @@ def verify_sync(self, cfg, call_report):
...
+        'distributors': True,
+        'importers': True,
+    }).json()
+
+
+def get_repo_path(repo):
+    """Return the root path to a published Python repository."""
+    path = urljoin('/pulp/python/web/', repo['id'])
+    path += '/'
+    return urljoin(path, 'simple/')
```
mhrivnak: I wonder if adding "simple" is perhaps not the right thing to do when running a sync. The docs are admittedly way too thin, which we'll get an issue filed for. But note that in this recipe, when doing a sync, it does not include "simple" in its path:

https://github.com/pulp/pulp_python/blob/master/docs/user-docs/getting_started.rst#synchronize-packages-from-pypi

Ichimonji10: I'll drop `simple/`.

For the record, the test result is the same - no units are synced.

Ichimonji10: It turns out that dropping `simple/` is the correct thing to do.

See: https://pulp.plan.io/issues/140#note-18

Ichimonji10:
Tested against a matrix of six systems, with one axis being Pulp 2.11 and Pulp 2.12, and the other axis being Fedora 24, Fedora 25, and RHEL 7.3. All tests either skip (due to the feature not being implemented in Pulp 2.11) or pass.

PR #533, opened 2017-02-01
Rewrite function import_upload_unit

Ichimonji10:
After a content unit has been uploaded to Pulp, an HTTP POST request
must be sent to actually import it into a repository. Depending on the
type of unit being imported, the body of that HTTP POST request varies.
Unfortunately, `pulp_smash.utils.upload_import_unit` is inflexible in
this regard.

Rewrite the function so that arbitrarily complex JSON data may be sent
as the body of the HTTP POST request. This allows `import_upload_unit`
to be used when uploading and importing Python content units.

See: 9973e8911f1f1d3d8bcd030b81dd2b1059d756d3


Ichimonji10:
Tested against a matrix of six systems, with one axis being Pulp 2.11 and Pulp 2.12, and the other axis being Fedora 24, Fedora 25 and RHEL 7.3.

PR #535, opened 2017-02-02
Make reset_squid() work with Squid version 4

Ichimonji10:
Function `pulp_smash.utils.reset_squid` stops squid, deletes and
recreates its cache, and starts squid. As part of the cache recreation
process, it executes the following command: [1]

    squid -z

This works in most cases. However, if all of the following are true...

* The command is executed over SSH.
* The remote system has SELinux enabled.
* Squid version 4 is installed.

...then the command will leave stray files in `/dev/shm/`. These files
will have SELinux contexts set in such a way that Squid is unable to
touch them, and as a result, Squid will be unable to successfully start.

For whatever reason, a solution is to use the `--foreground` flag:

    squid -z --foreground

This flag was introduced in Squid 4: [2]

> The squid binary now has a new --foreground command line option which
> prevents the process from exiting early while background workers
> continue their processing. When run with this option Squid will now
> wait for the worker(s) to finish before exiting. Unlike the old -N
> option --foreground supports SMP workers and multi-process features.
> --foreground is particularly useful for use with -z (disk cache
> structures creation), as it allows the caller to wait until Squid has
> finished.

Thus, if Squid 4 or newer is installed, the `--foreground` command is
appended to the cache-recreation command.

[1] `sudo` is prepended if appropriate.
[2] http://www.squid-cache.org/Versions/v4/RELEASENOTES.html

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -11,6 +11,7 @@ import uuid
 from urllib.parse import urljoin, urlparse
 
 import requests
+from packaging.version import Version
 
 from pulp_smash import api, cli, config, exceptions
 from pulp_smash.cli import _is_root as is_root  # for backward compatibility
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a matrix of six systems, where one axis is Fedora 24, Fedora 25 and RHEL 7.3, and the other axis is Pulp 2.11 and Pulp 2.12.

Ichimonji10:
@seandst You might find this PR interesting.

PR #538, opened 2017-02-03
Consolidate logic for searching for units

Ichimonji10:
Create function `pulp_smash.utils.search_units`. Use it in place of all
similar functions and all direct API calls to path
`{repo_href}/search/units/`.

### DIFF
--- a/pulp_smash/tests/docker/api_v2/test_tags.py
+++ b/pulp_smash/tests/docker/api_v2/test_tags.py
@@ -52,18 +52,6 @@ def create_docker_repo(cfg, upstream_name, use_v1=False):
     return client.post(REPOSITORY_PATH, body)
 
 
-def search_units(cfg, repo, criteria):
-    """Helper to search repository units.
-
-    :param pulp_smash.config.ServerConfig cfg: Information about a Pulp host.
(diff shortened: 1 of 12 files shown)


Ichimonji10:
Tested against a matrix of six systems, with one axis being Pulp 2.11 and Pulp 2.12, and the other axis being Fedora 24, Fedora 25 and RHEL 7.3. Test results were unchanged.

pulp_smash/tests/docker/api_v2/test_tags.py
```diff
@@ -96,12 +84,15 @@ def setUp(self):
         self.tags = self._get_tags()
 
     def _get_tags(self):
-        return search_units(self.cfg, self.repo, {'type_ids': ['docker_tag']})
+        return utils.search_units(
+            self.cfg,
+            self.repo, {'type_ids': ['docker_tag']}
```
elyezer: Do you mind breaking the line after the `,` so the criteria dict goes on it own line?

Ichimonji10: Done! This syntax is weird. I'm not sure why I used it.

PR #545, opened 2017-02-03
Skip some Python tests unless testing Pulp 2.12+

Ichimonji10:
Tests in module `pulp_smash.tests.python.api_v2.test_sync_publish` will
already skip if the relevant Pulp issues are open. However, the
"Platform Release" field is not set on the relevant issues, meaning that
the tests will run even for very old versions of Pulp, and thus causing
unnecessary test failures. Fix this.


Ichimonji10:
Merging to fix test failures.

Issue #528, opened 2017-01-30
unhandled exception when changing feed to equivalent repo


elyezer:
This can be done by having some different repositories and playing with syncing between them.

For more information see: https://pulp.plan.io/issues/1922

PR #548, opened 2017-02-09
Test a repository's feed can be changed

Ichimonji10:
Add `ChangeFeedTestCase`. From its docstring:

> The test case procedure is as follows:
>
> 1. Create three repositories — call them A, B and C.
> 2. Populate repository A and B with identical content, and publish
>    them.
> 3. Set C's feed to repository A. Sync and publish repository C.
> 4. Set C's feed to repository B. Sync and publish repository C.
>
> The entire procedure should succeed. This test case targets `Pulp
> #1922 <https://pulp.plan.io/issues/1922>`_.

Fix: https://github.com/PulpQE/pulp-smash/issues/528

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -27,9 +27,10 @@ from pulp_smash.constants import (
     RPM_MISSING_PRIMARY_FEED_URL,
     RPM_SIGNED_FEED_COUNT,
     RPM_SIGNED_FEED_URL,
+    RPM_UNSIGNED_FEED_URL,
     SRPM_SIGNED_FEED_URL,
 )
-from pulp_smash.tests.rpm.api_v2.utils import gen_repo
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against RHEL 7.3 with Pulp 2.12.

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -240,3 +241,69 @@ def do_test(self, feed_url):
...
+    def create_sync_publish_repo(self, body):
+        """Create, sync and publish a repository.
+
+        Also, schedule the repository for deletion.
+
+        :param body: A dict of information to use when creating the repository.
+        :return: A detailed dict of information about the repository.
+        """
+        client = api.Client(self.cfg)
+        repo = client.post(REPOSITORY_PATH, body).json()
```
elyezer: If call to `.json()` is being doing why not use `api.json_handler`?

Ichimonji10: No good reason. I whiffed.

PR #549, opened 2017-02-09
Refactor SyncInvalidFeedTestCase

Ichimonji10:
Make at least the following changes:

* Ensure the class cleans up after itself in the face of unexpected
  errors.
* Make a `skipTest` call conditional, not absolute.
* Make assertion error messages more verbose.


Ichimonji10:
Tested against RHEL 7.3 with Pulp 2.12.

This change is low priority. It's just one more edit that will make Pulp Smash better about cleaning up after itself.

Issue #526, opened 2017-01-30
rsync distributor without force_full incorrectly skips publishing some units


elyezer:
Using a yum repo as an example, here's a sequence of events which demonstrates the problem:

1. Create a yum repository, and populate it with some content.
2. Publish the repository with a yum distributor.
3. Add additional content to the repository.
4. Publish the repository with an rsync distributor. (This publish will not include the new unit, since a yum publish hasn't happened for the unit.)
5. Publish the repository with a yum distributor.
6. Publish the repository with an rsync distributor. (This publish should include the new unit, since a yum publish has happened for the unit.)

For more information see: https://pulp.plan.io/issues/2532

preethit:
Related issue 
https://pulp.plan.io/issues/2550#note-5

PR #554, opened 2017-02-10
Add rsync distributor test: AddUnitsTestCase

Ichimonji10:
Add `AddUnitsTestCase` to module
`pulp_smash.tests.rpm.api_v2.test_rsync_distributor`. From its
docstring:

> When executed, this test case does the following:
>
> 1. Create a yum repository with a yum and rsync distributor.
> 2. Add some content to the repository.
> 3. Publish the repository with its yum distributor.
> 4. Add additional content to the repository.
> 5. Publish the repository with its rsync distributor. This publish
>    shouldn't distribute the new content unit, as the new content unit
>    wasn't included in the most recent publish with the yum
>    distributor.
> 6. Publish the repository with its yum distributor.
> 7. Publish the repository with its rsync distributor. This publish
>    should distribute the new content unit, as the new content unit was
>    included in the most recent publish with the yum distributor.

Fix: https://github.com/PulpQE/pulp-smash/issues/526

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -286,6 +286,9 @@ REPOSITORY_PATH = '/pulp/api/v2/repositories/'
 RPM = 'bear-4.1-1.noarch.rpm'
 """The name of an RPM file. See :data:`pulp_smash.constants.RPM_SIGNED_URL`."""
 
+RPM2 = 'camel-0.1-1.noarch.rpm'
+"""The name of an RPM. See :data:`pulp_smash.constants.RPM2_UNSIGNED_URL`."""
+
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested against a matrix of four systems, with one axis being Fedora 24 and Fedora 25, and the other axis being Pulp 2.12 and 2.13. I tested with the `self.skipTest` line removed.

Ichimonji10:
@elyezer, notice that I'm uploading units instead of syncing an entire repository. I'm going to intentionally avoid giving any specific numbers at this time, but anecdotally, this change dramatically shortened the test execution time. If we want to make Pulp Smash run more quickly, it may be appropriate to upload content units in more cases.

PR #556, opened 2017-02-14
Make tests more verbose

Ichimonji10:
Several tests in module
`pulp_smash.tests.rpm.api_v2.test_download_policies` verify that HTTP
messages from Pulp have correct headers. When these tests fail, they
don't print information about the actual contents of the headers. Fix
this.


elyezer:
This is a much needed change, thank you.

PR #561, opened 2017-02-21
Skip certain Python tests

Ichimonji10:
Version 2 of the pulp_python plugin will add support for dealing with
python wheels, and most of the necessary work is already done. As a
result, certain issues on Pulp's issue tracker are marked as MODIFIED,
and several of Pulp Smash's selectors designed to skip tests were not
producing test skips. Fix this.


Ichimonji10:
This should fix four test failures in the 2.12.1 beta.

PR #562, opened 2017-02-22
Disable SELinux when managing services on Jenkins

elyezer:
When managing SYSV services from Jenkins services and related files gets
the `unconfined_java_t` label. That is because SYSV does not deal great
with the labels as systemd. In order to get better results, on RHEL6,
when running on Jenkins disable SELinux temporarily when restarting the
services for pulp_reset.


elyezer:
I will improve the code to both avoid code repetition and use a context manager. The latter will ensure that if for some reason an exception is raised when managing a service the SELinux will continue to be restored.

PR #570, opened 2017-02-28
Make test more robust in the face of failures

Ichimonji10:
Method `SyncPublishV2TestCase.setUpClass` does quite a bit of work,
including creating and syncing a repository and fetching some additional
information about the repository. This design has two weaknesses:

* If the repository sync fails, it doesn't make sense to continue on and
  get additional information about the repository. The test should stop.
* If the repository creation or sync fails, no clean-up work will occur.

Fix these issues. This is an issue that was encountered in the real
world during troubleshooting.


Ichimonji10:
Tested against F24 with Pulp 2.12 and 2.13, F25 with Pulp 2.12 and 2.13, RHEL 6.8 with Pulp 2.12, and RHEL 7.3 with Pulp 2.12.

PR #571, opened 2017-03-01
Skip test failing due to a bug in docker

Ichimonji10:
The bug is entitled:

> "blob" files are delivered with incorrect content headers

See: https://pulp.plan.io/issues/2618


Ichimonji10:
Merging this now, as Elyezer is out on PTO, and this is a simple change that'll fix several test failures.

Issue #540, opened 2017-02-03
Empty pkglist for the erratum may be generated which may cause yum not to find applicable packages


preethit:
https://pulp.plan.io/issues/2559

Look in redmine for more info. I will try to come up with steps to test this.

preethit:
https://github.com/PulpQE/pulp-smash/commit/96dc91a6c23eb29f620bdd2bb11abb28747353ec

Ichimonji10:
I just need to submit one more PR, and all of the errata tests should be fixed.

Ichimonji10:
https://github.com/PulpQE/pulp-smash/commit/96dc91a6c23eb29f620bdd2bb11abb28747353ec and https://github.com/PulpQE/pulp-smash/commit/9bd4aec75a5ea9b3bb3e29489f53b1841536c097 fix this issue.

PR #576, opened 2017-03-07
Rewrite test that updates updateinfo.xml

Ichimonji10:
Class `ErratumPkgListCountTestCase` does the following:

1. Create and sync an RPM repository with a feed and distributor.
2. Unassociate the "gorilla" package from the repository.
3. Publish the repository.

It then verifies that the published repository's `updateinfo.xml` file
is valid. Unfortunately, this test recently broke, primarily because of
the changes implemented for Pulp issue 2559: errata are omitted if they
don't reference any packages, or if they reference packages not in the
currently published repository. In this case, an erratum that references
the "gorilla" erratum is now being omitted, which broke the test.

Update the test case:

* Make the test case comply with the changes outlined in Pulp issue
  2559.
* Make the test clean up after itself even when an error occurs during
  test case set-up.
* Rename the test case to `UpdateRepoTestCase`, and modify its logic.

The test case now does the following:

1. Create, sync and publish an RPM repository with a feed and
   distributor.
2. Verify the repository's `updateinfo.xml` file.
3. Unassociate the "gorilla" package from the repository.
4. Verify the repository's `updateinfo.xml` file.

See:

* https://github.com/PulpQE/pulp-smash/issues/540
* https://github.com/PulpQE/pulp-smash/pull/574
* https://pulp.plan.io/issues/2559

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_updateinfo.py
+++ b/pulp_smash/tests/rpm/api_v2/test_updateinfo.py
@@ -246,76 +246,80 @@ class UpdateInfoTestCase(utils.BaseAPITestCase):
         )
 
 
-class ErratumPkgListCountTestCase(utils.BaseAPITestCase):
-    """Tests for ``updateinfo.xml`` erratum package list count."""
+class UpdateRepoTestCase(utils.BaseAPITestCase):
+    """Verify ``updateinfo.xml`` changes as the its repo changes."""
 
     @classmethod
     def setUpClass(cls):
-        """Create RPM repository, delete a package, and publish the repository.
-
-        More specifically, do the following:
-
-        1. Create an RPM repository with a distributor.
-        2. Sync the created repository.
-        3. Remove the ``gorilla`` package
-        4. Publish the repository. Fetch the ``updateinfo.xml`` file from the
-           distributor (via ``repomd.xml``), and parse it.
-        """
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Before this change, the test case was totally broken due to the changes listed in https://pulp.plan.io/issues/2559. After this change, the test case is *still* broken, but for a new reason.  Basically, fixing this one issue let me find a new one.

Once I nail down the new bug, I'll be able to fix *that* one by simply skipping `test_01_...`.

pulp_smash/tests/rpm/api_v2/test_updateinfo.py
```diff
@@ -246,76 +246,80 @@ def test_reboot_not_suggested(self):
...
-        2. Sync the created repository.
-        3. Remove the ``gorilla`` package
-        4. Publish the repository. Fetch the ``updateinfo.xml`` file from the
-           distributor (via ``repomd.xml``), and parse it.
-        """
-        super(ErratumPkgListCountTestCase, cls).setUpClass()
-
-        # Create a repository.
+        """Create an RPM repository with a feed and distributor."""
+        super().setUpClass()
```
elyezer: Python 3 :heart:

Ichimonji10: :D

PR #577, opened 2017-03-08
Update tests for `updateinfo.xml`

Ichimonji10:
When Pulp syncs and publishes a repository, it's free to restructure the
repository at will, so long as the result is semantically valid. One
such restructuring is to condense multiple `<pkglist>` elements. Another
is to change the "short" attribute of a `<collection>` element. There
are more.

Update module `pulp_smash.tests.rpm.api_v2.test_updateinfo` so that its
test cases only make assertions on important information, like errata
IDs and the presence of certain packages.


Ichimonji10:
This fixes two failures in `pulp_smash.tests.rpm.api_v2.test_updateinfo`. Tested against six systems, running nightlies of Pulp 2.12 and 2.13.

seandst:
This looks good to me, but as one of the co-conspirators that changed Pulp's behavior to necessitate this change I can't be trusted. :)

pulp_smash/tests/rpm/api_v2/test_updateinfo.py
```diff
@@ -331,52 +356,58 @@ class PkglistsTestCase(unittest.TestCase):
...
-        # Create and sync a repository.
+        # Create, sync and publish a repository.
         client = api.Client(cfg, api.json_handler)
         body = gen_repo()
         body['importer_config']['feed'] = RPM_PKGLISTS_UPDATEINFO_FEED_URL
         body['distributors'] = [gen_distributor()]
         repo = client.post(REPOSITORY_PATH, body)
-        repo = client.get(repo['_href'], params={'details': True})
         self.addCleanup(client.delete, repo['_href'])
+        repo = client.get(repo['_href'], params={'details': True})
```
Ichimonji10: I'm moving this line of code down so that `addCleanup()` executes as soon as possible.

pulp_smash/tests/rpm/api_v2/test_updateinfo.py
```diff
@@ -268,18 +305,13 @@ def setUpClass(cls):
...
-
-          .. code-block:: xml
-
-            <collection short="">
-                <name>1</name>
-            </collection>
+        * An ``<update>`` with an ``<id>`` of
+          :data:`pulp_smash.constants.RPM_ERRATUM_ID` is present, and
+        * one of its child ``<package>`` elements has a "name" attribute equal
+          to :data:`pulp_smash.constants.RPM_ERRATUM_RPM_NAME`.
```
Ichimonji10: As mentioned in the PR title, I'm making the test assertions only on "important" information. This change in the docstring reflects the change in focus.

Ichimonji10:
If it helps, here's an example of an `updateinfo.xml` file that Pulp can sync in: http://pastebin.com/g7E3Ra97

And here's an example of an `updateinfo.xml` file published by Pulp: http://pastebin.com/qBcC8BiX

The file is restructured but valid.

Issue #524, opened 2017-01-30
RPM import traceback (non-utf-8 metadata slipping through)


elyezer:
Observation: there is logic to deal with UTF-8 characters when syncing but not when uploading a RPM.

For more information see: https://pulp.plan.io/issues/1903

preethit:
https://pulp.plan.io/issues/2622

Ichimonji10:
Automating this issue requires that we upload an RPM that includes non-utf8 data in its header. Creating such a package requires that a `.spec` including this kind of data file be written, and that an RPM package be created from that.

Right now, I don't know how to create RPM packages, and I've fallen back to automating this issue manually. I can automate this issue after learning more about RPM packaging. Alternately, someone else with the relevant knowledge could add the necessary RPM to Pulp Fixtures.

elyezer:
@Ichimonji10 talked with @goosemania and she recommended this link http://www.redhat.com/archives/rpm-list/2006-April/msg00015.html

Then after creating the spec, rpmbuild should be run.

PR #586, opened 2017-03-15
Add tests for RPM metadata character encodings

Ichimonji10:
According to Pulp issue 1903, it should be possible to upload an RPM to
Pulp even when that RPM has non-utf8 metadata in its header. Add API and
CLI tests to check whether this is true.

See: https://pulp.plan.io/issues/1903

Fix: https://github.com/PulpQE/pulp-smash/issues/524


Ichimonji10:
This PR depends on https://github.com/PulpQE/pulp-fixtures/pull/53.

Issue #583, opened 2017-03-13
Test  puppet repo that doesn't have a feed  fail gracefully


preethit:
Test that trying to sync a puppet repo that does not have a feed fails gracefully

pulp-admin puppet repo create --repo-id pup1
pulp-admin -v rpm repo sync run --repo-id pup
https://pulp.plan.io/issues/2628

PR #591, opened 2017-03-16
Add test case to sync Puppet repo w/out a feed

Ichimonji10:
If a Puppet repository without a feed is synced, the sync should fail in
a graceful way. Add a test to verify that this is so.

See: https://pulp.plan.io/issues/2628

Fix: https://github.com/PulpQE/pulp-smash/issues/583


Ichimonji10:
Tested against F24, F25 and RHEL 7 with Pulp 2.12.2 beta. The test currently skips, and if the `self.skipTest()` statement is removed, the test fails.

I know this isn't high priority. I just... decided to take a mental break and do something easy.

Issue #521, opened 2017-01-30
Files ending in .gz are delivered with incorrect content headers


elyezer:
Files ending in .gz served up by pulp.server.content.web.views are served with an incorrect content type and encoding.

This affects .xml.gz files generated by yum-distributor. For example:

```
$ curl -v http://192.168.121.51/pulp/repos/f04a1d5b-32d8-4a60-a580-cc158ab25004/repodata/5f90923d6f2e4b3d0f27c13e927821b9a7001b349f99222752c37496a40531d8-updateinfo.xml.gz > /dev/null
*   Trying 192.168.121.51...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 192.168.121.51 (192.168.121.51) port 80 (#0)
> GET /pulp/repos/f04a1d5b-32d8-4a60-a580-cc158ab25004/repodata/5f90923d6f2e4b3d0f27c13e927821b9a7001b349f99222752c37496a40531d8-updateinfo.xml.gz HTTP/1.1
> User-Agent: curl/7.40.0
> Host: 192.168.121.51
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 17 Mar 2016 03:24:50 GMT
< Server: Apache/2.4.18 (Fedora) OpenSSL/1.0.2g-fips mod_wsgi/4.4.8 Python/2.7.10
< Last-Modified: Thu, 17 Mar 2016 00:57:43 GMT
< ETag: "278-52e341e355108" 
< Content-Length: 632
< Content-Type: text/xml
< 
```

The server incorrectly claims that the response content is Content-Type: text/xml, with no Content-Encoding. It should be: Content-Type: text/xml, Content-Encoding: gzip.

For more information check https://pulp.plan.io/issues/1781

preethit:
Test case- 1 Sync

1. Create and sync the yum  "rpm-signed" from fixtures
2. Curl -v GET --insecure https://localhost/pulp/repos/pulp/pulp/fixtures/rpm-signed/repodata/<name>-updateinfo.xml.gz
3. Check to see the Content-Type: is gzip
  Comment from commit: # Most OSes return the 'gzip' version, but some (el6) return 'x-gzip'. Either is fine.

Test Case2 upload

1. Create a yum repo
[...]

PR #592, opened 2017-03-17
Test Puppet download HTTP headers

Ichimonji10:
Rewrite test case `SyncValidFeedTestCase` in module
`pulp_smash.tests.puppet.api-v2.test_sync_publish`. Make at least the
following changes:

* Drop the call to function `reset_pulp`. This call worked around an
  issue fixed in Pulp 2.8. We're now developing Pulp 2.13, and keeping
  it in simply lengthens the test suite execution time.
* Make the test case more robust in the case of failures, by ensuring
  that repositories are deleted even when unexpected failures occur.
* Drop several assertions relating to the tasks spawned during a
  repository sync. These checks are already done by
  `pulp_smash.api.safe_handler` and `pulp_smash.api.json_handler`.
* Add assertions to verify that the content units synced into a
  repository can be downloaded. This is the most important addition.

Fix: https://github.com/PulpQE/pulp-smash/issues/521

See: https://pulp.plan.io/issues/1781

### DIFF
--- a/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/puppet/api_v2/test_sync_publish.py
@@ -36,6 +36,7 @@ from itertools import product
 from urllib.parse import urljoin
 
 from packaging.version import Version
+from requests.exceptions import HTTPError
 
 from pulp_smash import api, config, exceptions, selectors, utils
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested against the 2.12.2 beta on F24, F25 and RHEL 7.3.

PR #593, opened 2017-03-17
Test RPM download HTTP headers

Ichimonji10:
Tweak several existing RPM test cases. Add assertions verifying that,
when a client downloads an RPM, that RPM is identified with a correct
content-type header.


Ichimonji10:
Tested against the 2.12.2 beta on F24, F25 and RHEL 7.3.

Related to https://github.com/PulpQE/pulp-smash/pull/592.

PR #597, opened 2017-03-22
Update test for Pulp issue 1376

Ichimonji10:
Pulp issue 1376 claims that, for a certain type of failed task, the
`["progress_report"]["yum_importer"]["content"]["error_details"]`
attribute should list error details. However, further discussion shows
that this is not true. The actual fix is to improve the contents of the
`["error"]["description"]` attribute.

Update the test for Pulp issue 1376 so that assertions are made about
the `["error"]["description"]` attribute.

See: https://pulp.plan.io/issues/1376


Ichimonji10:
This should fix a test failure.

PR #599, opened 2017-03-23
Make CLI test correctly parse stdout

Ichimonji10:
The output produced by ``pulp-admin docker repo copy`` changes in Pulp
2.13. Rather than listing the checksum of every copied unit, the total
number of copied units is listed. Update `CopyAllManifestsTestCase` so
that it can deal with both types of output.


Ichimonji10:
Tested against Pulp 2.12 on F24, F25 and RHEL 7, and Pulp 2.13 on F24 and F25.

Ichimonji10:
I'm merging this PR, as it fixes a current test failure. We'll see if I got this right tomorrow morning.

Issue #555, opened 2017-02-12
Test schema version 2 manifests published by pulp docker


dkliban:
[Install docker](https://docs.docker.com/engine/installation/). Get docker v2 content. Deploy registry on localhost by following example [here](https://docs.docker.com/registry/deploying/). Go as far as pushing the ubuntu image into the registry.

Create a docker v2 repo that will sync from local registry created earlier:

```bash
pulp-admin docker repo create \
  --feed http://127.0.0.1:5000 \
  --upstream-name ubuntu \
  --repo-id mydocker2 \
  --auto-publish false
```

Sync v2 content from the local registry: `pulp-admin docker repo sync run --repo-id mydocker2`

Clone crane: `git clone git@github.com:/pulp/crane`

Add the following /etc/crane.conf:

```ini
[general]
debug: true
data_dir: /var/lib/pulp/published/docker/v2/app/
endpoint: localhost:5001
```

Sync mydocker2 repo from the local registry: `pulp-admin docker repo sync run --repo-id mydocker2`

Publish mydocker2 repository: `pulp-admin docker repo publish run --repo-id mydocker2`

[Configure crane](https://github.com/pulp/crane/blob/master/docs/index.rst#deployment). It should run  using apache and mod_wsgi.

Docker pull from crane: `sudo docker pull localhost:5001/mydocker2`

Verify that pulling from crane returns a schema version 2 manifest. Grep through `/var/log/httpd/ssl_access_log`. The path accessed during pull should include:

```
<repo id>/manifests/2/latest 127.0.0.1 - - [12/Feb/2017:22:03:24 +0000] "GET /pulp/docker/v2/mydocker2/manifests/2/latest HTTP/1.1" 200 1357
```

ipanova:
Install newer docker client 1.9 and newer.

with already synced repo from some registry into pulp, and published it, so crane is serving it, do:

*PULL BY TAG*

1. docker pull manifest by tag( where you would choose a tag that references schema 2 version)
2. check in httpd access log the path, which tells you that manifest schema 2 was returned

1. docker pull manifest by tag( where you would choose a tag that references *just* schema 1 version)
2. check in httpd access log the path, which tells you that manifest schema 1 was returned

1. docker pull manifest by invalid tag
2. confirm that 404 is returned

*PULL BY DIGEST*

1. docker pull manifest by digest( which would identify schema 2 version)
[...]

PR #606, opened 2017-03-24
Update DockerTagTestCase

Ichimonji10:
As of Pulp 2.13, Pulp supports docker schemas version 1 and 2, instead
of just version 1. As a result, searching for a tag by its name now
returns two results instead of one, among other changes. Update
`DockerTagTestCase` so that it correctly searches for docker tags.

See: https://pulp.plan.io/issues/2099


Ichimonji10:
It turns out that this PR only fixes the 2.13 failures about 50% of the time. There's probably some schema v1/v2 shenanigans going on. I'm closing this.

I'm also going to open a bug asking for more docs on this. I've read through the stuff on the master pulp_docker branch on github, and I'm still resorting to running `pulp-admin -vvv ...`.

PR #607, opened 2017-03-27
Update DockerTagTestCase

Ichimonji10:
As of Pulp 2.13, Pulp supports docker schemas version 1 and 2, instead
of just version 1. As a result, search results for tags and manifests
can include two almost-identical results. Update `DockerTagTestCase` so
that searches filter by schema version when appropriate.

See: https://pulp.plan.io/issues/2099


Ichimonji10:
This is a corrected version of https://github.com/PulpQE/pulp-smash/pull/606.

Tested against Pulp 2.12 and 2.13 on F24 and F25. Fixes a test failure.

elyezer:
I've merged this so it will be included on the next run I am about to trigger.

PR #608, opened 2017-03-28
Run Python sync tests for Pulp 2.13+

Ichimonji10:
The tests in module `pulp_smash.tests.python.api_v2.test_sync_publish`
should run when Pulp 2.13 or newer is under test.


Ichimonji10:
Tested against Pulp 2.12 and 2.13 on F24 and F25. The un-skipped tests pass.

PR #609, opened 2017-03-28
Skip Docker v2 tests on Pulp 2.13 (for now!)

Ichimonji10:
`SyncPublishV2TestCase` in module
`pulp_smash/tests/docker/cli/test_sync_publish.py` syncs and publishes a
v2 Docker repository. Unfortunately, this test case also interacts with
the pulp_docker plugin directly, instead of talking to the Crane
plug-in. As a result, the changes to pulp_docker in Pulp 2.13 break this
test case.

Updating the test case to work well with either version of Pulp is
infeasible. The changes to pulp_docker are extensive, and they're
internal anyway: clients should talk to Crane.

Skip this test case if Pulp 2.13 or newer is under test. This "fix" is a
band-aid. The test should be re-written to use Crane — and the CI jobs
should be updated to install and configure Crane!

### DIFF
--- a/pulp_smash/tests/docker/cli/test_sync_publish.py
+++ b/pulp_smash/tests/docker/cli/test_sync_publish.py
@@ -105,6 +105,11 @@ class SyncPublishV2TestCase(unittest.TestCase):
         some information for use by the test methods.
         """
         cls.cfg = config.get_config()
+        if cls.cfg.version >= Version('2.13'):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against Pulp 2.12 and 2.13 on F24 and F25.

I tried making this test work on both versions of Pulp. It was a waste of time.

Issue #487, opened 2017-01-17
Test pulp-manage-db fails to run if other pulp processes are running


preethit:
https://pulp.plan.io/issues/2186

With: pulp_celerybeat, pulp_resource_manager, and pulp_workers running
Run: pulp-manage-db
Assert: pulp-manage-db fails with error

With: pulp_celerybeat, pulp_resource_manager, and pulp_workers running
Run: pulp-manage-db
Assert: pulp-manage-db fails with error

With: pulp_celerybeat running pulp_resource_manager, and pulp_workers stopped
Run: pulp-manage-db
Assert: pulp-manage-db fails with error

With: pulp_resource_manager running, and pulp_celerybeat, pulp_workers stopped
Run: pulp-manage-db
Assert: pulp-manage-db fails with error

With: pulp_workers running and pulp_celerybeat, pulp_resource_manager, stopped
Run: pulp-manage-db
Assert: pulp-manage-db fails with error

With: pulp_celerybeat, pulp_resource_manager, and pulp_workers stopped
Run: pulp-manage-db
Assert: pulp-manage-db runs successfully

PR #612, opened 2017-03-30
Add tests for conditional pulp-manage-db execution

Ichimonji10:
Some of the tests may be skipped due to:

* https://pulp.plan.io/issues/2186
* https://pulp.plan.io/issues/2684

Fix: https://github.com/PulpQE/pulp-smash/issues/487

### DIFF
--- a/pulp_smash/tests/platform/cli/test_pulp_manage_db.py
+++ b/pulp_smash/tests/platform/cli/test_pulp_manage_db.py
@@ -1,0 +1,124 @@
+# coding=utf-8
+"""Tests for the ``pulp-manage-db`` executable.
+
+``pulp-manage-db`` should only run when :data:`REQUIRED_SERVICES` are running
+and when :data:`CONFLICTING_SERVICES` are stopped. See:
+
+* `Pulp #2186 <https://pulp.plan.io/issues/2186>`_
+* `Pulp Smash #487 <https://github.com/PulpQE/pulp-smash/issues/487>`_
+"""
+import inspect
+import unittest
+
+from pulp_smash import cli, config, exceptions, selectors, utils
+
+REQUIRED_SERVICES = frozenset(('mongod',))
+"""If any of these services are stopped, ``pulp-manage-db`` will abort."""
+
+CONFLICTING_SERVICES = frozenset((
+    'pulp_celerybeat',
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Tested against Pulp 2.12 nightlies on F24, F25 and RHEL 7, and Pulp 2.13 nightlies on F24 and F25.

pulp_smash/tests/platform/cli/test_pulp_manage_db.py
```diff
@@ -0,0 +1,124 @@
...
+    utils.reset_pulp(config.get_config())
+
+
+class BaseTestCase(unittest.TestCase):
+    """An abstract base class for the test cases in this module."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Maybe skip this test case."""
+        if inspect.getmro(cls)[0] == BaseTestCase:
```
elyezer: We are adding base test cases classes as we increase the coverage. That makes me thing if we should create a base class for the base test case classes where the logic to skip will not be repeated over and over again.

Off course this is not required on this PR but is something that I wanted to start a conversation about.

pulp_smash/tests/platform/cli/test_pulp_manage_db.py
```diff
@@ -0,0 +1,124 @@
...
+        cls.cmd = () if utils.is_root(cls.cfg) else ('sudo',)
+        cls.cmd += (
+            'runuser', '--shell', '/bin/sh', '--command', 'pulp-manage-db',
+            '-', 'apache'
+        )
+
+    def tearDown(self):
+        """Start all of Pulp's services."""
+        cli.ServiceManager(config.get_config()).start(
+            CONFLICTING_SERVICES.union(REQUIRED_SERVICES)
```
elyezer: :heart:

pulp_smash/tests/platform/cli/test_pulp_manage_db.py
Ichimonji10: The reason I have this conditional is so that:

* child classes that don't define `setUpClass()` will execute as expected, and
* child classes that do define `setUpClass()` can call `super()`.

I'm curious what a different approach would look like.

PR #617, opened 2017-04-05
Maybe skip test affected by Pulp issue 1376

Ichimonji10:
See: https://pulp.plan.io/issues/1376


Ichimonji10:
Tested against Pulp 2.12 and 2.13 nightlies on F24, F25 and RHEL 7.

PR #619, opened 2017-04-05
Refactor test case for last_unit_added repo attr

Ichimonji10:
Do at least the following:

* Add an assertion to method `test_update_on_copy`. Assert that the
  `last_unit_added` repository attribute is non-null immediately after a
  content unit is associated with a repository.
* Move assertions in `test_update_on_copy` into `subTest` context
  managers.
* Fiddle with docstrings, comments, and variable names.


Ichimonji10:
Tested against Pulp 2.12 and 2.13 nightlies on F24, F25 and RHEL 7.

Ichimonji10:
Testing @SatelliteQE

PR #620, opened 2017-04-12
Implement new configuration file format

elyezer:
Improve the configuration in order to support clustered Pulp deployment
where multiple machines compose a Pulp system. But it still supporting
deployments where all services are running within the same machine.

The new format adopt the system with roles format where each system
(machine) have one or more roles. Each role provide information about a
given fact, for example, the api role tells Pulp Smash that the API can
be accessed on a given system. Also it helps Pulp Smash check which
systems have a Pulp service running, then Pulp Smash can access that
system and manage the service if that is needed.

### DIFF
--- a/pulp_smash/__main__.py
+++ b/pulp_smash/__main__.py
@@ -5,7 +5,7 @@ from os.path import join
 
 from xdg import BaseDirectory
 
-from pulp_smash.config import ServerConfig
+from pulp_smash.config import PulpSmashConfig
 
 MESSAGE = tuple((
     '''\
@@ -16,33 +16,48 @@ MESSAGE = tuple((
     '''\
     {
         "pulp": {
-            "base_url": "https://pulp.example.com",
             "auth": ["username", "password"],
-            "verify": true,
-            "version": "2.7.5",
-            "cli_transport": "local"
+            "version": "2.12.2"
         },
-        "pulp agent": {
-            "base_url": "https://pulp-agent.example.com"
-        }
+        "systems": [
+            {
+                "hostname": "pulp.example.com",
+                "roles": {
+                    "amqp broker": {"service": "qpidd"},
+                    "api": {
+                        "scheme": "https",
+                        "verify": true
+                    },
+                    "mongod": {},
+                    "pulp cli": {},
+                    "pulp celerybeat": {},
+                    "pulp resource manager": {},
+                    "pulp workers": {},
+                    "shell": {"transport": "local"},
+                    "squid": {}
+                }
+            }
+        ]
     }''',
     '''\
-    Each section provides information about a single Pulp-related service and
-    the host on which that service is installed. The "pulp" and "pulp agent"
-    sections tell about the Pulp and Pulp Agent services, respectively. The
-    former is required. The latter is optional, and if omitted, relevant tests
-    are skipped.
+    This configuration file format allows configuring either single machine
+    Pulp deployment (as seen in the example above) or clustered Pulp deployment
+    where each service can be installed on its own machine. The "pulp" section
+    lets you declare the version of and authentication credentials for the Pulp
+    deployment under test. The "systems" section provides information about
+    each system that composes the deployment. A system must have its hostname
+    and the roles that system has.
     ''',
     '''\
-    The `verify`, `version` and `cli_transport` keys are optional. The `verify`
-    option can be used to explicitly enable or disable SSL verification. The
-    `version` option can be used to explicitly run tests suitable for an older
-    version of Pulp. By default, Pulp Smash assumes that the Pulp server under
-    test is the latest development version. The `cli_transport` key can be used
-    to explicitly choose how to contact the Pulp server when executing shell
-    commands. This can be set to "local" or "ssh". If omitted, Pulp Smash
-    guesses which transport to use by comparing the hostname in the `base_url`
-    against the current system's hostname.
+    Not all roles requires additional information. Currently, only the amqp
+    broker, api and shell roles do. The amqp broker expects the service to be
+    defined and it can be qpidd or rabbitmq. The api role means that httpd will
+    be running on the system. The api's scheme allows specifying if the API
+    should be accessed using http or https, verify allows specifying if the
+    request SSL should be verified (true or false or a path to a certificate
+    file). The shell role configures how the system will be accessed by using a
+    local or ssh transport, only set local if Pulp Smash is running on that
+    same system.
     ''',
     '''\
     Notes:
@@ -70,7 +85,7 @@ MESSAGE = tuple((
 
 def main():
     """Provide usage instructions to the user."""
-    cfg = ServerConfig()
+    cfg = PulpSmashConfig()
     cfg_path = join(
         # pylint:disable=protected-access
         BaseDirectory.save_config_path(cfg._xdg_config_dir),
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -172,16 +172,27 @@ class Client(object):
     mechanisms. We've covered response handling mechanisms — let's move on to
     request handling mechanisms.
 
-    When a client is instantiated, a :class:`pulp_smash.config.ServerConfig`
+    When a client is instantiated, a :class:`pulp_smash.config.PulpSmashConfig`
     must be passed to the constructor, and configuration options are copied
-    from the ``ServerConfig`` to the client. These options can be overridden on
-    a per-object or per-request basis. Here's an example:
+    from the ``PulpSmashConfig`` to the client. These options can be overridden
+    on a per-object or per-request basis. Here's an example:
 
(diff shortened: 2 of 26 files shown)


pulp_smash/__main__.py
```diff
@@ -16,33 +16,45 @@
...
+            "version": "2.12.2"
         },
-        "pulp agent": {
-            "base_url": "https://pulp-agent.example.com"
+        "systems": {
+            "pulp.example.com": {
+                "amqp broker": {"service": "qpidd"},
+                "api": {
+                    "scheme": "https",
+                    "verify": true
```
elyezer: auth is a global option, because that it is inside the pulp section not for each system

Ichimonji10: /me facepalms

elyezer:
I am marking this PR as DONOTMERGE because some additional changes are needed and also this branch should be run on automation to make sure it won't break anything. The settings is a very sensible part of Pulp Smash and is better to test as much as possible.

elyezer:
This PR is ready for review. I've tested it on Pulp 2.13 on RHEL7 and it got the same green build current master branch got.

Ichimonji10:
I'll have a look today.

pulp_smash/config.py
```diff
@@ -279,72 +178,151 @@ def read(self, section=None, xdg_config_file=None, xdg_config_dir=None):
...
-        return cfg
-
-    def get_requests_kwargs(self):
+        :param role: The role to filter the available systems, see
+            `pulp_smash.config.ROLES` for more information.
+        """
+        return [
+            system for system in self.systems
+            if role in ROLES and role in system.roles
+        ]
```
Ichimonji10: I see the `if role in ROLES` here, and wonder if this would be better:

```python
if role not in ROLES:
    raise ValueError('The given role, {}, is not recognized. Valid roles are: {}'.format(role, ROLES))
return [system for system in self.systems if role in system.roles]
```

The current code will just mysteriously ignore certain `role` arguments, and the user will never be told why.

Alternately, we could just raise a warning if the given `role` isn't recognized, and drop the `if role in ROLES` check.

pulp_smash/config.py
```diff
@@ -279,72 +178,151 @@ def read(self, section=None, xdg_config_file=None, xdg_config_dir=None):
...
+        """Map old config base_url to the system with api role."""
+        api_system = self.get_systems('api')[0]
+        return '{}://{}/'.format(
+            api_system.roles['api']['scheme'], api_system.hostname)
+
+    def get_base_url(self, pulp_system=None):
+        """Generate the base URL for a given ``pulp_sytem``.
+
+        If ``pulp_system`` is ``None`` then the first system found with the
+        ``api`` role will be choosen.
```
Ichimonji10: `s/choosen/chosen/`

pulp_smash/__main__.py
```diff
@@ -16,33 +16,49 @@
...
-    former is required. The latter is optional, and if omitted, relevant tests
-    are skipped.
+    This configuration file format allows configuring either single machine
+    Pulp deployment (as seen in the example above) or clustered Pulp deployment
+    where each service can be installed on its own machine. The "pulp" section
+    lets you declare the version of and authentication credentials for the Pulp
+    deployment under test. The "systems" section provides information about
+    each system that composes the deployment, with one entry for each system,
+    where the key is the system's hostname and the value is a dictionary with
+    the roles the system has.
```
Ichimonji10: Please drop "where the key is...".

pulp_smash/api.py
```diff
@@ -192,8 +203,9 @@ class Client(object):
     >>> response = client.get('/index.html')  # Do default SSL verification
 
     Anything accepted by the `Requests`_ functions may be placed in
-    ``request_kwargs`` or passed in to a specific call. You can set ``verify``,
-    ``auth`` and more.
+    ``request_kwargs`` or passed in to a specific call. You can set ``verify``
+    for example. ``scheme``, on the other hand, will be removed since it is
+    there only to tell if the system should be accessed over HTTP or HTTPS.
```
Ichimonji10: If one writes `client.request_kwargs['scheme'] = ...`, then `client.request_kwargs['scheme']` will be updated as expected. Similarly, if one writes `client.get('/some/path', scheme='http')`, then that client's `request_kwargs` will be updated before the HTTP GET request is issued. In this context, I don't think it's accurate to say that `scheme, on the other hand, will be removed.`.

pulp_smash/cli.py
```diff
@@ -326,25 +335,34 @@ def start(self, services):
...
+        services = set(services)
+        for system in self._cfg.systems:
+            intersection = services.intersection(
+                self._cfg.services_for_roles(system.roles))
+            if intersection:
+                client = Client(self._cfg, pulp_system=system)
+                svc_mgr = self._get_service_manager(self._cfg, system)
+                if svc_mgr == 'sysv':
+                    with self._disable_selinux(client):
+                        return self._start_sysv(client, intersection)
```
Ichimonji10: I think the goal of this method is to start the named `services` on each system for which a relevant role is declared. Does it make sense to `return` here, and two lines below? Or should the `CompletedProcess` objects be aggregated into a dict, keyed by the hostname of each system?

pulp_smash/cli.py
```diff
@@ -353,25 +371,34 @@ def stop(self, services):
...
+        services = set(services)
+        for system in self._cfg.systems:
+            intersection = services.intersection(
+                self._cfg.services_for_roles(system.roles))
+            if intersection:
+                client = Client(self._cfg, pulp_system=system)
+                svc_mgr = self._get_service_manager(self._cfg, system)
+                if svc_mgr == 'sysv':
+                    with self._disable_selinux(client):
+                        return self._stop_sysv(client, intersection)
```
Ichimonji10: See the comment on `start`.

pulp_smash/config.py
elyezer: Yeah, much better.

pulp_smash/config.py
elyezer: ops, thanks

pulp_smash/cli.py
elyezer: Excellent observation. I was completely missing this.

I don't have a strong opinion on if we should work on a single or multiple pulp_system objects. I think the former may be more interesting since we can control which pulp_system we want to stop a service and then properly test the fail over features.

We also could have stop_all and start_all that would do stop and start on all systems available on the deployment.

I like where this conversation is going.

Ichimonji10: > I think the former may be more interesting since we can control which pulp_system we want to stop a service and then properly test the fail over features.

Great point. We need the ability to manage services on a single system.

elyezer: Actually instead of another method, I think adding the pulp_system argument like the other methods and functions will be better. So if pulp_system is None then run on all systems.

elyezer:
@Ichimonji10 this PR is ready for another round of review. Thank you for the comments and all the help.

pulp_smash/utils.py
```diff
@@ -108,14 +108,19 @@ def reset_pulp(server_config):
...
     client.run('mongo pulp_database --eval db.dropDatabase()'.split())
-    client.run((
-        prefix + 'runuser --shell /bin/sh apache --command pulp-manage-db'
-    ).split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/content').split())
-    client.run((prefix + 'rm -rf /var/lib/pulp/published').split())
+
+    for index, system in enumerate(server_config.get_systems('api')):
+        prefix = '' if is_root(server_config, pulp_system=system) else 'sudo '
+        if index == 0:
```
Ichimonji10: Is the intent to execute `pulp-manage-db` on the same system hosting mongod? If so, I don't  think this will work. There's no guarantee that `server_config.get_systems('mongod')[0]` and `server_config.get_systems('api')[0]` are the same system. Can we use the same `client` object to execute both `db.dropDatabase()` and `pulp-manage-db`?

elyezer: The idea is to run `pulp-manage-db` on an api system where the apache user is available and the command. On the other hand the mongod system should be unique on a deployment.

The code is doing what is supposed to do which is: drop the database on the system with mongo installed and run `pulp-manage-db` once and clean the /var/lib/pulp for every api system.

Ichimonji10: Ahh, right. Nevermind my comment. I mis-understood what should be done.

elyezer: No problem, thanks for pointing this out. It could have an issue.

pulp_smash/config.py
Ichimonji10: bump

Ichimonji10:
All comments have been addressed. I can run a subset of the test suite against my own systems, with both the old and new configuration file formats.

Ichimonji10:
:shipit:

PR #621, opened 2017-04-15
Make code comply with Pylint 1.7.0

Ichimonji10:
See: https://docs.pylint.org/en/latest/whatsnew/1.7.html

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -82,8 +82,7 @@ def _split_path(path):
     head, tail = os.path.split(path)
     if head == '':
         return (tail,)
-    else:
-        return _split_path(head) + (tail,)
+    return _split_path(head) + (tail,)
 
 
 def setUpModule():  # pylint:disable=invalid-name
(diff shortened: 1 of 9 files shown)


Ichimonji10:
I'm marking this as DO NOT MERGE because the changes need to be tested against some real, live Pulp systems. Having this PR open still helps me tracks the work, though, and gives teammates a chance to see what's coming down the pike.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -404,7 +403,7 @@ def test_predistributor_id(self):
                 repo = api_client.post(REPOSITORY_PATH, body)
         except AssertionError as err:
             self.addCleanup(api_client.delete, repo['_href'])
-            raise err
+            raise err  # pylint:disable=raising-bad-type
```
elyezer: I wonder if just calling raise would allow removing the pylint:disable

Ichimonji10: Great suggestion! This works perfectly.

Ichimonji10:
I've executed all of the modified test cases against VMs, and they pass. I'm reasonably confident in these changes.

PR #628, opened 2017-04-28
Add an ISO upload test case

Ichimonji10:
Add `UploadIsoTestCase`. This new test case uploads an ISO file to an
ISO repository.

The test case is simple and can (should!) be expanded, but even in its
current state, it adds some much-needed coverage for a very basic use
case.


Ichimonji10:
Tested against a matrix of six systems, where one matrix is F24, F25 and RHEL 7, and the other matrix is nightly builds of Pulp 2.13 and 2.14.

PR #629, opened 2017-05-02
Use the newest version of pydocstyle

Ichimonji10:
flake8-docstrings version 1.1.0 requires pydocstyle 2.0.0. Older
versions of flake8-docstrings require older versions of pydocstyle.
Un-pin the dependency on pydocstyle<2, so that the newest version of
flake8-docstrings and pydocstyle may both be used.

See: https://gitlab.com/pycqa/flake8-docstrings/blob/master/HISTORY.rst

Update the code base to be compatible with the newest version of
pydocstyle.


Ichimonji10:
This fixes a Travis failure.

PR #631, opened 2017-05-05
Add jsonschema dependency, test docker manifests

Ichimonji10:
Add a new dependency to `setup.py`: jsonschema. This library lets one
define schemas for JSON documents, and validate that arbitrary JSON
documents abide by those schemas.

Expand `SyncPublishV2TestCase`. Let the test case make HTTP GET requests
to `/v2/{repo_id}/manifests/latest` with varying headers, and use
jsonschema to assert that the response is as described by one of the
following:

* https://docs.docker.com/registry/spec/manifest-v2-1/
* https://docs.docker.com/registry/spec/manifest-v2-2/

Fiddle with `SyncPublishV1TestCase` so that it's structured like
`SyncPublishV2TestCase`.

See: https://python-jsonschema.readthedocs.io/en/latest/

### DIFF
--- a/pulp_smash/tests/docker/cli/test_sync_publish.py
+++ b/pulp_smash/tests/docker/cli/test_sync_publish.py
@@ -12,6 +12,101 @@ from pulp_smash.constants import (
 from pulp_smash.tests.docker.cli import utils as docker_utils
 from pulp_smash.tests.docker.utils import set_up_module
 
+from jsonschema import validate
+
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested against a matrix of six systems, where one axis is F24, F25 and RHEL 7, and the other axis is Pulp 2.13 and 2.14.

elyezer:
I will rebase the validate settings branch after this being merged.

PR #641, opened 2017-05-11
Add config validation

elyezer:
Validate Pulp Smash configuration file and ensure only expected values
and roles are present.

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -8,12 +8,14 @@ server. :class:`pulp_smash.config.PulpSmashConfig` eases the task of managing
 that information.
 """
 import collections
+import itertools
 import json
 import os
 import warnings
 from copy import deepcopy
 from urllib.parse import urlparse
 
+import jsonschema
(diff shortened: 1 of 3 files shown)


pulp_smash/config.py
```diff
@@ -8,12 +8,15 @@
...
 import collections
+import itertools
 import json
 import os
 import warnings
 from copy import deepcopy
 from urllib.parse import urlparse
 
+import jsonschema
+import jsonschema.validators as jsonschema_validators
```
Ichimonji10: Is there any special reason to use this alias? The following is valid:

```python
>>> import jsonschema
>>> jsonschema.validators.validator_for
<function validator_for at 0x7f3d88cffb70>
```

elyezer: That is stuff that remained from the first version. Before only jsonschema.validators was needed but as I wanted to have better error messages then I started importing the jsonchema module.

Thanks for noticing it, I've updated the PR and dropped that second import.

Issue #637, opened 2017-05-05
Test rsync publish


preethit:
https://pulp.plan.io/issues/2722
1. create repo with yum and rsync dist 
2. publish with rsync --> observe skipped publish result 
3. publish with rsync again and observe the traceback

Ichimonji10:
Among the issues in the 2.13.1 milestone, I think this one is a good candidate for easy automation: it doesn't require creating any new fixtures, and it doesn't require installing new tools on the Pulp systems.

PR #645, opened 2017-05-19
Test publishing an RPM rsync distributor twice

Ichimonji10:
Expand `PublishBeforeYumDistTestCase` so that, when executed, it will do
the following:

1. Create a repository with a yum distributor and rsync distributor.
2. Publish with the rpm rsync distributor. Verify that:

   * The publish has a result of "skipped."
   * No files are placed on the remote system.

3. Publish with the rpm rsync distributor again. Perform the same
   verification steps.

Step 3 is new.

Fix: https://github.com/PulpQE/pulp-smash/issues/637

See: https://pulp.plan.io/issues/2722

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -214,9 +214,18 @@ class PublishBeforeYumDistTestCase(
     Do the following:
 
     1. Create a repository with a yum distributor and rsync distributor.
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a matrix of systems where the axes are:

* Pulp 2.13 nightly, Pulp 2.14 nightly
* F24, F25, RHEL 7

I'm spinning up some additional VMs so that I can test against Pulp 2.13 beta.

Issue #615, opened 2017-03-31
Test Rsync publish for RPM repo run in fast forward mode


preethit:
https://pulp.plan.io/issues/2666

ipanova:
https://github.com/pulp/pulp/pull/2998#issuecomment-295252998

specifically steps 1, 6-11

elyezer:
Pasting the steps here:

```
1. create repo with 2 dist( yum and rsync)
2. add feed zoo
3. publish with rsync dist and check publish has skipped result( ensure you cannot publish if no publish happened with predist, also that no traceback is raised)
4. sync repo and publish with both dist
5. check 32 num of processed units with rsync dist
6. add 1 rpm
7. publish with both dist
8. check 1 num of processed units with rsync dist( check the fast forward functionality)
9. publish with rsync and check publish has skipped result 
10. publish with both dist
11. check 0 num of processed units with rsync dist( check the fast forward functionality)
12. remove 1 rpm
13. publish with both dist
14. check 32 items processed with rsync dist( ensure publish from scratch happened)
15. add 1 rpm
16. publish with both dist
17. check 1 num of processed units with rsync dist( check the fast forward functionality)
```

Ichimonji10:
Among the issues in the 2.13.1 milestone, I think this one is a good candidate for easy automation: it doesn't require creating any new fixtures, and it doesn't require installing new tools on the Pulp systems.

PR #646, opened 2017-05-19
Add test case for multiple RPM rsync publishes

Ichimonji10:
Add `PublishTwiceTestCase` to module
`pulp_smash.tests.rpm.api_v2.test_rsync_distributor`. From its
docstring:

    Do the following when executed:

    1. Create a yum repository with a yum and rsync distributor.
    2. Add some content to the repository.
    3. Publish with the yum and rsync distributor.
    4. Publish with the yum and rsync distributor again.

    The second publish with the rsync distributor should be a
    fast-forward publish, as no new units were added to the repository
    between the first and second publishes. Verify that the first rsync
    publish is not a fast-foward publish and that the second one is.

Fix: https://github.com/PulpQE/pulp-smash/issues/615

See: https://pulp.plan.io/issues/2666

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -150,6 +150,24 @@ class _RsyncDistUtilsMixin(object):  # pylint:disable=too-few-public-methods
         self.addCleanup(api_client.delete, repo['_href'])
         return api_client.get(repo['_href'], params={'details': True})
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a matrix of nine systems, where the axes are:

* Pulp 2.13 nightly, 2.13 beta, 2.14 nightly
* F24, F25, RHEL 7

The test is currently skipped on all systems. When the skipping logic is removed, the test fails in an expected fashion.

PR #648, opened 2017-05-22
Fix logic error in rpm rsync distributor test case

Ichimonji10:
`PublishTwiceTestCase` verifies that the rpm rsync distributor performs
a fast-forward publish when the same content is published two or more
times. The logic used to count the number of units rsync'd to the
destination is incorrect. Fix it.

See: 0b16e140a9120af404de8dfb82820374b27c9ad4

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
+++ b/pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
@@ -710,9 +710,9 @@ class PublishTwiceTestCase(
                 utils.publish_repo(cfg, repo, {'id': dists[dist]['id']}).json()
             )
             publish_task = self.get_publish_task(cfg, report)
-        total_num_processed = self.get_total_num_processed(publish_task)
+        num_processed = self.get_num_processed(publish_task)
         with self.subTest(comment='first rsync publish'):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a matrix of nine systems, where the axes are:

    Pulp 2.13 nightly, 2.13 beta, 2.14 nightly
    F24, F25, RHEL 7

The test succeeds against all systems.

pulp_smash/tests/rpm/api_v2/test_rsync_distributor.py
```diff
@@ -710,23 +710,24 @@ def test_all(self):
...
         with self.subTest(comment='second rsync publish'):
-            self.assertEqual(total_num_processed, 0, publish_task)
+            self.assertEqual(num_processed, 0, publish_task)
 
     @staticmethod
-    def get_total_num_processed(publish_task):
-        """Return total ``num_processed`` for each step in ``publish_task``."""
+    def get_num_processed(publish_task):
+        """Return ``num_processed`` for unit query step in ``publish_task``."""
         return sum(
```
Ichimonji10: `sum` should gracefully deal with the case where no matching steps are found.

```python
>>> sum((1, 2, 3))
6
>>> sum(())
0
```

Issue #638, opened 2017-05-05
Test that repos with same errata can be synced in parallel


preethit:
https://pulp.plan.io/issues/2721

Add test to that sync is successful for all repos with same , when syncing in parallel.

Ichimonji10:
Among the issues in the 2.13.1 milestone, I think this one is a good candidate for easy automation: it doesn't require creating any new fixtures, and it doesn't require installing new tools on the Pulp systems.

PR #651, opened 2017-05-23
Add test case for parallel errata synchronization

Ichimonji10:
Pulp issue 2721 describes how a race condition can arise when multiple
repositories with identical errata are synchronized at the same time.
This issue should be fixed in Pulp 2.13.1. Add a test case for this
issue.

See: https://pulp.plan.io/issues/2721

Fix: https://github.com/PulpQE/pulp-smash/issues/638

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_sync_publish.py
@@ -11,6 +11,7 @@ For information on repository sync and publish operations, see
 """
 import inspect
 import unittest
+from threading import Thread
 from urllib.parse import urljoin
 
 from packaging.version import Version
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested against a matrix of nine VMs, where the axes are:

* Pulp 2.13 beta, 2.13 nightly, 2.14 nightly

pulp_smash/tests/rpm/api_v2/test_sync_publish.py
```diff
@@ -333,3 +335,76 @@ def get_feed(self, repo):
...
+
+            :param repo: A dict of information about a repository.
+            """
+            repos.append(client.get(repo['_href']))
+
+        threads = tuple(Thread(target=create_repo) for _ in range(5))
+        for thread in threads:
+            thread.start()
+        for thread in threads:
+            thread.join()
```
Ichimonji10: `join()` doesn't return the called function's return value. Instead, it returns None. In fact, AFAIK, the `Thread` class makes no provision for returning values at all. That's why I make calls to `repos.append()`.

elyezer: I don't know either, what I know is that multiprocessing is able to return values so maybe using it can make the code shorter, but not that shorter so I think using the current approach is fine.

Ichimonji10: I'll look into that module. I need to lear more about Python's support for parallel programming in general.

PR #658, opened 2017-05-30
Fiddle with `pulp-smash settings create` wording

Ichimonji10:
Tell the user which message broker will be configured if the default of
Qpid is not chosen. (It's RabbitMQ.)


Ichimonji10:
Minor nitpick that I missed during the initial review.

PR #659, opened 2017-05-30
Configure secure HTTPS by default

Ichimonji10:
`pulp-smash settings create` can configure the settings file so that
connections to Pulp's API are performed over either HTTP or HTTPS. In
the latter case, Pulp will ask whether the validity of HTTPS connections
should be verified. The default answer is "no." Change the default
answer to "yes."

It's common for development environments to have insecure certificates.
In that case, it makes sense for the validity of HTTPS connections to be
ignored. However, there's value in making the user explicitly
acknowledge that they're doing something insecure, and ensuring that
they understand when something dangerous is happening. Enabling insecure
HTTPS connections is as easy as pressing "n" when answering prompts, so
this isn't a terribly onerous change.


Ichimonji10:
This is the last of the things I forgot to mention during the PR review, I swear. :wink:

PR #660, opened 2017-05-30
Fix issues with docker sync publish tests

Ichimonji10:
Several Docker CLI tests issue HTTP GET requests to:

* /crane/repositories
* /crane/repositories/v1
* /crane/repositories/v2

The tests issue these requests to verify that synced and published
repositories contain appropriate. However, according to Pulp issue 2723,
`/crane/repositories/v2` returns incorrect messages, and the fix is
different than originally anticipated by the test. Update the tests that
fetch these paths. See: https://pulp.plan.io/issues/2723

Several tests get `/crane/repositories` when they should be getting
`/crane/repositories/v2`. Fix those paths.

`SyncPublishV1TestCase` inappropriately deletes the repository it
creates. Fix the clean up procedure.

### DIFF
--- a/pulp_smash/tests/docker/cli/test_sync_publish.py
+++ b/pulp_smash/tests/docker/cli/test_sync_publish.py
@@ -3,6 +3,8 @@
 import unittest
 from urllib.parse import urlsplit, urlunsplit
 
+from packaging.version import Version
+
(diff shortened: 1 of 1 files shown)


Ichimonji10:

* RHEL 7, F24, F25
* Pulp 2.13 nightly and 2.14 nightly

Issue #662, opened 2017-06-01
Update the retain old count test to include Issue#2785


preethit:
https://pulp.plan.io/issues/2785
Check test_retain_old_count test and update if needed
pulp_smash.tests.rpm.api_v2.test_retain_old_count

Ichimonji10:
#666 and #668 verify that `--retain-old-count` works correctly when set to zero. To fully close out this issue, we need tests to verify that `--retain-old-count` works when set to a non-zero value.

PR #666, opened 2017-06-01
Refactor API test for retain_old_count sync option

Ichimonji10:
Make the test shorter and more robust in the face of unexpected
failures.


Ichimonji10:

* F24, F25, RHEL 7
* Pulp 2.13 and 2.14 nightly.

Fixing https://github.com/PulpQE/pulp-smash/issues/662 requires that a CLI test for the `retain_old_count` option be written. In the process of researching how to do that, I read through the corresponding API test in this module, and felt compelled to update it.

PR #668, opened 2017-06-02
Test --retain-old-count option via CLI

Ichimonji10:
Add a test for pulp-admin's `--retain-old-count` flag. Specifically,
ensure that `--retain-old-count 0` works correctly. This new test
mirrors the API test in
`pulp_smash.tests.rpm.api_v2.test_retain_old_count`. Neither of these
tests verify that `--retain-old-count` works with non-zero values.

Related: https://github.com/PulpQE/pulp-smash/issues/662

See: https://pulp.plan.io/issues/2785

### DIFF
--- a/pulp_smash/tests/rpm/cli/test_retain_old_count.py
+++ b/pulp_smash/tests/rpm/cli/test_retain_old_count.py
@@ -1,0 +1,85 @@
+# coding=utf-8
+"""Test the `retain_old_count`_ feature.
+
+When more than one version of an RPM is present in a repository and Pulp syncs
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/rpm/cli/test_retain_old_count.py
```diff
@@ -0,0 +1,85 @@
+# coding utf-8
```
elyezer: missing either a : or = after coding

Ichimonji10: oops!

PR #670, opened 2017-06-05
Expand CLI test case for `retain_old_count`

Ichimonji10:
The CLI test case for the `retain_old_count` feature already ensures
that the flag works correctly when set to zero. However, it doesn't
ensure that the flag works for values greater than zero. Fix this
omission.

Fix: https://github.com/PulpQE/pulp-smash/issues/662

See: https://pulp.plan.io/issues/2785


Ichimonji10:

* F24, F25 and RHEL 7.

PR #671, opened 2017-06-06
Update installation document

Ichimonji10:
`pyvenv` is deprecated in favor of `python -m venv`. Update the docs
accordingly.

`pip install ...` offers some advantages over `python setup.py install`.
Update the docs accordingly. See: https://stackoverflow.com/a/15731459

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -8,10 +8,10 @@ and activate a virtual environment:
 
 .. code-block:: sh
 
-    pyvenv env
+    python3 -m venv env
     source env/bin/activate  # run `deactivate` to exit environment
 
-To install Pulp Smash from `PyPi`_:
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -23,26 +23,27 @@ To install Pulp Smash from source (`GitHub`_):
...
-files are reflected in the working environment:
+Pulp Smash can also be installed from source in "develop mode," where changes to
+source code are reflected in the working environment:
 
 .. code-block:: sh
 
     git clone https://github.com/PulpQE/pulp-smash.git
     cd pulp-smash
     pip install -r requirements.txt -r requirements-dev.txt
+    pip install -e .
```
elyezer: `pip install -r requirements.txt` already does `-e .` do we still want this extra line?

Ichimonji10: Ahhh, no, we don't. Thanks!

PR #672, opened 2017-06-06
Skip tests affected by Pulp issue 2798

Ichimonji10:
Pulp issue 2798 describes how background and on demand downloads are
broken. See: https://pulp.plan.io/issues/2798


Ichimonji10:

* F24, F25 and RHEL 7.

Ichimonji10:
I'm going to go ahead and merge this. Doing so will fix a significant number of test failures. If I've messed up, then we're just back where we started - no harm done.

PR #673, opened 2017-06-07
Update method `utils.sync_repo`

Ichimonji10:
The `sync_repo` method has been around for a long time. A typical usage
is as follows, where `repo_href` is a simple string path:

```python
utils.sync_repo(cfg, repo_href)
```

Update the method so that it accepts a dict of repository information
instead of a simple string path. A typical usage is as follows, where
`repo` is a dict like `{'_href': repo_href}`:

```python
utils.sync_repo(cfg, repo)
```

This change has two advantages. First, Pulp's API returns information
about repositories as a dict. Letting `sync_repo` accept a dict means
that it can accept the data returned by Pulp's API with no munging.
Second, this change makes `sync_repo` more consistent with other modern
methods like `publish_repo`. Both of these advantages are well explained
by an example. Consider this legacy code:

```python
repo = api_client.post(REPOSITORY_PATH, body)
utils.sync_repo(cfg, repo['_href'])
utils.publish_repo(cfg, repo)
```

The code above can now be written as follows:

```python
repo = api_client.post(REPOSITORY_PATH, body)
utils.sync_repo(cfg, repo)
utils.publish_repo(cfg, repo)
```

### DIFF
--- a/pulp_smash/tests/docker/api_v2/test_tags.py
+++ b/pulp_smash/tests/docker/api_v2/test_tags.py
@@ -80,7 +80,7 @@ class DockerTagTestCase(utils.BaseAPITestCase):
         super().setUp()
         self.repo = create_docker_repo(self.cfg, DOCKER_UPSTREAM_NAME)
         self.addCleanup(api.Client(self.cfg).delete, self.repo['_href'])
-        utils.sync_repo(self.cfg, self.repo['_href'])
+        utils.sync_repo(self.cfg, self.repo)
         self.repo = api.Client(self.cfg, api.json_handler).get(
             self.repo['_href'], params={'details': True})
         self.tags = self._get_tags()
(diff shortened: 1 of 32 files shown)


Ichimonji10:
Partially tested against a matrix off systems, where the axes are:

* Fedora 24, Fedora 25, and RHEL 7.

Executing all affected tests would have taken quite a while, so I ran the following:

* python -m unittest pulp_smash.tests.ostree.api_v2.test_sync
* python -m unittest pulp_smash.tests.puppet.api_v2.test_sync_publish
* python -m unittest pulp_smash.tests.python.api_v2.test_sync_publish
* python -m unittest pulp_smash.tests.rpm.api_v2.test_signatures_saved_for_packages
* python -m unittest pulp_smash.tests.rpm.api_v2.test_sync_publish

PR #677, opened 2017-06-09
Add content applicability test case

Ichimonji10:
Add module `pulp_smash.tests.rpm.api_v2.test_content_applicability`.
Give this module a test case that performs several variations of the
following:

1. Create a repository and consumer.
2. Bind the consumer to the repository.
3. Create a consumer profile that includes several RPMs in the
   repository.
4. Regenerate applicability for the consumer.
5. Get an applicability report for the consumer, and verify that the
   correct consumers and RPMs are listed in the report.

Fix: https://github.com/PulpQE/pulp-smash/issues/676

PR #678, opened 2017-06-09
Rename constant

Ichimonji10:
Let the constant's name reflect the path it represents.


Ichimonji10:
I'm mostly relying on the linters to make sure this change is correct. I also re-ran the tests added by https://github.com/PulpQE/pulp-smash/pull/677.

Issue #643, opened 2017-05-11
Testing headers option for alternate content source


werwty:
https://pulp.plan.io/issues/1282

1. Create alternate content source configuration
`vi /etc/pulp/content/sources/conf.d/test.conf`
```
[content-world]
enabled: 1
priority: 0
expires: 3d
name: Content World
type: yum
base_url: http://bizhang.fedorapeople.org/
paths: zoo/ \
       test
max_concurrent: 10
max_speed: 1000
headers: X-RHUI-ID=1234 \
               X-Csrf-Token=31415926535897
```

2. Validate that headers shows up when `pulp-admin content sources list` is ran
3. run `pulp-admin content sources refresh`, monitoring the network traffic to check the HTTP headers.
I'm using tcpflow for this: `sudo tcpflow -p -c -i eth0 port 80`

4. Update test.conf with invalid headers
```
headers: X-RHUI-ID1234
```

5. run  `pulp-admin content sources refresh`  and make sure `Unparsable` is raised in the systemd logs
Note that the cli will show Task Succeeded despite nothing being downloaded. From what I can see this is because the task is running with no content sources, since the alternate content source is designed to fail over to the original source.

Ichimonji10:
I've made a shell script that performs all of the necessary steps to reproduce this issue. I am *not* able to find any of the given headers, such as `X-RHUI-ID`. Can you give any insight into this, @werwty ?

```bash
#!/usr/bin/env bash
# coding=utf-8
#
# Run this script as root on a Pulp host. When executed, this script will verify
# https://pulp.plan.io/issues/1282.
set -euo pipefail

readonly path=/etc/pulp/content/sources/conf.d/test.conf
readonly cfg="[test-content]
enabled: 1
priority: 0
expires: 1d
name: Test Content
type: yum
base_url: http://repos.fedorapeople.org/pulp/pulp/fixtures/
paths: rpm-unsigned/
max_concurrent: 10
max_speed: 1000"

login() (
    pulp-admin login -u admin -p admin
)

# Create and use a valid configuration file. Verify that:
#
# * The word "unparseable" doesn't show up in journalctl's output.
# * The given headers show up in the output of `pulp-admin content sources
#   list`.
test_valid_headers() (
    readonly local out_dir=valid_headers
    echo "$cfg" | cat >"$path"
    cat >>"$path" <<EOF
headers: X-RHUI-ID=1234 X-CSRF-TOKEN=31415926535897
EOF
[...]

PR #687, opened 2017-06-23
Test content source files with headers

Ichimonji10:
Add module `pulp_smash.tests.rpm.api_v2.test_content_sources`. This
module contains tests that create, list, refresh and delete content
sources. The tests are especially concerned with what happens when valid
or invalid HTTP headers are listed in the content source files.

This new module obsoletes module
`pulp_smash.tests.platform.cli.test_content_sources`. Drop it.

See: https://github.com/PulpQE/pulp-smash/issues/643


Ichimonji10:
@werwty, can you please look over these new test cases and verify that they're making correct assertions? You don't need to look through the exact code too deeply. It would be helpful enough if you inspected the docstrings on the new test methods.

@kersommoura or @kdelee, please review this PR.

Ichimonji10:

* Pulp 2.13 and 2.14 nightly

pulp_smash/tests/rpm/api_v2/test_content_sources.py
```diff
@@ -0,0 +1,167 @@
...
+        source_headers = self.cs_kwargs['headers']
+        self.assertEqual(source_headers, sources_by_id[source_id]['headers'])
+
+    def test_refresh_content_sources(self):
+        """Refresh Pulp's content sources.
+
+        Assert that the refresh completes successfully.
+        """
+        client = api.Client(self.cfg)
+        client.post('/pulp/api/v2/content/sources/action/refresh/')
```
elyezer: does this post request returns something, if yes what if we add a explicit assertion?

pulp_smash/tests/rpm/api_v2/test_content_sources.py
```diff
@@ -0,0 +1,167 @@
...
+        self.assertNotIn(source_id, sources_by_id)
+
+    def test_refresh_content_sources(self):
+        """Refresh Pulp's content sources.
+
+        Assert that the refresh completes successfully. It should complete
+        successfully because the broken content source is completely skipped.
+        """
+        client = api.Client(self.cfg)
+        client.post('/pulp/api/v2/content/sources/action/refresh/')
```
elyezer: Can a explicit assertion be added?

pulp_smash/tests/rpm/utils.py
```diff
@@ -76,6 +76,17 @@ def check_issue_2798(cfg):
...
+    """Return true if `Pulp #1282`_ affects the targeted Pulp system.
+
+    :param pulp_smash.config.PulpSmashConfig cfg: The Pulp system under test.
+
+    .. _Pulp #1282: https://pulp.plan.io/issues/1282
+    """
+    return (cfg.version < Version('2.14') or
+            selectors.bug_is_untestable(1282, cfg.version))
+
+
```
kdelee: More for my edification than anything, why do you sometimes define a "check_issue" function and sometimes not? 
Is it simply because this has two conditions where either might cause the bugfix to be untestable (because it doesn't yet exist)?

pulp_smash/tests/rpm/api_v2/test_content_sources.py
```diff
@@ -0,0 +1,167 @@
...
+
+class InvalidHeadersTestCase(unittest.TestCase):
+    """Define a content source with invalid headers.
+
+    See: `Pulp Smash #643 <https://github.com/PulpQE/pulp-smash/issues/643>`_.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create a content source with valid headers."""
```
kdelee: Did you mean to say 
```python
"""Create a content source with INVALID headers.""" # emphasis mine
```

pulp_smash/tests/rpm/api_v2/test_content_sources.py
```diff
@@ -0,0 +1,167 @@
...
+        cls.cfg = config.get_config()
+        if check_issue_1282(cls.cfg):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/1282')
+        cls.cs_kwargs = {
+            'source_id': utils.uuid4(),
+            'headers': ' '.join(header + utils.uuid4() for header in _HEADERS),
```
kdelee: So, what makes these headers "invalid" is that they have the header and the utils.uuid4() just concatenated together instead of having a '=' between them?

Ichimonji10: Correct.

pulp_smash/tests/rpm/api_v2/test_content_sources.py
```diff
@@ -0,0 +1,167 @@
...
+        cls.cs_path = _gen_content_source(cls.cfg, cs_body)
+
+    def test_list_content_sources(self):
+        """List Pulp's content sources.
+
+        Assert that:
+
+        1. The list includes an entry corresponding to the newly defined
+           content source configuration file.
+        2. This liste entry includes the headers defined in the configuration
```
werwty: s/liste/list/

pulp_smash/tests/rpm/api_v2/test_content_sources.py
Ichimonji10: This API call triggers an asynchronous task in Pulp, and Pulp returns an HTTP 202. When the response is received, the client's `response_handler` does two things:

1. Assert that the response's status code is less than 400.
2. Call a function for handling HTTP 202 responses.

The HTTP 202 response handling function then does the following:

1. Check the response's content type.
2. Gets the response's body and decodes it as JSON.
3. Calls a function for walking the tree of tasks that's listed in the JSON-decoded body.

The task-walking function then does some stuff, including checking each response's JSON-decoded body for error messages...

...and so on. What we care about here is that the HTTP 202 handling logic in the [`safe_handler`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.api.html#pulp_smash.api.safe_handler) response handler completes without raising any exceptions. In a sense, we already making an assertion: we're asserting that no exceptions are raised.

pulp_smash/tests/rpm/api_v2/test_content_sources.py
Ichimonji10: Yes. Thank you.

pulp_smash/tests/rpm/api_v2/test_content_sources.py
Ichimonji10: I think adding an explicit assertion is a bad idea. See above.

pulp_smash/tests/rpm/utils.py
Ichimonji10: Right. Using this function lets the bug-checking logic in the actual test stay short. That's the only reason.

pulp_smash/tests/rpm/api_v2/test_content_sources.py
Ichimonji10: :scream:

Issue #611, opened 2017-03-30
Verify Scheduled RPM installation and uninstallation work


bmbouter:
[Issue 2680](https://pulp.plan.io/issues/2680) tracks an issue where scheduling RPM installation and uninstallation does not work. To verify this fix do the following:

1. Set up a blank Pulp 2.12 environment with at least one consumer (this assumes the consumer is called `myconsumer`).

2. Create an RPM repository (I used the Fedora "zoo" demo repo). This was done with:

`pulp-admin rpm repo create --repo-id=zoo --relative-url=zoo --feed=http://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/`.

3. Sync and publish that repo with:

`pulp-admin rpm repo sync run --repo-id zoo`
`pulp-admin rpm repo publish run --repo-id zoo`

4. Bind it to the consumer.  `pulp-admin rpm consumer bind --repo-id zoo --consumer-id myconsumer`

5. Schedule the installation of an RPM package. The command below was used, but you need to adjust the date so it targets a point in time that's just a few minutes in the future

`pulp-admin rpm consumer package install schedules create --schedule R1/2017-03-29T14:20:00Z/PT1H --consumer-id myconsumer --name tiger`

6. Wait until the schedule triggers

7. Check the results

Ichimonji10:
After a review with @kersommoura today, I think it's fair to conclude that this test can't be implemented in a reliable way. There's just too many bugs in Pulp. If you're curious where we left off, see https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:pr-770

I'll file bugs and post them on https://github.com/PulpQE/pulp-smash/issues/611 as soon as I get around too it.

Ichimonji10:
See:

* [Pulp #3039](https://pulp.plan.io/issues/3039)
* [Pulp #3040](https://pulp.plan.io/issues/3040)
* [Pulp #3041](https://pulp.plan.io/issues/3041)
* [Pulp #3042](https://pulp.plan.io/issues/3042)

PR #692, opened 2017-06-27
Add test of scheduled install on pulp-consumer

kdelee:
NOTE: Packages necessary for pulp-consumer to function must be installed
for this test run. 

I will be filing an issue in [pulp/pulp_packaging](https://github.com/pulp/pulp_packaging) and linking it here momentarily.

closes issue #611

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -41,6 +41,7 @@ REQUIRED_ROLES = {
 OPTIONAL_ROLES = {
     'pulp cli',
     'squid',
+    'pulp consumer',
 }
 """Set of roles that can be present or not on a Pulp deployment."""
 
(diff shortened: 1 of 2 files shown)


kdelee:
See pulp_packaging issue [#2845](https://pulp.plan.io/issues/2845) for story of what Jenkins user would need to get tests like this one running

Ichimonji10:
This PR is temporarily being set aside. A proper solution will be nasty and gnarly. It can be done, but I think it's appropriate to come back to it with a little more experience in hand.

kdelee:
@Ichimonji10 
Sounds good.

Ichimonji10:
Well, unless you're really raring to keep on working at this one. I won't stop you. :wink: Take the above as a suggestion, not a command.

kdelee:
I think I'm going to work on [#2845](https://pulp.plan.io/issues/2845), which is a blocker for this, before I get back to refactoring this code.

PR #696, opened 2017-06-29
add test to test last_unit_removed does not change when no units are …

kersommoura:
…removed

closes#601

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_unassociate.py
+++ b/pulp_smash/tests/rpm/api_v2/test_unassociate.py
@@ -94,6 +94,32 @@ class RemoveUnitsTestCase(unittest.TestCase):
         }
         self.assertEqual(removed_ids & remaining_ids, set())
 
+    def test_03_last_unit_removed(self):
+        """Assert that ``last_unit_removed`` timestamp was not updated.
+
+        Whenever disassociate_units is called ``last_unit_removed`` should not
+        be updated even if that call does not remove units.
+
+        Do the following:
+
+        1. Note the repository's ``last_unit_removed`` field.
+        2. Sleep for at least one second.
+        3. Attempt to remove a non defined ``unit_id_name`` .
+        4. Note the repository's ``last_unit_removed`` field.
+        """
+        lur_before = self.get_repo_last_unit_removed()
+        time.sleep(1)  # ensure last_unit_removed increments
+
+        # Select an unit and mess it up
+        unit = random.choice(self.initial_units).copy()
+        unit_id_name, _ = _get_unit_id(unit)
+        unit['metadata'][unit_id_name] = utils.uuid4()
+
+        # Remove the non-existent unit
+        _remove_unit(self.cfg, self.repo, unit)
+        lur_after = self.get_repo_last_unit_removed()
+        self.assertEqual(lur_before, lur_after)
+
     def do_remove_unit(self, type_id):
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/rpm/api_v2/test_unassociate.py
```diff
@@ -94,6 +94,50 @@ def test_02_remaining_units(self):
...
+    def test_03_last_unit_removed(self):
+        """Assert that `last_unit_removed` timestamp was not updated.
+
+        Whenever disassociate_units is called `last_unit_removed` should not
+        be updated if that cal does not remove units.
+
+        Each verification is wrapped in a subTest.
+        See :meth:`do_verify_last_unit_timestamp`.
+        """
+        type_ids = ['erratum', 'package_category', 'package_group', 'rpm']
```
Ichimonji10: I appreciate the thoroughness, but I think this is overkill. If anything, it would be more useful to try removing a pair of made-up units where one has a bogus ID and the other has a bogus type_id. Please simplify this function and merge `do_verify_last_unit_timestamp` into this.

pulp_smash/tests/rpm/api_v2/test_unassociate.py
```diff
@@ -94,6 +94,50 @@ def test_02_remaining_units(self):
...
+
+        Timestamps``last_unit_removed``,  before and after the attempt
+        to unassociate an unit was not updated.
+        """
+        lur_before = self.get_repo_last_unit_removed()
+
+        time.sleep(1)  # ensure last_unit_removed increments
+        unit = random.choice(_get_units_by_type(self.initial_units, type_id))
+        self.removed_units.append(unit)
+        _unassociate_not_defined_unit(self.cfg, self.repo, unit)
```
Ichimonji10: While defining `_unassociate_not_defined_unit` and `_get_not_defined_unit` works, it's a very verbose way of doing things. Functions are great for encapsulating code that is re-used, or that are especially logically distinct. That's not the case here. This is a simpler way to do things:

```python
unit = copy.copy(random.choice(self.initial_units))  # don't change original object
unit[_get_unit_id(unit)] = utils.uuid4()
# remove unit from repository
# capture lur_after, and assert lur_before == lur_after
```

Bam. Three lines of code instead of two functions.

Ichimonji10:
Looks pretty good. Please squash commits.

```bash
git checkout fix_issue601
git rebase -i HEAD~2 # "squash" second commit into first
```

pulp_smash/tests/rpm/api_v2/test_unassociate.py
```diff
@@ -94,6 +94,32 @@ def test_02_remaining_units(self):
         }
         self.assertEqual(removed_ids & remaining_ids, set())
 
+    def test_03_last_unit_removed(self):
+        """Assert that ``last_unit_removed`` timestamp was not updated.
+
+        Whenever disassociate_units is called ``last_unit_removed`` should not
+        be updated even if that call does not remove units.
```
Ichimonji10: This comment isn't quite right. Whenever `disassociate_units` is called, `last_unit_removed` *should* be updated. The only exception is when no units are removed. This could read:

> If `disassociate_units` is called and no units are removed, `last_unit_removed` should not update.

I'll update this in a follow-up commit if this is left as is, so don't stress out about it.

Issue #694, opened 2017-06-29
Test Tasks stuck in waiting after restart of pulp services issue#2835


preethit:
https://pulp.plan.io/issues/2835

preethit:
@Ichimonji10 is this something that can be worked on soon?

Ichimonji10:
Sure.

Ichimonji10:
Upon looking into it, the developers have found that there are three issues at play:

- [ ] [Pulp #2861](https://pulp.plan.io/issues/2861) (Issue A)
- [ ] [Pulp #2835](https://pulp.plan.io/issues/2835) (Issue B)
- [ ] [Pulp #2837](https://pulp.plan.io/issues/2837) (Issue C)

PR #699, opened 2017-06-30
Test failing Pulp workers

Ichimonji10:
Add a new module, `pulp_smash.tests.rpm.api_v2.test_service_resiliency`.
This module should contain tests that check Pulp's ability to continue
working in the face of service failures. Initially, it contains only
`MissingWorkersTestCase`. When executed, it will:

> 1. Ensure there is only one Pulp worker.
> 2. Create a repository. Let its feed reference a large repository.
>    (For example, EPEL.)
> 3. Start a sync. Immediately restart the ``pulp_workers`` service.
>    (It's important that ``pulp_workers`` be restarted, not started and
>    stopped.  For details, see `Pulp #2835`_.) This should cause the
>    first sync to abort.
> 4. Update the repository. Let its feed reference a small repository.
>    (For example, :data:`pulp_smash.constants.RPM_UNSIGNED_FEED_URL`.)
> 5. Start a sync. Verify that it completes. If `Pulp #2835`_ still
>    affects Pulp, then the worker will be broken, and the sync will
>    never start.

Add a new function, `pulp_smash.api.echo_handler`. Add unit tests for
it.

Fix: https://github.com/PulpQE/pulp-smash/issues/694

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -89,6 +89,19 @@ def echo_handler(server_config, response):  # pylint:disable=unused-argument
     return response
 
 
+def code_handler(server_config, response):  # pylint:disable=unused-argument
+    """Check the response status code, and return the response.
+
+    Unlike :meth:`safe_handler`, this method doesn't wait for asynchronous
+    tasks to complete if ``response`` has an HTTP 202 status code.
+
+    :raises: ``requests.exceptions.HTTPError`` if the response status code is
+        in the 4XX or 5XX range.
+    """
+    response.raise_for_status()
+    return response
+
+
 def safe_handler(server_config, response):
     """Check status code, wait for tasks to complete, and check tasks.
 
--- a/pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
+++ b/pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
@@ -1,0 +1,86 @@
+# coding=utf-8
+"""Test the resiliency of Pulp's services.
+
+Pulp is designed to be resilient. It should be possible for a service to
+disappear and reappear, with the only consequence being slower processing of
+tasks. This module has tests for Pulp's resilience in the face of such issues.
+"""
+import unittest
+from urllib.parse import urljoin
+
+from pulp_smash import api, cli, config, selectors, utils
+from pulp_smash.constants import (
+    PULP_SERVICES,
+    REPOSITORY_PATH,
+    RPM_MIRRORLIST_LARGE,
+    RPM_UNSIGNED_FEED_URL,
+)
+from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
+
+_PULP_WORKERS_CFG = '/etc/default/pulp_workers'
+
+
+class MissingWorkersTestCase(unittest.TestCase):
+    """Test that Pulp deals well with missing workers.
+
+    When executed, this test case will do the following:
+
+    1. Ensure there is only one Pulp worker.
+    2. Create a repository. Let its feed reference a large repository. (For
+       example, EPEL.)
+    3. Start a sync. Immediately restart the ``pulp_workers`` service. (It's
+       important that ``pulp_workers`` be restarted, not started and stopped.
+       For details, see `Pulp #2835`_.) This should cause the first sync to
+       abort.
+    4. Update the repository. Let its feed reference a small repository. (For
+       example, :data:`pulp_smash.constants.RPM_UNSIGNED_FEED_URL`.)
+    5. Start a sync. Verify that it completes. If `Pulp #2835`_ still affects
+       Pulp, then the worker will be broken, and the sync will never start.
+
+    .. _Pulp #2835: https://pulp.plan.io/issues/2835
+    """
+
+    def setUp(self):
+        """Ensure there is only one Pulp worker."""
+        self.cfg = config.get_config()
+        if selectors.bug_is_untestable(2835, self.cfg.version):
+            self.skipTest('https://pulp.plan.io/issues/2835')
+        sudo = '' if utils.is_root(self.cfg) else 'sudo'
+        cli.Client(self.cfg).machine.session().run(
+            "{} bash -c 'echo PULP_CONCURRENCY=1 >> {}'"
+            .format(sudo, _PULP_WORKERS_CFG)
+        )
+        cli.GlobalServiceManager(self.cfg).restart(PULP_SERVICES)
+
+    def test_all(self):
+        """Test that Pulp deals well with missing workers."""
+        # Create a repository. No repository addCleanup is necessary, because
+        # Pulp will be reset after this test.
+        client = api.Client(self.cfg, api.json_handler)
+        body = gen_repo()
+        body['importer_config']['feed_url'] = RPM_MIRRORLIST_LARGE
+        body['distributors'] = [gen_distributor()]
+        repo = client.post(REPOSITORY_PATH, body)
+        repo = client.get(repo['_href'], params={'details': True})
+
+        # Start syncing the repository and restart pulp_workers.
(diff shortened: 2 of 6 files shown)


Ichimonji10:
@bmbouter @dkliban @dralley Can one of you please look over this test and make sure it's correctly testing [Pulp #2835](https://pulp.plan.io/issues/2835), part B?

@kersommoura @kdelee Can one of you please review? It doesn't need to be in depth. A sanity check from another QE would be helpful.

* Pulp 2.13 and 2.14 nightly

When the `self.skipTest` call is removed, this test causes several tasks to enter an infinite waiting step. That's the correct symptom according to Pulp issue 2835.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
```diff
@@ -0,0 +1,90 @@
...
+            ))
+            responses.append(client.post(
+                urljoin(repo['_href'], 'actions/publish/'),
+                {'id': repo['distributors'][0]['id']}
+            ))
+
+        # Restart pulp_workers, and verify that queued tasks complete.
+        cli.GlobalServiceManager(self.cfg).restart(('pulp_workers',))
+        for response in responses:
+            tuple(api.poll_spawned_tasks(self.cfg, response.json()))
```
elyezer: Can we check fi the tasks are in the completed state or whatever is the expected state?

elyezer: Maybe I have already forgotten, but my previous comment will be relevant if pool_spawned_tasks does not raise an exception if any of the spawned tasks fails or are not in the completed state.

Ichimonji10: Yes. Great catch! I'll do this.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
```diff
@@ -0,0 +1,90 @@
...
+            responses.append(client.post(
+                urljoin(repo['_href'], 'actions/publish/'),
+                {'id': repo['distributors'][0]['id']}
+            ))
+
+        # Restart pulp_workers, and verify that queued tasks complete.
+        cli.GlobalServiceManager(self.cfg).restart(('pulp_workers',))
+        for response in responses:
+            # pylint:disable=protected-access
+            api._handle_202(self.cfg, response)
```
Ichimonji10: @elyezer Mind if I make a follow-up PR where I tweak `_handle_202` and make it public?

pulp_smash/api.py
```diff
@@ -89,6 +89,16 @@ def echo_handler(server_config, response):  # pylint:disable=unused-argument
...
+def code_handler(server_config, response):  # pylint:disable=unused-argument
+    """Check the response status code, and return the response.
+
+    :raises: ``requests.exceptions.HTTPError`` if the response status code is
+        in the 4XX or 5XX range.
+    """
+    response.raise_for_status()
+    return response
+
+
```
kdelee: One thing that I would find helpful in the doc string for this handler is a short explanation of why I might choose it versus the `safe_handler`

Ichimonji10: Done.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
```diff
@@ -0,0 +1,90 @@
...
+            repos.append(client.get(repo['_href'], params={'details': True}))
+
+        # Request that the repo be synchronized and published. Don't wait.
+        client.response_handler = api.code_handler
+        responses = []
+        for repo in repos:
+            responses.append(client.post(
+                urljoin(repo['_href'], 'actions/sync/')
+            ))
+            responses.append(client.post(
```
bmbouter: Calling a publish is fine but not necessary. Having at least one task type per repo is the important part.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
```diff
@@ -0,0 +1,90 @@
...
+                urljoin(repo['_href'], 'actions/publish/'),
+                {'id': repo['distributors'][0]['id']}
+            ))
+
+        # Restart pulp_workers, and verify that queued tasks complete.
+        cli.GlobalServiceManager(self.cfg).restart(('pulp_workers',))
```
bmbouter: It's probably good to assert that at least one of the tasks dispatched has been assigned to a worker prior to the restart. For example if you submit your tasks and restart too quickly then the test won't have the right conditions to have the assertions be useful. We need to ensure that the worker is actively working on one task. I expect that task to be in the running state and have the worker name set.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
Ichimonji10: My intent is to give the pulp worker that's handling this task more tasks to process, thus widening the window of time where I can restart the `pulp_workers` service and trigger the bug under test. Is it correct to say that this helps accomplish that?

In my personal testing, the workers can be *very* fast. On several of my VMs, workers get through three of the five sync/publish cycles before `pulp_workers` restart, and that's with six VMs running in parallel.

pulp_smash/tests/rpm/api_v2/test_service_resiliency.py
Ichimonji10: I think this test is fundamentally non-deterministic. I could try to make it more deterministic by doing something like this:

1. Line up a *much* greater number of tasks for workers than is currently done. For example, line up 50 sync/publish tasks instead of 5.
2. Wait, so that a task can be assigned. The metric for how long to wait is an interesting question. For example, the test could be dumb and wait one second, or it could be smart but fragile and wait based on the output of `qpid-stat -q`.

...but that increases the time necessary for this test and the load on the resources we use (e.g. repos.fedorapeople.org) without solving the fundamental problem with this test, which is that it's non-deterministic.

bmbouter: It's a hard thing to test for. I was trying to come up with some ideas and I maybe have one. How about syncing epel and one sync of zoo behind it? EPEL's repo should be highly available and it takes like 45 min to sync. It would allow you to check that the task is in the the running state and then issue the restart. If this is working correctly, after restart I expect that task to go to 'canceled' and the zoo sync to begin and complete.

Ichimonji10: Maybe a good solution is to assign different types of work that stress different types of resources. For example, one could assign several tasks that sync and publish repositories (as is done now), and one could also assign tasks that calculate applicability for each of those repositories. The sync/publish tasks stress the system's disk and the applicability calculations stress the system's CPU. This gives the test a greater ability to find and stress bottlenecks.

A danger with this approach is that one might start assigning work that can be consumed faster than it's being assigned, thus muting the effects of any bottlenecks found. For example, if applicability calculation is a multi-threaded process, then a good Xeon, i7, Epyc or Ryzen CPU might just blow right through those tasks. Hrmmm.

Ichimonji10: Perfect!

bmbouter: Sweet! I'm so glad you are Pulp QE.

Ichimonji10:
@bmbouter, I believe the test is now designed as per our earlier conversation. Do you notice any screw-ups

When the `selectors` LOC is commented out, the issue under test is consistently triggered. When `pulp_workers` restarts, the EPEL sync aborts, and the second sync is forever stuck in the waiting state.

bmbouter:
@Ichimonji10 This looks right, and it also is reproducing the issue so that is good. I think it's good to merge.

PR #704, opened 2017-07-03
Use `code_handler` throughout code base

Ichimonji10:
Let the newly added function `pulp_smash.api.code_handler` can replace
several calls to `requests.response.raise_for_status` throughout the
code base.


Ichimonji10:
@kersommoura @kdelee Please review.

Issue #693, opened 2017-06-28
Provide decomposed RPM names in pulp_smash/constants.py


kdelee:
Currently the name, epoch, version, release, and arch are encoded in the pulp_smash.constants.RPM and pulp_smash.constants.RPM2 constants, but not easily accessible. Some tests hard code in this data as a work around:

See:
* https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/rpm/api_v2/test_content_applicability.py#L25
* https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/rpm/api_v2/test_updateinfo.py#L100

To avoid this in the future, it has been suggested to instead provide this in a dicitonary in pulp_smash/constants.py and then concatenate the contents to form the full RPM name so only one reference will have to be updated in the future.

For reference: [RPM naming convention](http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html)

Roughly it would look like this:
```python
# file: pulp_smash/constants.py
from types import MappingProxyType

RPM_DATA = MappingProxyType({
    'name' : 'bear',
     'epoch' : None,
     'version': '4.1',
     'release' : 1,
     'arch' : 'noarch',
     'vendor' : None,
 })

RPM  = '{}-{}{}-{}-{}.rpm'.format(
        RPM_DATA['name'],
        RPM_DATA['epoch'] + '!' if RPM_DATA['epoch'] != None else '',
        RPM_DATA['version'],
        RPM_DATA['release'],
        RPM_DATA['arch'],
 )
```

PR #706, opened 2017-07-06
Use RPM_DATA instead of hard coded values

kdelee:
Continues to address #693 

I ran the tests in the affected file in Fedora 25 VM with pulp 2.13.2 installed. All is well.

@Ichimonji10 I was working on another issue and noticed these hard coded values. 
I could perhaps put the 'sum' in the `constants.py` RPM_DATA metadata as well, if that seems like a good idea. Let me know what you think.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_updateinfo.py
+++ b/pulp_smash/tests/rpm/api_v2/test_updateinfo.py
@@ -49,6 +49,7 @@ from pulp_smash.constants import (
     ORPHANS_PATH,
     REPOSITORY_PATH,
     RPM,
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I don't think that an rpm checksum should go into the metadata section of `RPM_DATA`. The problem is that we need the checksum for the entire RPM, whereas the checksum in an RPM's metadata (if any) would describe some portion of the RPM, not the whole thing.

Making this checksum a constant would be nice. I think it would be more appropriate to make a constant such as `RPM_SHA256_CHECKSUM`.

PR #708, opened 2017-07-07
Add tests that specify checksumtype in drpm upload

### LINKED ISSUE #585: Update drpm upload tests to include --check-sum type
Update the DRPM upload test cases for both cli & api to include --checksum-type

https://pulp.plan.io/issues/2627

We have smash tests for both CLI & API but it doesn't seem to be using the --checksum-type
pulp_smash.tests.rpm.api_v2.test_upload_publish.UploadDrpmTestCase
pulp_smash.tests.rpm.cli.test_upload.UploadDrpmTestCase.test_upload

https://github.com/PulpQE/pulp-smash/commit/7cbb369f8b89222b71d8eb81e1012e2b8b5dcabb

kdelee:
Includes tests in both API and CLI

Closes #585 

@Ichimonji10 
Let me know what you think.

While re-creating this issue, [Pulp issue #2871](https://pulp.plan.io/issues/2871) was discovered. The consequence of 2871 is that if a checksumtype is specified that does not match a specified checksum, no error is thrown, which is a problem. 

Another consequence seems to be that, no matter what "checksumtype" is specified, "sha256" is chosen.

I have not filed a pulp-smash issue yet because it seems there is still some debate about what the problem is -- i.e. should all references be "checksumtype" or is it valid for some to be "checksum_type".

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
@@ -76,6 +76,57 @@ class UploadDrpmTestCase(utils.BaseAPITestCase):
         )
 
 
+class UploadDrpmTestCaseWithCheckSumType(utils.BaseAPITestCase):
+    """Test whether one can upload a DRPM into a repository.
+
+    `Pulp issue #2627 <https://https://pulp.plan.io/issues/2627>`_ caused
+    uploading to fail when "checksumtype" was specified.
+
+    This test case targets `Pulp Smash #585
+    <https://github.com/PulpQE/pulp-smash/issues/585>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Import a DRPM into a repository and search it for content units.
+
+        Specifically, this method does the following:
+
+        1. Create a yum repository.
+        2. Upload a DRPM into the repository with "checksumtype" set to
+            "sha256"
+        3. Search for all content units in the repository.
+        """
+        super(UploadDrpmTestCaseWithCheckSumType, cls).setUpClass()
+
+    def test_all(self):
+        """Test that uploading DRPM with checksumtype specified works."""
+        if selectors.bug_is_untestable(1806, self.cfg.version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/1806')
+        if selectors.bug_is_untestable(2627, self.cfg.version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/2627')
+        client = api.Client(self.cfg)
+        repo = client.post(REPOSITORY_PATH, gen_repo()).json()
+        self.addCleanup(client.delete, repo['_href'])
+        drpm = utils.http_get(DRPM_UNSIGNED_URL)
+        utils.upload_import_unit(
+            self.cfg,
+            drpm,
+            {
+                'unit_type_id': 'drpm',
+                'unit_metadata': {'checksumtype': 'sha256'},
+            },
+            repo,
+        )
+        units = utils.search_units(self.cfg, repo, {})
+        self.assertEqual(len(units), 1, units)
+        # Test if DRPM extracted correct metadata for creating filename.
+        self.assertEqual(
+            units[0]['metadata']['filename'],
+            DRPM,
+        )
+
+
 class UploadSrpmTestCase(utils.BaseAPITestCase):
     """Test whether one can upload a SRPM into a repository.
 
--- a/pulp_smash/tests/rpm/cli/test_upload.py
+++ b/pulp_smash/tests/rpm/cli/test_upload.py
@@ -18,7 +18,8 @@ class UploadDrpmTestCase(unittest.TestCase):
     """Test whether one can upload a DRPM into a repository.
 
     This test case targets `Pulp Smash #336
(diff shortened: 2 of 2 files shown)


pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -76,6 +76,61 @@ def test_drpm_file_name_is_correct(self):
...
+        utils.upload_import_unit(
+            self.cfg,
+            drpm,
+            {
+                'unit_type_id': 'drpm',
+                'unit_metadata': {'checksumtype': 'sha256'},
+            },
+            repo,
+        )
+        units = utils.search_units(self.cfg, repo, {}, api.safe_handler)
```
Ichimonji10: The default response handler is `api.json_handler`. If the default is used, then the calls to `.json()` can be omitted from the code below.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -76,6 +76,61 @@ def test_drpm_file_name_is_correct(self):
...
+
+        with self.subTest(comment='Test if DRPM uploaded.'):
+            self.assertEqual(len(units.json()), 1)
```
Ichimonji10: When this assertion fails, the user will be faced with an error message like "assertion failed: 2 not equal to 1." The user won't know what `units.json()` is, though. This assertion gives the user a little extra information about what went wrong then the test fails:

```python
self.assertEqual(len(units.json()), 1, units.json())
```

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -76,6 +76,61 @@ def test_drpm_file_name_is_correct(self):
...
+            drpm,
+            {
+                'unit_type_id': 'drpm',
+                'unit_metadata': {'checksumtype': 'sha256'},
+            },
+            repo,
+        )
+        units = utils.search_units(self.cfg, repo, {}, api.safe_handler)
+
+        with self.subTest(comment='Test if DRPM uploaded.'):
```
Ichimonji10: What if `len(units.json()) == 0`? It's a pretty reasonable outcome. In that case, the assertion on the line below will fail... and the test will continue on to the next assertion, which will blow up on the `units.json()[0]` line. In this case, it's probably easier to just avoid `subTest` context managers:

```python
self.assertEqual(len(units.json()), 1, units.json())
self.assertEqual(units.json()[0]['metadata']['filename'], DRPM)
```

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -75,3 +76,59 @@ def test_upload(self):
...
+        drpm_file = os.path.join(temp_dir, os.path.split(DRPM)[-1])
+        client.run(
+            'curl -o {} {}'.format(drpm_file, DRPM_UNSIGNED_URL).split()
+        )
+
+        # Upload the DRPM into the repository using checksum-type
+        client.run(
+            'pulp-admin rpm repo uploads drpm --repo-id {} --file {}'
+            .format(repo_id, drpm_file).split() +
+            '--checksum-type sha256'.split()
```
Ichimonji10: This is funky. How about this?

```python
'pulp-admin rpm repo uploads drpm --repo-id {} --file {} '
'--checksum-type sha256'
.format(repo_id, drpm_file).split()
```

kdelee:
@Ichimonji10 
Great feedback, thanks. I'll work on those changes.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -75,3 +76,59 @@ def test_upload(self):
...
+        proc = client.run(
+            'pulp-admin rpm repo content drpm --repo-id {} --fields filename'
+            .format(repo_id).split()
+        )
+        self.assertEqual(proc.stdout.split('Filename:')[1].strip(), DRPM)
+
+        # Upload the DRPM into the repository. Pass --skip-existing.
+        proc = client.run(
+            ('pulp-admin rpm repo uploads drpm --repo-id {} --file {} '
+             '--skip-existing')
```
Ichimonji10: Nitpick: These parentheses are redundant.

kdelee:
@Ichimonji10 
Made those changes you suggested.

I have not done it, but I could refactor the API test that I based mine off of to be more like this one. 

`pulp_smash.tests.rpm.api_v2.UploadDrpmTestCase` tests the return code (default code_handler does this, right?), and does the same thing with the `units = utils.search_units(self.cfg, repo, {}, api.safe_handler)` that I had done.

Does that sound like something that should be done in this PR or should I stick to the tests I was writing to close this issue?

Ichimonji10:
`pulp_smash.tests.rpm.api_v2.test_upload_publish.UploadDrpmTestCase` makes use of an anachronistic coding style. I'd love to see it refactored to look more like the new test case that you've written.

> Does that sound like something that should be done in this PR or should I stick to the tests I was writing to close this issue?

If I had my druthers, I'd refactor that existing test case in a separate commit. A refactor would be great regardless of how it lands.

kdelee:
@Ichimonji10 

Ready to merge now. Thanks for the feedback. 
I will do the refactor in a different PR.

PR #710, opened 2017-07-10
Add tests for vendor data in uploaded files.

### LINKED ISSUE #680: Test that the vendor field is available for uploaded and synced rpms
https://pulp.plan.io/issues/2781
1. Create a repo with synced and uploaded content
2. Compare the rpm (search) to see that vendor field is populated in both

kdelee:
Closes #680

Includes a refactor of existing test in pulp_smash.tests.rpm.cli.test_upload

@Ichimonji10 I maybe got a little crazy on the refactor. I just felt silly literally copying and pasting the code to make a repo, upload, and check for results of a search query. 

I believe that because the individual tests each call these "helper functions" that if an exception is thrown within them, only that test will fail and addCleanup will still run if resources like the repo were successfully created and need to be deleted.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
@@ -20,6 +20,9 @@ from pulp_smash.constants import (
     REPOSITORY_PATH,
     RPM,
     RPM_DATA,
+    RPM_WITH_VENDOR_URL,
+    RPM_WITH_VENDOR_DATA,
+    RPM_WITH_VENDOR,
     RPM_UNSIGNED_URL,
     SRPM,
     SRPM_UNSIGNED_URL,
@@ -254,7 +257,8 @@ class UploadRpmTestCase(utils.BaseAPITestCase):
         Verify that only one content unit is in ``repo``, and that several of
         its metadata attributes are correct. This test targets `Pulp #2365
         <https://pulp.plan.io/issues/2365>`_ and `Pulp #2754
-        <https://pulp.plan.io/issues/2754>`_
+        <https://pulp.plan.io/issues/2754>`_ and `Pulp #2781
+        <https://pulp.plan.io/issues/2781>`_
         """
         units = utils.search_units(self.cfg, repo)
         self.assertEqual(len(units), 1)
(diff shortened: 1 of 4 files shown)


pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -332,3 +336,82 @@ def verify_repo_download(self, repo):
...
+
+    1. Create a repository.
+    2. Upload an RPM to the first repository, and publish it.
+    3. Test that RPM metadata includes all available fields, including vendor.
+    """
+
+    def test_all(self):
+        """Create a pair of RPM repositories."""
+        if selectors.bug_is_untestable(2781, self.cfg.version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/2781')
```
Ichimonji10: Within an instance method, call `self.skipTest()`, not `raise unittest.SkipTest()`. Raising an exception is only necessary in places like static methods, where `self.skipTest()` isn't available.

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -332,3 +336,82 @@ def verify_repo_download(self, repo):
...
+        rpm = utils.http_get(RPM_WITH_VENDOR_URL)
+        client = api.Client(self.cfg, api.json_handler)
+        body = gen_repo()
+        body['distributors'] = [gen_distributor()]
+        repo = client.post(REPOSITORY_PATH, body)
+
+        # Info about repo distributors is needed when publishing.
+        repo = client.get(repo['_href'], params={'details': True})
+        self.addCleanup(client.delete, repo['_href'])
```
Ichimonji10: What happens in the case where `client.post(...)` succeeds, but `client.get(...)` fails? The repository is never cleaned up. Please re-position this addCleanup call

pulp_smash/tests/rpm/api_v2/test_upload_publish.py
```diff
@@ -332,3 +336,82 @@ def verify_repo_download(self, repo):
...
+        self.addCleanup(client.delete, ORPHANS_PATH)
```
Ichimonji10: BaseAPITestCase does this already. Either drop BaseAPITestCase as a parent class, or drop this line of code.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -14,7 +21,115 @@ def setUpModule():  # pylint:disable=invalid-name
...
+    def setUpClass(cls):
+        cls.cfg = config.get_config()
+        cls.client = cli.Client(cls.cfg)
+
+    def temp_download(self, file_url, filename='tmp.rpm'):
+        """Download a file from the given url to a temp directory.
+
+        Args:
+            file_url: String. The URL of the file to download. Required
+            filename: String. Assigns a name to the file downloaded. Optional.
```
Ichimonji10: Sphinx isn't configured to parse Google/napoleon-style docstrings. Please use the format that Sphinx expects:

```
:param file_url: The URL of the file to download.
```

It's not necessary to write that `file_url` is required. Python will throw an exception if this parameter isnt' passed.

Run `make docs-clean && make docs-html` to see whether these docstrings are compiled correctly.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -14,7 +21,115 @@ def setUpModule():  # pylint:disable=invalid-name
...
+        """
+        temp_dir = self.client.run('mktemp --directory'.split()).stdout.strip()
+        self.addCleanup(self.client.run, 'rm -rf {}'.format(temp_dir).split())
+        file_path = os.path.join(temp_dir, os.path.split(filename)[-1])
+        try:
+            self.client.run(
+                'curl -o {} {}'.format(file_path, file_url).split()
+            )
+        except urllib.error.HTTPError:
+            raise self.skipTest('Skipping test, could not download resource')
```
Ichimonji10: Maybe the exception should just bubble up and cause the test to fail? I don't see the advantage to obscuring an HTTPError.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -14,7 +21,115 @@ def setUpModule():  # pylint:disable=invalid-name
...
+        if 'args' not in kwargs.keys():
+            kwargs['args'] = ''
+        return self.client.run(
+            'pulp-admin {} repo uploads {} --repo-id {} --file {} {}'
+            .format(
+                kwargs['plugin'],
+                kwargs['filetype'],
+                kwargs['repo_id'],
+                kwargs['filename'],
+                kwargs['args']
```
Ichimonji10: This isn't a great way to use kwargs. What happens if `kwargs['filetype']` isn't present? A KeyError exception is thrown. Given that these arguments are required, it's best to just list them as required arguments.

```python
upload_to_repo(self, plugin, filetype, repo_id, ...)
```

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -14,7 +21,115 @@ def setUpModule():  # pylint:disable=invalid-name
...
+            filename: String. Full path of file. Required
+            args: String. Additional arguments to uploads command. Optional
+
+        Returns:
+            Completed process.
+
+        """
+        if 'args' not in kwargs.keys():
+            kwargs['args'] = ''
+        return self.client.run(
```
Ichimonji10: `cli.Client` objects are't good candidates for being passed around. Their `response_handler` is designed to be changed, and if the state of a mutable `cli.Client` object isn't exactly as expected, hard-to-diagnose errors can occur. Better to not reference `self.client` at all, and make a `cli.Client` object as needed. And at that point, a simple function will do. There's several examples of those.

pulp_smash/tests/rpm/cli/test_upload.py
```diff
@@ -35,45 +150,37 @@ def test_upload(self):
...
-        proc = client.run(
-            ('pulp-admin rpm repo uploads drpm --repo-id {} --file {} '
-             '--skip-existing')
-            .format(repo_id, drpm_file).split()
+        proc = self.upload_to_repo(
+            plugin='rpm',
+            filetype='drpm',
+            repo_id=repo_id,
+            filename=drpm_file_path,
+            args='--skip-existing',
```
Ichimonji10: The word `args` isn't a keyword, but it's conventionally used to indicate a list of un-named function parameters:

```python
def print_args(*args):
    for arg in args:
        print(arg)
```

I think the design of the existing RPM CLI helper functions in `pulp_smash.tests.docker.cli.utils` may serve as a good starting point, design-wise.

kdelee:
@Ichimonji10 
I agree, I will roll this back to just the changes with the constants and we can work on these other refactors in another PR. Thanks for taking the time to look at it!

kdelee:
@kersommoura here is the PR I was working on for #680

Ichimonji10:
Is it correct to assume that this PR won't be updated? Should it be closed?

kersommoura:
I talked with @kdelee yesterday. We discussed that split in 2 PR's could be another solution. One just to update the constant. And a second PR for test scenario itself.  I can work on it.

kdelee:
I can make a separate PR with just the constant info.

PR #711, opened 2017-07-11
Work around a Pulp 2 defect causing false positive

dralley:
The test for fastforward functionality falsely fails due
to a fuzzy timestamp issue in Pulp which likely won't be
fixed in Pulp 2.


Ichimonji10:
Tested. Works.

PR #718, opened 2017-07-24
Update OSTree constants and tests that use them.

kersommoura:
From now on OSTree files will be generated using pulp_fixtures scripts.
OSTree fixtures are now available. See: https://github.com/PulpQE/pulp-fixtures/pull/56

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -186,16 +186,14 @@ ORPHANS_PATH = 'pulp/api/v2/content/orphans/'
     http://docs.pulpproject.org/en/latest/dev-guide/integration/rest-api/content/orphan.html
 """
 
-OSTREE_BRANCH = 'fedora-atomic/f21/x86_64/updates/docker-host'
+OSTREE_BRANCHES = ['rawhide', 'stable']
 """A branch in :data:`OSTREE_FEED`. See OSTree `Importer Configuration`_.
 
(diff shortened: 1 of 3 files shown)


pulp_smash/constants.py
```diff
@@ -186,15 +186,15 @@
...
+OSTREE_BRANCHES = ['rawhide', 'stable']
 """A branch in :data:`OSTREE_FEED`. See OSTree `Importer Configuration`_.
 
 .. _Importer Configuration:
     http://docs.pulpproject.org/plugins/pulp_ostree/tech-reference/importer.html
 """
 
 OSTREE_FEED = (
-    'https://repos.fedorapeople.org/pulp/pulp/demo_repos/test-ostree-small'
+    'https://repos.fedorapeople.org/pulp/pulp/fixtures/ostree/small/'
```
Ichimonji10: `OSTREE_FEED` is now part of pulp-fixtures, so its URL should be derived from `PULP_FIXTURES_BASE_URL`:

```python
 OSTREE_FEED = urljoin(PULP_FIXTURES_BASE_URL, 'ostree/small/')
```

Ichimonji10:
I ran the tests in the affected modules, and they continue to pass, which is good.

pulp_smash/constants.py
kersommoura: Updated.

PR #721, opened 2017-07-26
Drop duplicate private constant

Ichimonji10:
Use `pulp_smash.constants.DOCKER_UPSTREAM_NAME` instead of a private
constant.


Ichimonji10:
I ran the tests in the affected modules against Pulp 2.14 beta 1 on F24, F25 and RHEL 7. They still pass.

PR #722, opened 2017-07-26
Add constant for RPM package size

Ichimonji10:
There are at least three ways to measure the size of an RPM package:

* installed size
* archive size
* package size

The `RPM_DATA` constant already lists the installed size of an RPM. Add
another entry for the package size of an RPM.

See:
https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-package-structure.html#id623000


Ichimonji10:
This fixes a test failure in the Pulp 2.14 beta release.

Ichimonji10:
@kdelee Good idea. Updated.

kdelee:
👍

PR #738, opened 2017-07-31
Move Docker tests into more appropriate location

Ichimonji10:
Most of the tests in `pulp_smash.tests.docker.cli.test_sync_publish` use
Pulp's CLI and API interfaces. Rewrite the tests to use just Pulp's API,
and move them to `pulp_smash.tests.docker.api_v2.test_sync_publish`.


Ichimonji10:
Tested against Pulp 2.14 beta 3 on F24, F25 and RHEL 7.

Related to https://github.com/PulpQE/pulp-smash/pull/724

Ichimonji10:
I just pushed some additional changes that update some docstrings and test case names. I think it's appropriate to include the changes in this PR, as this PR is focused on reorganizing code. I don't think that these changes invalidate the review already done - I re-ran the same tests against the same VMs, and the tests still succeed.

PR #741, opened 2017-08-02
Port and expand tests for copying between repos

Ichimonji10:
The tests in module `pulp_smash.tests.cli.test_copy` verify that images,
tags and manifests can be copied between Docker repositories. There are
a few drawbacks to the current implementation:

* There is currently one test case for each type of content that can be
  copied. This is wasteful given that the test cases for copying tags
  and manifests both include the same set-up work.
* The code for determining which images, tags or manifests have been
  copied from one repository to another is fragile. The `pulp-admin`
  tool's output varies based on both the version of the tool (2.13
  changed things) and the type of content being copied.
* As of Pulp 2.14, manifest lists can be copied between repositories.
  There are no tests for this feature.

Drop these CLI tests, and replace them with corresponding API tests:

* There is now one test case that deals with all schema v1 content, and
  another that deals with all schema v2 content. This greatly reduces
  the amount of set-up work that must be done.
* The code for determining which content is in a repository now makes
  use of the API, which has a much more stable interface.
* There is now a test for copying manifest lists between repositories.


Ichimonji10:
Tested against Pulp 2.14 beta 3 on F24, F25 and RHEL 7.

Ichimonji10:
We're short on experienced manpower right now, so I'm going to go ahead and merge this PR. If the automation results tank, it's my fault.

Issue #713, opened 2017-07-17
Test Docker Manifest Support


Ichimonji10:
Pulp 2.14 adds support for Docker manifest lists. This change is significant, and it really needs to be tested before 2.14 is released. Some of the testing for this new feature can be automated, but given the lack of Docker clients on our automated systems, some of the testing will need to be manual.

A docker manifest list is a list of docker manifests, and a docker manifest describes the layers of a docker image. A typical docker manifest list will list out one docker manifest per supported architecture, where supported architectures are values like amd64-linux or arm32v7. In addition, there's two versions of docker manifests: schema v2 and schema v1.

New docker clients can natively understand docker manifest lists. When a new docker client contacts the docker hub (or a clone like Pulp + Crane), it will express its support for this new format by sending certain HTTP headers saying so. When an older docker client contacts the docker hub (or a clone), the hub must be capable of downgrading its response on-the-fly. Pulp + Crane do this by pre-emptively contacting the docker hub and fetching all the content that each generation of docker clients might want.

We need to create automated tests that do the following:

* Create a Docker repository, and let it sync content from a repository that has a manifest list with entries for more than one architecture.
* Create a Docker repository, and let it sync content from a repository that has a manifest list with no entry for the amd64-linux architecture.

For each repository, we need the following tests:

* Pretend to be a modern Docker client, and get (or at least list) the contents of the repository.
* Pretend to be an older Docker client, and get (or at least list) the contents of the repository.
* Pretend to be an even older Docker client, and get (or at least list) the contents of the repository.

We also need to manually test this new feature by installing and configuring the official Docker client and ensuring that it can `docker pull` from each repository.

To fully understand the testing that needs to be done, please read https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility

The following Docker repositories may be useful for testing:

* https://hub.docker.com/r/dmage/manifest-list-test/tags/
* https://hub.docker.com/r/dmage/busybox/tags/

Related to: https://github.com/PulpQE/pulp-smash/issues/555

Ichimonji10:
I suspect the following failing test cases are related:

* `pulp_smash.tests.docker.cli.test_copy.CopyAllManifestsTestCase`
* `pulp_smash.tests.docker.cli.test_sync_publish.SyncPublishV2TestCase`

preethit:
Adding issues that may be related to this 
https://pulp.plan.io/issues/2924
https://pulp.plan.io/issues/2920

[...]

PR #744, opened 2017-08-02
Test fetching an image w/out an amd64/linux build

Ichimonji10:
Add a new Docker test case, `NoAmd64LinuxTestCase`. From its docstring:

> A manifest list lets a single Docker repository contain multiple
> images.  This is useful in the case where an image contains
> platform-specific code, and an image must be built for each each
> supported architecture, OS, etc.
>
> When a modern Docker client fetches an image, it does the following:
>
> 1. Get a manifest list.
> 2. Look through the list of available images.
> 3. Pick out an image that functions on the current host's platform.
> 4. Get a manifest for that image.
> 5. Use the information in the manifest to get the image layers.
>
> Older Docker clients aren't aware of manifest lists, and when they go
> to fetch an image, they just ask for any old manifest from a
> repository. When a Docker registry receives such a request, it does
> the following:
>
> 1. Look through the list of available images.
> 2. If an image with an `architecture` of `amd64` and an `os` of
>    `linux` is available, return its manifest. Otherwise, return an
>    HTTP 404.
>
> This test case verifies Pulp's behaviour in the case where an upstream
> Docker repository has content described by a manifest list.
>
> This test case doesn't verify Pulp's behaviour in the case where an
> upstream Docker repository has content described by a v2 manifest or
> v1 manifest. In these cases, the correct behaviour of a Docker
> registry is not well defined. See the [backward
> compatibility](https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility)
> documentation.

Fix: https://github.com/PulpQE/pulp-smash/issues/713

### DIFF
--- a/pulp_smash/tests/docker/api_v2/test_sync_publish.py
+++ b/pulp_smash/tests/docker/api_v2/test_sync_publish.py
@@ -494,6 +494,162 @@ class NonNamespacedImageTestCase(SyncPublishMixin, unittest.TestCase):
             self.assertFalse(repos[repo_id]['protected'])
 
 
+class NoAmd64LinuxTestCase(SyncPublishMixin, unittest.TestCase):
+    """Sync a Docker image with no amd64/linux build.
+
+    A manifest list lets a single Docker repository contain multiple images.
+    This is useful in the case where an image contains platform-specific code,
+    and an image must be built for each each supported architecture, OS, etc.
+
+    When a modern Docker client fetches an image, it does the following:
+
+    1. Get a manifest list.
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against Pulp 2.14 beta 3 on F24, F25 and RHEL 7 with:

    python -m unittest pulp_smash.tests.docker.api_v2.test_sync_publish

We're short on experienced manpower right now, so I'm going to go ahead and merge this PR, pending Travis results. If the automation results tank, it's my fault.

PR #745, opened 2017-08-02
Use more modern method to list optional deps

Ichimonji10:
setuptools lets one define optional dependencies in `setup.py`. Take
advantage of this capability. See:
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies


Ichimonji10:
According to https://stackoverflow.com/a/4806441, this features has been available since Pip 1.1, released in February 2011.

Thanks to @elyezer for informing me of this feature.

Ichimonji10:
We're short on experienced manpower right now, so I'm going to go ahead and merge this PR, pending Travis results. If the automation results tank, it's my fault.

Issue #663, opened 2017-06-01
Test for verifying --checksum-type works with upload


preethit:
Verify that --checksum-type works consistently with repo upload

https://pulp.plan.io/issues/2774

preethit:
This may be another one for @kersommoura and @kdelee

kersommoura:
@kdelee, I will work on this issue.

PR #756, opened 2017-08-15
Verify that checksum value was calculated according to checksum type.

kersommoura:
* Problem: Calculated checksum value is not according to specified checksum type.
  Even when `checksumtype` is provided as "md5", the calculated value is according
  to "sha256".

* Test:
        1. Create a yum repository.
        2. Upload a DRPM into the repository with "checksumtype" set to
           "md5"
        3. Assert that "checksumtype" was set to "md5".
        4. Assert that checksum value was calculated according to "md5".

Closes: #663

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
+++ b/pulp_smash/tests/rpm/api_v2/test_upload_publish.py
@@ -127,6 +127,59 @@ class UploadDrpmTestCaseWithCheckSumType(utils.BaseAPITestCase):
         )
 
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Sorry I didn't get to this today. I'm going to have to look at it on Monday.

Ichimonji10:
Merged as 48c3109b88cb69c6af56531a873fa05d1f3ca0d6, with a follow-up commit of d18fbd1a2f02571389d70122a6931f557df41c9a.

Issue #572, opened 2017-03-03
Validate unit count ( removed_count) in sync report after remove_missing


preethit:
Add this scenario to the remove_missing coverage
https://pulp.plan.io/issues/2616

https://github.com/PulpQE/pulp-smash/issues/463

kersommoura:
@kdelee, I am working on this one.

PR #761, opened 2017-08-18
Re-sync a child repository with the remove_missing enabled.

kersommoura:
* Problem: After sync with remove_missing option the next publish is no-op sometimes.

* Test:
    1 - Create 1st repo with, feed, sync and publish.
    2 - Create 2nd repo with feed pointed to 1st repo, and
        `remove_missing` enabled, sync.
    3 - Remove an arbitrary number of units from 1st repo, re-publish it.
    4 - Re-sync the 2nd repo. Assert that `removed_count` is equal to
    number of removed units.

Closes: #572

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_remove_unit.py
+++ b/pulp_smash/tests/rpm/api_v2/test_remove_unit.py
@@ -8,7 +8,7 @@ import random
 import unittest
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Merged as f4fea497fa82b86210e3c4249f65961b0b6707cc, with a follow-up commit of 2c8e6b0de887715ecd15263e345791d0cd08cce0.

Issue #286, opened 2016-06-01
Test publish yum repo with different checksums


mhrivnak:
1. create an "rpm" repo in pulp with one or more RPMs, SRPMs, and DRPMs.
2. publish it with the "sha256" checksum type
3. verify publish (see below)
4. publish it with the "sha1" checksum type
5. verify publish (see below)
### Verification

It depends on how fancy and robust you want to get when it comes to inspecting the published XML. You'll need to somehow look at primary, other, filelists, and prestodelta XML files and make sure all of the unit-related checksums are using the checksum type you asked for.

Smash could download the files and actually parse them as XML. That wouldn't be too difficult, but who likes parsing XML...

Smash could just ensure that the string "sha256" shows up in the case when it should, and doesn't when it shouldn't.

In the case of "other" and "filelists", the checksum type itself won't be included. Each package will have a `pkgid` attribute that is the checksum value of the specified type. You could just verify that it's the correct length. Or you could actually cross-reference it with the checksum for the corresponding package listed in primary.

Another general way to help with this verification is to just make sure yum and/or dnf can actually read the metadata and use it. In my manual testing, I published a rhel7 kickstart repo and kickstarted a VM from it, just to make sure it was generally sane.

kersommoura:
@mhrivnak as we talk the step 4 - publish it with the "sha1" checksum type has to executed with the --force-full option to force a full re-publish to happen.

Ichimonji10:
@kersommoura 

```python
with self.subTest(comment='filelists.xml'):
    self.verify_filelists_xml(cfg, repo['distributors'][0], 'sha256')
with self.subTest(comment='other.xml'):
    self.verify_other_xml(cfg, repo['distributors'][0], 'sha256')
if feed == DRPM_UNSIGNED_FEED:
    with self.subTest(comment='prestodata.xml'):
        self.verify_prestodata_xml(cfg, repo['distributors'][0], 'sha256')
```

kersommoura:
@Ichimonji10 

I will work on it.

Thanks for the review.

PR #765, opened 2017-08-24
Add tests to verify checskum types on yum repo metadata.

kersommoura:
* Problem: Update the checksum type on distributor should be reflected on repo metadata after re-publishing the repo.

* Test:
    1. Create, sync and publish a repository with checksum type "sha256".
    2. Assert that the checksum type is the only one present on the
       `primary.xml`
    3. Update the checksum type to "sha1", and use option `force_full`
       set as True.`force_full` will force a complete re-publish to happen.
       Re-publish the repository.
    4. Assert that the updated checksum type is the only checksum type
       present on `primary.xml`.

Closes: #286


Ichimonji10:
Rebased and merged as 897ac1c33fd527e6ff12ee139ecee4d291571154, with a follow-up commit of f6548606d8a7b389c844880f465b8e02cc49622d.

PR #767, opened 2017-08-30
Add more tests to inspect XML files.

kersommoura:
Inspected XML files:

1. `filelists.xml` and `other.xml` - Assert that length of `pkgid`
    is according to the given checksum type.
2.  `prestodelta.xml` - Assert that given checksum type is the only
     one present.

Add more XML namespaces to constants file, RPM_NAMESPACES.

See also: 897ac1c

Closes: #286

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
+++ b/pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
@@ -84,13 +84,30 @@ class UpdatedChecksumTestCase(unittest.TestCase):
     Do the following:
 
     1. Create, sync and publish a repository with checksum type "sha256".
-    2. Assert that the checksum type is the only one present on the
-       ``primary.xml``.
+
+    2. After that several XML files - part of repodata will be inspected.
+
+       ``primary.xml``
+           Assert that the given checksum type is the only one present.
+
+       ``filelists.xml`` and ``other.xml``
+           Assert that the length of ``pkgid`` is according to the given
+           checksum type.
+
+       ``prestodelta.xml``
+           Assert that the given checksum type is the only one present.
+
+    For ``sha256`` the length of ``pkgid`` should be 64 hex digits, and
+    for ``sha1`` the length should be 40 hex digits.
+    For DRPM repositories the ``prestodelta.xml`` will be inspected as
+    well.
+
     3. Update the checksum type to "sha1", and use option ``force_full`` set as
-       True. ``force_full`` will force a complete re-publish to happen.
-       Re-publish the repository.
-    4. Assert that the updated checksum type is the only checksum type present
-       on ``primary.xml``.
+    True. ``force_full`` will force a complete re-publish to happen.
+    Re-publish the repository.
+
+    4. All aforementioned assertions performed on the step 2 will be executed
+    again to assure that new checksum type was updating properly.
 
     This test targets `Pulp Smash #286
     <https://github.com/PulpQE/pulp-smash/issues/286>`_.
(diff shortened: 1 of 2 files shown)


pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
```diff
@@ -146,3 +173,46 @@ def verify_primary_xml(self, cfg, distributor, checksum_type):
...
+    def verify_pkgid_len(self, pkgid_len, checksum_type, comment):
+        """Perform assertion based on given the checksum type.
+
+        Based on the checksum type provided different assertions will
+        be performed over the length of pkgid.
+        """
+        if checksum_type == 'sha256':
+            # sha256 is 256 bit long, and uses 4 bits
+            # then 256/4 = 64
+            with self.subTest(comment=comment):
```
Ichimonji10: How about dropping this `self.subTest` context manager? It adds complexity without improving the behaviour of this test in any way that I can see. The calls to `verify_filelists_xml`, `verify_other_xml`, etc are already all wrapped in `subTest` context manager.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
```diff
@@ -146,3 +173,46 @@ def verify_primary_xml(self, cfg, distributor, checksum_type):
...
+                self.assertEqual(pkgid_len, {64})
+        if checksum_type == 'sha1':
+            # sha1 is 160 bit long, and uses 4 bits
+            # then 160/4 = 40
+            with self.subTest(comment=comment):
```
Ichimonji10: How about dropping this `subTest` context manager for the reason outlined above?

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
```diff
@@ -146,3 +173,46 @@ def verify_primary_xml(self, cfg, distributor, checksum_type):
...
+
+    def verify_pkgid_len(self, pkgid_len, checksum_type, comment):
+        """Perform assertion based on given the checksum type.
+
+        Based on the checksum type provided different assertions will
+        be performed over the length of pkgid.
+        """
+        if checksum_type == 'sha256':
+            # sha256 is 256 bit long, and uses 4 bits
+            # then 256/4 = 64
```
Ichimonji10: FYI, this relationship can be expressed algebraicallyl. If a single term (like "bits") is in both a numerator and denominator, it gets crossed out, just like for numbers.

    # 256 bits / (4 bits / 1 hex digit) == 64 hex digits.

What I find nice about a representation like this is that every unit is identified. No need to wonder what "4" corresponds to, or what "64" corresponds to.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
```diff
@@ -84,13 +84,22 @@ class UpdatedChecksumTestCase(unittest.TestCase):
     Do the following:
 
     1. Create, sync and publish a repository with checksum type "sha256".
-    2. Assert that the checksum type is the only one present on the
-       ``primary.xml``.
+    2. After that several XML files - part of repodata will be inspected.
+       ``primary.xml``  - Assert that the checksum type is the only
+       one present.
```
Ichimonji10: This text is line-wrapped. Semantically, there is no break of any kind between the end of this sentence and the beginning of the next. reStructuredText offers semantic constructs for doing this, like definition lists:

```
term one
    Definition of term one.
term two
    Definition of term two.
```

They could be adapted to this situation.

kersommoura: Updated per comments.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
kersommoura: Updated per comments.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
kersommoura: It was an attempt to bring more info in case of a fail.  Updated per comments.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
kersommoura: Updated per comments.

@Ichimonji10 Thanks for the review.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
Ichimonji10: Aye. It makes it more obvious where this method is being called from, right? Understood. Still - I think avoiding nested `subTest` context managers is probably for the better. AFAIK, the correct behaviour of a test runner when dealing with nested `subTest` context managers isn't documented.

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
Ichimonji10: Try one more time. Here's the documentation for reST [definition lists](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists).

pulp_smash/tests/rpm/api_v2/test_unavailable_checksum.py
kersommoura: I created an extra function to avoid to have the same piece of code in 2 different places - doing the same assertions. I decided to add  `subTest` to bring more info in case a test failed. As you mentioned, I ended up with nested `subTest` context managers. I did few tests to see if there is any benefit keeping it. I could not see any improvement in the error description.

Ichimonji10: Oh, neat. I figured the inner subtest's comment would be displayed. I guess I'm wrong.

Ichimonji10:
Good job. Merged as 794bf736bd954d6929d3fbc4e2ef142a46f8625f, with a follow-up of 6f40675d08ef9432969b53fd2c5f4024bb3d4409.

Issue #769, opened 2017-09-12
Test that recursive errata copy copies the rpms


preethit:
https://pulp.plan.io/issues/3004

pulp-admin rpm repo create --repo-id=foo2  --feed=https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_errata_install/
 pulp-admin rpm repo sync --repo-id=foo2
 pulp-admin rpm repo sync run --repo-id=foo2
 pulp-admin rpm repo create --repo-id=foo3
 pulp-admin rpm repo  copy errata --from-repo-id=foo2 --to-repo-id=foo3 --recursive

PR #772, opened 2017-09-20
Test that recursive copy of erratas copies RPM packages.

kersommoura:
* Problem:
	When copying erratas from a repo to another using recursive
	as True, the RPMs were not copied.
* Test:
        1 - Create, and sync 1st repo with errata, and RPM packages.
        2 - Create 2nd repo.
        3 - Copy units from from 1st repo to 2nd repo using
        `recursive` as True, and filter  `type_id` as `erratum`.
        4 - Assert that RPMs packages were copied as well.

Closes: #769


Ichimonji10:
Rebased and merged as 5b829b894373c3922ad96da8967b5b5441555519.

Issue #525, opened 2017-01-30
Progress report for failed sync contains no errors


elyezer:
Fore more information see: https://pulp.plan.io/issues/1376

elyezer:
This was moved to Spint 15

PR #775, opened 2017-09-27
Test whether error report is according to error cause.

kersommoura:
* Problem: It is possible to trigger the sync of repo with an invalid feed URL,
 ater read the task, the error report fields were empty.

* Test:

        1. Create, and sync a repository using an invalid feed URL.
        2. Read the related task.
        3. Assert that error report is according to error cause.

Closes: #525


Ichimonji10:
Merged, with a follow-up commit of cba8d4a2d755831e86d372f3a90a2ead982578be.

PR #778, opened 2017-10-03
Update docstrings classes GlobalServiceManager and ServiceManager.

kersommoura:
Pulp can deployed in different ways. Update the docstrings of
`GlobalServiceManager` and `ServiceManager` classes to clarify
when and how to use them to manager services with different
`Pulp deployments`.

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -269,7 +269,7 @@ class BaseServiceManager(metaclass=ABCMeta):
 
     Subclasses should take advantage of the helper methods offered by this
     class in order to manage services and check the proper service manager
-    softeare available on a system.
+    software available on a system.
 
     This base class also offers a context manager to temporary disable SELinux.
     It is useful when managing services on RHEL systems earlier than 7 which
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Thanks for this. It's good to get concrete feedback on the ways in which the documentation is and isn't clear.

I do want to make sure that code examples are focused and limited. This is because prose-based explanations of concepts tend to be easier to maintain than code-based explanations of concepts. It's easy to write a code sample that has minor imperfections, and these minor imperfections can greatly frustrate people who are learning. This is doubly true given that classes, functions, etc tend to evolve over time.

At the same time, the documentation needs to effectively teach readers. No question about that.

Have a look at what I've done in 675d67566294affb49eedb49874f286f8ee640f5. We can make more revisions as needed.

kersommoura:
I just read what you updated. It is more clear in my opinion. Thanks. 

My attempt to updated the documentation was an exercise to see how much I did understand about the use of these classes.

Ichimonji10:
> My attempt to update the documentation was an exercise to see how much I did understand about the use of these classes.

You're doing well.

Issue #650, opened 2017-05-22
Testing for Tasking System Bugfix


bmbouter:
This is a test for [Issue 2770](https://pulp.plan.io/issues/2770).

When a sync happens the message flow goes from httpd -> broker -> resource_manager -> broker -> worker. There are two tests, each asserting a message being sent to the broker.

The first test tests httpd dispatching a task while the broker is down:

1. Start with a healthy Pulp system with no tasks executing
2. Stop the broker
3. Sync a single repo. The call will fail with a 500 error, not great but expected.
4. Check for tasks in the waiting state, and observe that there should be 0 if this bug is fixed. If this bug is not fixed I expect there to be 1.

The second test tests the resource_manager dispatching a task to the broker. This is for rarely occurring situations because if the resource_manager receives the message from the broker the chances of it being down when it goes to send a message back are rare. The only way I know to test it is by adding a line like the patch below.

```patch
diff --git a/server/pulp/server/async/tasks.py b/server/pulp/server/async/tasks.py
index 0d57197..725be2c 100644
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -468,6 +468,7 @@ class Task(PulpTask, ReservedTaskMixin):
         group_id = kwargs.pop('group_id', None)
 
         try:
+            raise KeyError('oh no')
             async_result = super(Task, self).apply_async(*args, **kwargs)
         except:
             if 'task_id' in kwargs:
```

1. Start with a healthy Pulp system with no tasks executing
2. Patch the installation as shown above and restart services because the patch is applied
3. sync a repo
4. assert that the task enters the state 'error' soon

Ichimonji10:
Can I get a link to a pulp.plan.io issue?

bmbouter:
I accidentally hit submit with only the title. There is content there now.

Ichimonji10:
Thank you!

PR #785, opened 2017-10-10
Test whether httpd dispatches a task while the broker is down.

kersommoura:
* Test:

  With a Pulp system with `no tasks` executing, and AMP broker `stopped`.

  1. Create, and sync a repository.
  2. Check for status of created task. It should not have any task in the `waiting` state.

Closes: #650


Ichimonji10:
Rebased as 9e5c54d6db22b878e2e93648d44a56e6d3e896bc with a follow-up commit of 0858d27be0d7115b8f825bc459c1314b2368d703. Nice work.

PR #786, opened 2017-10-11
Test whether copy of directory will copy the mtime.

kersommoura:
* Test:

  1. Create and sync a repository.
  2. Publish the repository with the option `generate_sqlite` as True.
  3. Verify the `mtime` for the sqlite files.
  4. Upload a RPM package to the repository.
  5. Sync the repository again.
  6. Assert that `mtime` for old sqlite files remain the same.

### DIFF
--- a/pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
+++ b/pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
@@ -1,0 +1,93 @@
+# coding=utf-8
+"""Test whether copy of directory will copy the ``mtime``."""
+import os
+import re
+import time
+import unittest
+
+from pulp_smash import api, cli, config, selectors, utils
+from pulp_smash.constants import (
+    REPOSITORY_PATH,
+    RPM_SIGNED_URL,
+    RPM_UNSIGNED_FEED_URL,
+)
+from pulp_smash.tests.rpm.api_v2.utils import gen_distributor, gen_repo
+from pulp_smash.tests.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+_PATH = '/var/lib/pulp/published/yum/https/repos/'
+
+
+class CopyDirectoryStepTestCase(unittest.TestCase):
+    """Test whether copy of directory will copy the ``mtime``."""
+
+    def test_all(self):
+        """Test whether copy of directory will copy the ``mtime``.
+
+        This test targets the following issues:
+
+        * `Pulp Smash #720 <https://github.com/PulpQE/pulp-smash/issues/720>`_
+        * `Pulp #2783 <https://pulp.plan.io/issues/2783>`_
+
+        Do the following:
+
+        1. Create and sync a repository.
+        2. Publish the repository with the option ``generate_sqlite``
+           as True.
+        3. Verify the ``mtime`` for the sqlite files.
+        4. Upload a RPM package to the repository.
+        5. Sync the repository again.
+        6. Assert that ``mtime`` for old sqlite files remain the same.
+        """
+        cfg = config.get_config()
+        if selectors.bug_is_untestable(2783, cfg.version):
+            self.skipTest('https://pulp.plan.io/issues/2783')
+        client = api.Client(cfg, api.json_handler)
+        body = gen_repo()
+        body['importer_config']['feed'] = RPM_UNSIGNED_FEED_URL
+        body['distributors'] = [gen_distributor()]
+        repo = client.post(REPOSITORY_PATH, body)
+        self.addCleanup(client.delete, repo['_href'])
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
```diff
@@ -0,0 +1,93 @@
...
+        * `Pulp Smash #720 <https://github.com/PulpQE/pulp-smash/issues/720>`_
+        * `Pulp #2783 <https://pulp.plan.io/issues/2783>`_
+
+        Do the following:
+
+        1. Create and sync a repository.
+        2. Publish the repository with the option ``generate_sqlite``
+           as True.
+        3. Verify the ``mtime`` for the sqlite files.
+        4. Upload a RPM package to the repository.
```
omaciel: Is this a new package or an existing package already synced?

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
```diff
@@ -0,0 +1,93 @@
...
+        5. Sync the repository again.
+        6. Assert that ``mtime`` for old sqlite files remain the same.
```
omaciel: Is it necessary to re-publish? Sorry to ask, but I'm more familiar with Satellite terms and flow :)

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
kersommoura: It is a different package. I used as feed url a unsigned package, and to upload a signed package.

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
kersommoura: It was not required as part of steps to reproduce the issue. 
Sync in pulp will download the contents of the remote repository and store in our new repository. It is connected to the importer.
Publish will make available for the end user this new content. It is related to the distributor.
I will test if there is any difference, re-publishing before making the assertions.

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
```diff
@@ -0,0 +1,93 @@
...
+            'id': repo['distributors'][0]['id'],
+            'override_config': {'generate_sqlite': True},
+        })
+        repo = client.get(repo['_href'], params={'details': True})
+        path = os.path.join(
+            _PATH,
+            repo['distributors'][0]['config']['relative_url'],
+            'repodata'
+        )
+        system = cfg.get_systems('pulp cli')[0]
```
Ichimonji10: FYI, the `pulp cli` role indicates that Pulp's CLI tools are installed. In other words, the role indicates that `pulp-admin` is available. It doesn't indicate that shell access is available. The `shell` role indicates that shell access is available.

The host that fulfills the `api` role runs Apache, and it's therefore likely to also have access to the sqlite files relevant to this test.

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
```diff
@@ -0,0 +1,93 @@
...
+        sudo = () if utils.is_root(cfg, system) else ('sudo',)
+        cli_client = cli.Client(cfg, cli.echo_handler)
+        responses = cli_client.run(sudo + (
+            'ls', '-l', '--full-time', path
```
Ichimonji10: `ls` produces human-readable output. It's output isn't consistent over time or over different distributions, and therefore, it's output shouldn't be parsed by programming tools. Something like `stat --printf %Y /path/to/file` is more appropriate.

pulp_smash/tests/rpm/api_v2/test_copy_mtime.py
```diff
@@ -0,0 +1,93 @@
...
+        body = gen_repo()
+        body['importer_config']['feed'] = RPM_UNSIGNED_FEED_URL
+        body['distributors'] = [gen_distributor()]
+        repo = client.post(REPOSITORY_PATH, body)
+        self.addCleanup(client.delete, repo['_href'])
+        repo = client.get(repo['_href'], params={'details': True})
+        utils.sync_repo(cfg, repo)
+        utils.publish_repo(cfg, repo, {
+            'id': repo['distributors'][0]['id'],
+            'override_config': {'generate_sqlite': True},
```
Ichimonji10: If `generate_sqlite` can be set when the distributor is created, then that would be a more clean and concise way of doing things.

Ichimonji10:
Rebased and merged as 29cf01b7489ca4c96bfdd9d6c641a578df2a933f, with a follow-up commit of 8157631db1ffbf86d0c10530198b69f418c8a018.

PR #789, opened 2017-10-13
Remove format string and split in the pulp_admin_login function.

kersommoura:
* In the utils module, `pulp_admin_login` function, it was using `'...'.format().split()`
what can produce unexpected results in case of whitespace sneaks in.
To avoid this situation the function was refactored, and the use of `.format` followed by `.split` was removed.


kersommoura:
@Ichimonji10, there are other few cases making use of  `'...'.format().split()`, but I considered this re-factored one, the most important one. Since this function is called in many other tests.

Ichimonji10:
:100:

PR #790, opened 2017-10-13
Add a new section related to Pulp3.

kersommoura:
Why?

* Pulp-Smash will be used to test different versions major releases of Pulp.
  Right now, there are 2 main Pulp X releases: Pulp2 and Pulp3.

* Since both releases will be tested using Pulp-Smas, it is convenient to separate tests, and files
  related to a specific release. This will help address different needs, and run different test in a relative
  easy manner.

* Add tests related to authentication, using `BasicAuth` and `JWT`.

* These tests will part of smoke test of Pulp3 , however there are still a lot of changes to adjust Pulp-Smash to fit all the needs of Pulp3.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -18,6 +18,12 @@ developers, not a gospel.
     api/pulp_smash.exceptions
     api/pulp_smash.pulp_smash_cli
     api/pulp_smash.selectors
+    api/pulp_smash.pulp3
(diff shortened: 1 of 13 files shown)


Ichimonji10:
Merged with a follow-up commit of 6bc45e788d83714e1effdbacb2d0253e36c9fd25, as per our conversation.

PR #794, opened 2017-10-19
Let tests skip based on Pulp's version

Ichimonji10:
Update the existing `setUpModule` functions so that they cause tests to
skip if the incorrect version of Pulp is under test. Tests in
`pulp_smash.tests.pulp3` only execute when Pulp 3 is under test, and
other tests (which will eventually be moved to `pulp_smash.tests.pulp2`)
only execute when Pulp 2 is under test.


Ichimonji10:
@kersommoura FYI

PR #798, opened 2017-10-25
Clean up usages of uuid4() function

Ichimonji10:
The `pulp_smash.utils.uuid4()` function is written with Python 2 in
mind. Given that Pulp Smash now requires Python 3, that
backwards-compatibility logic may be dropped. Do so.

Where appropriate, use `pulp_smash.utils.uuid4()` instead of
`uuid.uuid4()` from the standard library.

### DIFF
--- a/pulp_smash/tests/pulp2/docker/api_v2/test_tags.py
+++ b/pulp_smash/tests/pulp2/docker/api_v2/test_tags.py
@@ -3,7 +3,6 @@
 import hashlib
 import random
 import unittest
-import uuid
 from urllib.parse import urljoin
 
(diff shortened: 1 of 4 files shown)


Ichimonji10:
Tested by executing the following with a collection of Pulp 2 hosts:

```console
$ python -m unittest pulp_smash.tests.platform.api_v2.test_search
ss............
Ran 12 tests in 4.621s

OK (skipped=2)
$ python -m unittest pulp_smash.tests.rpm.cli.test_environments
.
Ran 1 test in 22.365s

OK
```

PR #802, opened 2017-10-30
Let user specify a port number in config file

Ichimonji10:
Expand the schema for the Pulp Smash configuration file, and allow users
to set a port number. (The config file typically lives at
`~/.config/pulp_smash/settings.json`.) By default, the port number to
use when communicating with Pulp's API is inferred. For example, if the
scheme is http, then Pulp Smash will communicate with Pulp's API on port
80. This commit makes it possible to override that behaviour. A user can
now specify a specific port number to use when communicating with Pulp's
API.

This is mainly useful for the Pulp 3 API tests. Currently, the API tests
have port number 8,000 hard-coded into them. This commit removes those
hard-coded port numbers.

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -13,7 +13,7 @@ import json
 import os
 import warnings
 from copy import deepcopy
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urlunsplit
 
 import jsonschema
 from packaging.version import Version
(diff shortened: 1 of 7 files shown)


Ichimonji10:
Tested against Pulp 3 on F25 and F26:

```console
(pulp-smash) [ichimonji10@pine:~]$ python -m unittest discover pulp_smash.tests
sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss........s.s.
Ran 12 tests in 18.423s

OK (skipped=77)
```

I also ran some simple Pulp 2 tests as a sanity check, to ensure that they aren't broken by this change.

Ichimonji10:
See: https://github.com/pulp/pulp_packaging/pull/445

kersommoura:
@Ichimonji10, very helpful change. I read your comments and code. I will try to run them as well.

Ichimonji10:
Glad to hear it.

I executed the tests in all of the affected modules against a  matrix of six test systems. They all passed.

PR #805, opened 2017-11-01
Drop outdated property `version`

Ichimonji10:
This property was useful during a transition period some time back when
the config file format changed. It's no longer needed.


Ichimonji10:
Take two. `git grep '\.version'` only returns references to `packaging.version`.

kersommoura:
There was a total of 191 cfg.version before the change, now there is a total of cfg.pulp_vesion after the change.

PR #808, opened 2017-11-01
Verify vendor info in case of sync/upload of RPM. 

kersommoura:
There are 2 tests:

First:

 1. Create a repository then upload content with **vendor information**
    to this repository.
 2. Assert that vendor information is present and according to the data
    that was uploaded.

Second:

 1. Create a repository then sync content with **vendor information**
    to this repository.
 2. Assert that vendor information is present and according to feed URL.

Closes: #680

### DIFF
--- a/pulp_smash/constants.py
+++ b/pulp_smash/constants.py
@@ -655,9 +655,15 @@ RPM_WITH_NON_UTF_8_URL = urljoin(
 )
 """The URL to an RPM with non-UTF-8 metadata in its header."""
 
-RPM_WITH_VENDOR_URL = urljoin(
+RPM_WITH_VENDOR_FEED_URL = urljoin(
     PULP_FIXTURES_BASE_URL,
-    'rpm-with-vendor/rpm-with-vendor-1-1.fc25.noarch.rpm'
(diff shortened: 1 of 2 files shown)


pulp_smash/constants.py
```diff
@@ -655,6 +655,12 @@
...
 """The URL to an RPM with non-UTF-8 metadata in its header."""
 
+RPM_WITH_VENDOR_FEED_URL = urljoin(
+    PULP_FIXTURES_BASE_URL,
+    'rpm-with-vendor/'
+)
+"""A repository whose primary.xml file has an vendor section."""
+
 RPM_WITH_VENDOR_URL = urljoin(
     PULP_FIXTURES_BASE_URL,
```
Ichimonji10: Can this constant be derived from `RPM_WITH_VENDOR_FEED_URL` instead of `PULP_FIXTURES_BASE_URL`?

kersommoura: Done.

Ichimonji10:
Merged with a follow-up commit of ad43ecd01264a6caedaa823b568f54975715cb63.

Issue #748, opened 2017-08-14
Authentication - As an authenticated user I can manage user(s)


kersommoura:
[MVP Authentication](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Authentication)
-     Add a user
-     View user(s)
-     Update any user details
-     Delete a user

~~-     When deleting a user 'foo', all of user 'foo's existing JWTs are invalidated.~~
~~-     Invalidate a user's JWTs in the same operation as updating the password.~~

**At the time of creating this issue, these taxed items were not confirmed in the MVP document.**

kersommoura:
A full description of each of the fields can be retrieved with this call:
`http://pulp3.dev:3000/api/v3/users/`

```
"actions": {
        "POST": {
            "_href": {
                "type": "field",
                "required": false,
                "read_only": true,
                "label": " href"
            },
            "username": {
                "type": "string",
                "required": true,
                "read_only": false,
                "label": "Username",
                "help_text": "Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only."
            },
            "is_superuser": {
                "type": "boolean",
                "required": false,
                "read_only": false,
                "label": "Is superuser",
                "help_text": "Designates that this user has all permissions without explicitly assigning them."
            },
            "password": {
                "type": "string",
                "required": true,
                "read_only": false,
                "label": "Password",
                "help_text": "Password"
            }
        }
    }

```

kersommoura:
**Add a user**
`http://pulp3.dev:3000/api/v3/users/`
```
{
    "_href": "http://pulp3.dev:3000/api/v3/users/test/",
    "username": "test",
    "is_superuser": true
}
```

kersommoura:
Add test to verity that.

> As an authenticated user, I can filter users by:
> username: (equality, username_in_list)

[3142](https://pulp.plan.io/issues/3142)

PR #810, opened 2017-11-10
Add CRUD users tests.

kersommoura:
Add CRUD user tests for `Pulp3`.

See: #748

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -131,6 +131,7 @@ developers, not a gospel.
     api/pulp_smash.tests.pulp3.pulpcore.api_v3
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_auth
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_repos
+    api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_users
     api/pulp_smash.tests.pulp3.pulpcore.utils
     api/pulp_smash.tests.pulp3.utils
     api/pulp_smash.utils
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/pulp3/pulpcore/api_v3/test_crud_repos.py
```diff
@@ -43,7 +43,7 @@ def test_02_read_repo(self):
     def test_02_read_repos(self):
         """Search for the repository by its name.
 
-        Assert that just one search result is returned, and that that result
+        Assert that just one search result is returned, and that result
```
Ichimonji10: This is a nice grammar improvement, but it's not related to the rest of this test. While we're at it, it'd be good to change the line wrapping:

    Assert that just one search result is returned, and that result has a
    correct name.

kersommoura: @Ichimonji10 , done. If there is any minor update in the future will be in a separate PR.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_crud_repos.py
```diff
@@ -43,8 +43,8 @@ def test_02_read_repo(self):
     def test_02_read_repos(self):
         """Search for the repository by its name.
 
-        Assert that just one search result is returned, and that that result
-        has a correct name.
+        Assert that just one search result is returned, and that result has a
```
Ichimonji10: The point I was driving at in my previous comment is that this change shouldn't be made as part of this PR. This change is logically unrelated from the rest of this PR, and should be submitted as a separate PR.

Speaking of grammar, this change actually makes less grammatical sense. If you'd like to change this sentence to be less redundant, use a construct like "that the result has a correct name," not "that result has a correct name."

Ichimonji10:
Merged with a follow-up commit of c9fa2f7a02aea59e0dfc123f079e5dde1ac95da8.

Issue #751, opened 2017-08-14
Authentication, misc


kersommoura:
[MVP Authentication](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Authentication)

~~- Disable JWT token expiration. This configuration is in the settings file and is system-wide.~~
~~- Configure the JWT tokens to expire after a configurable amount of time. This configuration is in the settings file and is system-wide.~~

The JWT shall have a username identifier.

**At the time of creating this issue, these taxed items were not confirmed in the MVP document.**

bmbouter:
Both basic auth and JWT auth are merged. You can see their docs here: http://docs.pulpproject.org/en/3.0/nightly/integration_guide/rest_api/authentication.html

Ichimonji10:
@kersommoura I think that this issue has been resolved. The relevant tests are documented [here](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_auth.html). Do you agree? Can we close this issue?

kersommoura:
@Ichimonji10, I was waiting to have the full documentation in place to be able to confirm that we covered all the test cases. There are few other items listed in the MVP, and still not defined in the current version of documentation.

[Authentication](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Authentication)

For instance:

> As an API user, I can authenticate any API call (except to request a JWT) with a JWT. (not certain if this should be the behavior) [in progress]
> 
> As an API user, I can invalidate all existing JWT tokens for a given user. [done]
> 
> As an authenticated user, when deleting a user 'foo', all of user 'foo's existing JWTs are invalidated. [done]
>

Ichimonji10:
That makes sense. There's authentication-related topics that we aren't testing.

Ichimonji10:
We now have good test coverage for JWT tokens. The only line in the MVP that we aren't testing is this one:

> The JWT shall have a username identifier [done]

As it turns out, JWT tokens are a little tricky to decode. We could either dig into how JWT tokens and write some custom code to do this (it involves a little bit of base64 encoding/decoding trickery and possibly a little bit of encryption trickery), or we could add a dependency on a relevant library, like pyjwt, python-jose or pycrypto.

[...]

PR #811, opened 2017-11-13
Add tests to reset JWT.

kersommoura:
Pulp3 - Add tests related to reset JWT, and invalidate secret keys.

Tests:
 1. Verify whether resetting a secret key will invalidate all tokens of a certain use.
 2. Verify whether deleting a user the user`s token will be invalidate as well.

See: #751


Ichimonji10:
Merged, with a follow-up of e708250bcde788f6446cc94f0e444a5246abf95f.

PR #812, opened 2017-11-14
Remove duplicate word in docstring.

kersommoura:
Remove the use of the word `that` twice in a row in a docstring. Besides to be grammatically correct, it can be misleading.


Ichimonji10:
Thank you.

PR #813, opened 2017-11-15
Expand user CRUD tests

Ichimonji10:
Do the following:

* Assert that user passwords are never returned when searching for users
  or getting info about a particular user. Formerly, the tests simply
  ignored passwords.
* Add a test that searches for users.
* Fix a method name in `UsersCRUDTestCase`.

See: https://github.com/PulpQE/pulp-smash/issues/748


Ichimonji10:
Tested against Pulp 3 installations on F25 and F26.

I think that, with this merged, we will have fully addressed #748.

kersommoura:
@Ichimonji10 , I think that with these changes we can close #748.

Issue #754, opened 2017-08-14
Task Management - As an authenticated user, I can


kersommoura:
```

-  List all tasks 

    Filtering support on ['state', 'id', 'group']
    This does not include associated progress reports

```

```
- See a detail view for a specific task

    All attributes of a task
    All associated progress reports
```
`- Cancel a task`

kersommoura:
Schema for tasks.
```
 "tasks": {
        "list": {
            "_type": "link",
            "url": "/api/v3/tasks/",
            "action": "get",
            "fields": [
                {
                    "name": "cursor",
                    "location": "query",
                    "schema": {
                        "description": "The pagination cursor value.",
                        "_type": "string",
                        "title": "Cursor"
                    }
                },
                {
                    "name": "state",
                    "location": "query",
                    "schema": {
                        "description": "",
                        "_type": "string",
                        "title": ""
                    }
                },
                {
                    "name": "worker__name",
                    "location": "query",
                    "schema": {
                        "description": "",
                        "_type": "string",
                        "title": ""
                    }
                },
                {
                    "name": "tags",
                    "location": "query",
                    "schema": {
                        "description": "Multiple values may be separated by commas.",
                        "_type": "string",
                        "title": ""
                    }
                }
            ]
        },
        "read": {
            "_type": "link",
            "url": "/api/v3/tasks/{id}/",
            "action": "get",
            "fields": [
                {
                    "name": "id",
                    "required": true,
                    "location": "path",
                    "schema": {
                        "description": "A UUID string identifying this task.",
                        "_type": "string",
                        "title": "id"
                    }
                }
            ]
        }
    }
```

asmacdo:
Also make sure that Tasks can be deleted as long as they are in final states NOT("running", "waiting")

PR #814, opened 2017-11-16
Add tests to perform operations over tasks.

kersommoura:
Test:
 1. Create and partially update a repository.
 2. List current tasks.
 3. Read a specific task.

See: #754

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -132,6 +132,7 @@ developers, not a gospel.
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_auth
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_repos
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_users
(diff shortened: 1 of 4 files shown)


Ichimonji10:
What does the module introduced by this PR do? Does it test whether repository CRUD works? Does it verify that GET calls to `TASKS_PATH` work? Does it verify that GET calls to individual tasks work?

1. We already have tests for repositories in `pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_repos`. So, that isn't necessary.
2. We don't have tests that GET `TASKS_PATH`. However, a test for this feature should be more fully fleshed out. It should do things like filter tasks on "state," "id" and "group."
3. Get information about a specific task. However, if asynchronous task polling is implemented, then we will have effectively tested this feature.

The module introduced by this PR lacks focus. Please re-work it or close it.

kersommoura:
@Ichimonji10 , this module was created based on the current version of documentation.
My idea was to add basic coverage for the task system, and update in a near future when the full documentation is available.

PR #818, opened 2017-11-28
Refactor tests for process recycling

Ichimonji10:
The test case in `pulp_smash.tests.pulp2.rpm.cli.test_process_recycling`
sets and unsets `PULP_MAX_TASKS_PER_CHILD` in
`/etc/default/pulp_workers`, and verifies that Pulp respects this option
when starting up. Unfortunately, this test case frequently fails on
Fedora 26, during step four. It's unclear why this test fails, and even
worse, the failure is a heisenbug.

Refactor this module:

* Expand the debugging messages in the assertions. This should make it
  easier to understand whether or not the test failures are significant,
  or if they're insignificant, such as due to a dying process living too
  long.
* Deduplicate clean-up logic in the test case, and make clean-up more
  robust in the case of unexpected failures.
* Drop a function for restarting services, in favor of function
  `GlobalServiceManager.restart`.
* Make "sudo" logic more concise.
* Pre-split strings, to ensure no whitespace surprises.

### DIFF
--- a/pulp_smash/tests/pulp2/rpm/cli/test_process_recycling.py
+++ b/pulp_smash/tests/pulp2/rpm/cli/test_process_recycling.py
@@ -14,20 +14,6 @@ def setUpModule():  # pylint:disable=invalid-name
     utils.pulp_admin_login(config.get_config())
 
 
-def restart_pulp(cfg):
-    """Restart all of Pulp's services.
-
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against Pulp 2.14 and master on F24, F25, F26 and RHEL 7.

Issue #784, opened 2017-10-09
Automate https://pulp.plan.io/issues/3047


preethit:
https://pulp.plan.io/issues/3047
1. Create 2 two repos and sync them both. I recommend using file paths as feeds.
2. Edit an iso in the repo and make sure you update the PULP_MANIFEST file
3. Re-sync repo 1
4. Re-sync repo 2

preethit:
https://pulp.plan.io/issues/3100
This is another sat issue with 2 cases. Could we prioritize this as well? @Ichimonji10 @kersommoura

PR #819, opened 2017-11-29
Verify how ISO repos handlers update in content.

kersommoura:
Verify how ISO repos handles changes in content already in Pulp.

Test:

 1. Create and sync an ISO repository.
 2. Update the repository's feed URL, and sync it. This simulates a
    change in the contents of the source ISOs.
 3. Assert that number of units remain the same, but the content has changed.

Closes: #784

### DIFF
--- a/pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
+++ b/pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
@@ -2,16 +2,18 @@
 """Test CRUD for ISO RPM repositories."""
 import csv
 import json
+import unittest
 from unittest import SkipTest
 from urllib.parse import urljoin, urlparse
 
 import requests
 from packaging.version import Version
 
-from pulp_smash import api, exceptions, selectors, utils
+from pulp_smash import api, config, exceptions, selectors, utils
 from pulp_smash.constants import (
     FILE_FEED_URL,
     FILE_MIXED_FEED_URL,
+    FILE2_FEED_URL,
 )
 from pulp_smash.tests.pulp2.constants import REPOSITORY_PATH
 from pulp_smash.tests.pulp2.rpm.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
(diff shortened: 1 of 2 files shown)


pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        * `Pulp #2773 <https://pulp.plan.io/issues/2773>`_
+        * `Pulp #3047 <https://pulp.plan.io/issues/3047>`_
+        * `Pulp #3100 <https://pulp.plan.io/issues/3100>`_
+
+        Do the following:
+
+        1. Create and sync an ISO repository.
+        2. To recreate the scenario of change in units already in Pulp. After
+           the repository is synced, the feed url will be updated, and the
+           repository will be synced again.
```
Ichimonji10: The reader already knows that this is happening after the repository has been synced, because step one says that the repository has been synced. Something like this would read more nicely:

> 2. Update the repository's feed URL, and sync it. This simulates a change in the contents of the source ISOs.

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        3. Assert that number of units remain the same, but the content has
+           changed.
+        """
+        cfg = config.get_config()
+        if selectors.bug_is_untestable(2773, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/2773')
+        if selectors.bug_is_untestable(3047, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/3047')
+        if selectors.bug_is_untestable(3100, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/3100')
```
Ichimonji10: A for loop would make this skipping logic more concise:

```python
for issue_id in (2773, 3047, 3100):
    if selectors.bug_is_untestable(issue_id, cfg.pulp_version):
        self.skipTest('https://pulp.plan.io/issues/{}'.format(issue_id))
```

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        client = api.Client(cfg, api.json_handler)
+        repo = client.post(REPOSITORY_PATH, _gen_iso_repo(FILE2_FEED_URL))
```
Ichimonji10: This is a small thing, but I think it's more natural to initially sync the repository from `FILE_FEED_URL`, and later, sync the repository from `FILE2_FEED_URL`.

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        self.addCleanup(client.delete, repo['_href'])
+        repo = client.get(repo['_href'], params={'details': True})
+        utils.sync_repo(cfg, repo)
+        units = utils.search_units(cfg, repo)
```
Ichimonji10: In order to write good assertions, we need access to the units in the repository both before and after the re-sync. How about naming this variable something like `units_pre`?

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        units_before = len(units)
```
Ichimonji10: This is a single-use variable. How about dropping it and writing the following instead?

```python
self.assertEqual(len(units_pre), len(units_post))
```

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        iso_details = {
+            unit['metadata']['name']: unit['metadata']['checksum']
+            for unit in units
+        }
+        importer_config = {'importer_config': {'feed': FILE_FEED_URL}}
+        client.put(repo['importers'][0]['_href'], importer_config)
+        repo = client.get(repo['_href'], params={'details': True})
```
Ichimonji10: This read is unnecessary. `repo` is used by `sync_repo` and `search_units`, and both methods contain only simple references like this:

```python
urljoin(repo['_href'], 'actions/sync/')
```

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        if selectors.bug_is_untestable(3100, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/3100')
+        client = api.Client(cfg, api.json_handler)
+        repo = client.post(REPOSITORY_PATH, _gen_iso_repo(FILE2_FEED_URL))
+        self.addCleanup(client.delete, repo['_href'])
+        repo = client.get(repo['_href'], params={'details': True})
+        utils.sync_repo(cfg, repo)
+        units = utils.search_units(cfg, repo)
+        units_before = len(units)
+        iso_details = {
```
Ichimonji10: This dict contains only one detail about each ISO: their checksum. How about calling this something like `iso_checksums` or, even better, `unit_checksums`?

pulp_smash/tests/pulp2/rpm/api_v2/test_iso_crud.py
```diff
@@ -362,3 +364,70 @@ def test_invalid_file_feed(self):
...
+        utils.sync_repo(cfg, repo)
+        units = utils.search_units(cfg, repo)
+        units_after = len(units)
+
+        # Assert that the number of units has not changed.
+        self.assertEqual(units_before, units_after)
+
+        # Assert that content of units has changed.
+        for unit in units:
+            if unit['metadata']['name'] in iso_details:
```
Ichimonji10: What if the first search returned units named `1.iso`, `2.iso` and `3.iso`, and the second search returned units named `2.iso`, `3.iso` and `4.iso`? In that case, the `len()` assertion above would pass, and this `if` statement would hide the error. What's the benefit in allowing that?

kersommoura: @Ichimonji10, thanks for your review. It was updated per comments.

Basically we assure that number of units has not changed, that all the previous unit names remain in the repo, and their content has changed.

Ichimonji10:
Merged as is, with a follow-up of 3e9e322409b54f51100d3e0796af13e0de8edeaf. Nice work.

PR #820, opened 2017-12-01
Create dedicated namespace Pulp3 - platform.

kersommoura:
Pulp3 is composed by Pulp Core and Plugin API. To test the Plugin API
the File Plugin will be used.

Create a dedicated namespace called platform which one will be compose
by Pulp Core and File Plugin tests cases.

Future other plugins will have their dedicate namespace, thus avoiding
to mix what is part of Pulp as platform and extra plugins.


Ichimonji10:
We already have a namespace for Pulp 3 platform tests. It's `pulp_smash.tests.pulp3.pulpcore`. Other plugins will go into packages such as `pulp_smash.tests.pulp3.file`.

Issue #795, opened 2017-10-20
Adjust response_handlers for Pulp3 needs.


kersommoura:
As part of changes required to adjust Pulp-Smash to fit Pulp3 needs, it will be necessary to adjust the callback function, `response_handler`. These functions are used together with `api.Client` mainly.

Pulp2 and Pulp3 differ in the way the responses for certain actions are sent back. 

In Pulp3 the update of a data in a certain repository will generate the follow **202** response.

```json
{
    "_href": "http://pulp3.dev:8000/api/v3/tasks/4fc2e1fe-2564-4818-9b0d-4116a32ec2d0/",
    "task_id": "4fc2e1fe-2564-4818-9b0d-4116a32ec2d0"
}
```

Using the` api.json_handler` as response_handler to handle the previous response, it will generate an error, since it is not ready to parse this pattern of response. 

Error will occur in the function [poll_spawned_tasks](https://github.com/PulpQE/pulp-smash/blob/394d3510d052660d8ed378cc919861a18e2f7400/pulp_smash/api.py#L346), and [poll_task](https://github.com/PulpQE/pulp-smash/blob/394d3510d052660d8ed378cc919861a18e2f7400/pulp_smash/api.py#L371)

There are basically 2 choices, create other functions to handle Pulp3 response, or adapt the current ones to verify what version of Pulp is under test and perform different steps as needed.

Possibly changes to CLI Client`s callbacks will be required as well, but not identified at the moment.

Example Pulp 2 task:

```json
{
 "_href": "/pulp/api/v2/tasks/0fe4fcab-a040-11e1-a71c-00508d977dff/",
 "state": "running",
 "worker_name": "reserved_resource_worker-0@your.domain.com",
 "task_id": "0fe4fcab-a040-11e1-a71c-00508d977dff",
 "task_type": "pulp.server.tasks.repository.sync_with_auto_publish",
 "progress_report": {"contents vary": "depending on operation"},
 "result": null,
 "start_time": "2012-05-17T16:48:00Z",
 "finish_time": null,
 "exception": null,
 "traceback": null,
 "tags": [
   "pulp:repository:f16",
   "pulp:action:sync"
 ],
 "spawned_tasks": [{"href": "/pulp/api/v2/tasks/7744e2df-39b9-46f0-bb10-feffa2f7014b/",
                    "task_id": "7744e2df-39b9-46f0-bb10-feffa2f7014b" }],
 "error": null
}
```

Example Pulp 3 task:

```json
[...]

PR #821, opened 2017-12-06
Adapt task system to accept Pulp3 task system.

kersommoura:
Pulp3 task system still under development, but to be able to use the
current available reponse_handlers - mainly _handle_202 of API client,
few adjustments were created.

Adjusted functions: _check_tasks, _handle_202, poll_spawned_tasks and
poll_task.

Add test coverage to verify behaviour change in json_handler. Now
json_handler will treat in a different manner responses that status code
is equal to 204.

According to `HTTP/1.1 RFC 7231, Section 6.3.5`, 204 (No Content) status
code indicates that the server has sucessfully fullfiled the request and
that there is no additional content to send in response payload body.

See: #795

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -12,12 +12,13 @@ from time import sleep
 from urllib.parse import urljoin, urlparse
 
 import requests
+from packaging.version import Version
 
 from pulp_smash import exceptions
 
-
 _SENTINEL = object()
 _TASK_END_STATES = ('canceled', 'error', 'finished', 'skipped', 'timed out')
+_P3_TASK_END_STATES = ('canceled', 'completed', 'failed', 'skipped')
 
 
 def _check_http_202_content_type(response):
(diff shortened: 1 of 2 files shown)


kersommoura:
@Ichimonji10, I ran the updated code against Pulp2 functions that use api.json_handler, and in functions that have specific calls for [poll_spawned_tasks](https://github.com/PulpQE/pulp-smash/blob/b77fef5aa1845d35d712e072b9bc4ab973a8262e/pulp_smash/api.py#L346).  
Also ran against Pulp3, I refactored the test_crud_repos, an test_crud_users to use api.json_handler.
However, HTTP delete in Pulp3 will return a 204. 
Pulp3 task system still under development, and more changes will impact how to handle the response.

A more robust solution is to created dedicated response_handlers, and according to Pulp3 specificities. When they are defined.

pulp_smash/api.py
```diff
@@ -128,6 +133,8 @@ def json_handler(server_config, response):
     response body as JSON and return the result.
     """
     response.raise_for_status()
+    if response.status_code == 204:
+        return response
```
Ichimonji10: This is actually a pretty significant change in behaviour. Please add some test coverage:

```diff
diff --git a/tests/test_api.py b/tests/test_api.py
index 8b5b3fc..bc76133 100644
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -94,6 +94,14 @@ class JsonHandlerTestCase(unittest.TestCase):
             api.json_handler(**kwargs)
         self.assertEqual(handle_202.call_count, 1)
 
+    def test_204_check_run(self):
+        """Assert HTTP 204 responses are treated specially."""
+        kwargs = {key: mock.Mock() for key in ('server_config', 'response')}
+        kwargs['response'].status_code = 204
+        with mock.patch.object(api, '_handle_202'):
+            api.json_handler(**kwargs)
+        self.assertEqual(kwargs['response'].json.call_count, 0)
+
 
 class ClientTestCase(unittest.TestCase):
     """Tests for :class:`pulp_smash.api.Client`."""
```

kersommoura:
@Ichimonji10, thanks for your review. I updated as requested.

PR #826, opened 2017-12-13
Add test for Pulp 2788

Ichimonji10:
The first commit just does some refactoring. The second adds a test case for [Pulp 2788](https://pulp.plan.io/issues/2788).


kersommoura:
@Ichimonji10, as we talk. Adding a negative test case for ``remove_old_repodata`` as False will improve the test cases covered.

Ichimonji10:
Updated and re-tested.

kersommoura:
@Ichimonji10, with this new change, the test will behave as expected.

Ichimonji10:
Re-tested against beta 3. The test now passes.

Issue #656, opened 2017-05-27
Test syncing errata with large package lists


dkliban:
1. Create an RPM repository Foo with feed URL of https://repos.fedorapeople.org/pulp/pulp/demo_repos/large_errata/zoo/.
2. Create an RPM repository Bar with the same feed URL of https://repos.fedorapeople.org/pulp/pulp/demo_repos/large_errata/zoo/.
3. Sync repository Foo.
4. Sync repository Bar.
5. Assert that sync from step 4 finished successfully without error.

kersommoura:
@dkliban, can you clarify what do you mean by large errata? Is it the size of rpm package in bytes or the number of rpm packages per errata?

kersommoura:
Related redmine issue:

https://pulp.plan.io/issues/2681

PR #827, opened 2017-12-14
Add test to sync errata with large package list.

kersommoura:
Test:

 1. Create two repositories, with a feed URL that contains a large  errata package list.
 2. Sync both repositories.
 3. Assert that sync of second repository was finished without any error.

 Closes: #656


Ichimonji10:
Merged as cd519c0573430965a03ce782718359aa29ca6379 with a follow-up commit of e7d3e0d342ec02514714054ce3686b7fc7ea0afb.

PR #832, opened 2017-12-19
Prepend "sudo" to shell commands as needed

Ichimonji10:
`RemoveOldRepodataTestCase` executes several commands that require root
permissions. Prepend "sudo" to these commands when they're executed as a
non-root user.

See: fe6de191d6ea06b5f22cb96e0a07d136fb0a277c


Ichimonji10:
This should fix several new test failures.

kersommoura:
@Ichimonji10, I just ran the modified tests against a F26, and RHEL7 machines.

I installed the nightly version of Pulp2.15, and all tests passed.

Issue #752, opened 2017-08-14
Importer CRUD via REST API


kersommoura:
__As an authenticated user, I can CRUD importers with the REST API.__

[MVP Importers](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Importers)
**For the pulp3 release alpha the file plugin will be the only one available for tests.**

-     Create an importer
-     Read an importer
-     Update all mutable importer fields
-     Delete an importer (asynchronous)

```
-     Configure attributes on an Importer, such as:

    1.  validate (bool) [optional: defaults to True]: If true, the plugin will validate imported content.
    2.  ssl_ca_certificate (str) [optional] String containing a PEM encoded CA certificate used to validate the server certificate presented by the external source.
    3.  ssl_client_certificate (str) [optional] Contains a PEM encoded client certificate used for authentication.
    4.  ssl_client_key (str) [optional] Contains a PEM encoded private key used for authentication.
    5.  ssl_validation (bool) [optional: defaults to True]: If true, SSL peer validation must be performed.
    6.  proxy_url (str) [optional] Contains the proxy URL. Format: scheme://user:password@host:port.
    7.  username (str) [optional] Contains the username to be used for authentication when syncing.
    8.  password (str) [optional] Contains the password to be used for authentication when syncing.
    9.  download_policy (str) [optional] Contains the downloading policy name. This is a choice of three options:
      - immediate (default) - Downloading occurs during sync. The sync task does not complete until downloading is done.
      - background - Downloading is started by the sync but occurs in the background. The sync task completes before downloading is complete.
      - on-demand - The sync task records everything that would be downloaded but does not download content. Downloading occurs on demand as driven by client requests for content.
    10.  feed_url (str) [optional] Contains the URL of an external content source. This is optional.
    11.  sync mode: (str) [optional]. It has two choices 'additive' and 'mirror':
      - additive (default) - all remote content is added to the local repository on sync. During sync no content is removed from the local repository.
      - mirror - the local content will mirror the remote content exactly, removing local content if not also present in the remote content.
    12.  name (str) [required] Contains the name.
    13.  last_updated (datetime) [read-only] Contains the datetime of the last importer update.
    14.  last_synced (datetime) [read-only] Contains the datetime of the last importer sync.
```
~~15.  max_count (int) - ???~~
**At the time of creating this issue, this taxed item was not confirmed in the MVP document.**

kersommoura:
[...]

PR #835, opened 2017-12-20
 Add tests for CRUD importers Pulp3.

kersommoura:
To be able to CRUD importers a plugin of any kind must be installed.
These tests require the `File Plugin` to be installed to be able to run.

See: #752

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
@@ -1,0 +1,99 @@
+# coding=utf-8
+"""Tests that CRUD importers."""
+import unittest
+
+from requests.exceptions import HTTPError
+
+from pulp_smash import api, config, selectors
+from pulp_smash.tests.pulp3.constants import FILE_IMPORTER_PATH, REPO_PATH
+from pulp_smash.tests.pulp3.file.api_v3.utils import (
+    gen_importer,
+    modify_importer_down_policy,
+    modify_importer_sync_mode,
+)
+from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
+from pulp_smash.tests.pulp3.utils import get_auth
+
+
+class CRUDImportersTestCase(unittest.TestCase):
+    """CRUD importers."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables.
+
+        In order to create an importer a repository has to be created first.
+        """
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client.request_kwargs['auth'] = get_auth()
+        cls.importer = {}
+        cls.repo = cls.client.post(REPO_PATH, gen_repo())
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.repo['_href'])
+
+    @selectors.skip_if(bool, 'repo', False)
+    def test_01_create_importer(self):
+        """Create an importer."""
+        body = gen_importer()
+        body['repository'] = self.repo['_href']
+        type(self).importer = self.client.post(FILE_IMPORTER_PATH, body)
+
+    @selectors.skip_if(bool, 'importer', False)
+    def test_02_read_importer(self):
+        """Read an importer by its href."""
+        importer = self.client.get(self.importer['_href'])
+        self.assertEqual(self.importer['name'], importer['name'])
+
+    @selectors.skip_if(bool, 'importer', False)
+    def test_02_read_importers(self):
+        """Read an importer by its name."""
+        page = self.client.get(FILE_IMPORTER_PATH, params={
+            'name': self.importer['name']
+        })
+        self.assertEqual(len(page['results']), 1)
+        self.assertEqual(page['results'][0]['name'], self.importer['name'])
+
+    @selectors.skip_if(bool, 'importer', False)
+    def test_03_partially_update(self):
+        """Update an importer using HTTP PATCH."""
+        attr = {
+            'download_policy': modify_importer_down_policy(
+                {self.importer['download_policy']}
+            ),
+            'sync_mode': modify_importer_sync_mode(
+                {self.importer['sync_mode']}
+            )
+        }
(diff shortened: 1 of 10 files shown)


pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+        cls.client.request_kwargs['auth'] = get_auth()
+        cls.importer = {}
+        cls.repo = cls.client.post(REPO_PATH, gen_repo())
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.repo['_href'])
+
+    @selectors.skip_if(bool, 'repo', False)
```
Ichimonji10: This selector can be dropped. If `cls.client.post(REPO_PATH, gen_repo())` returns an error, then `setUpClass` will not successfully complete, and none of the `test*` methods will run.

pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+        """Create an importer."""
+        body = gen_importer()
+        body['repository'] = self.repo['_href']
+        type(self).importer = self.client.post(FILE_IMPORTER_PATH, body)
+
+    @selectors.skip_if(bool, 'importer', False)
+    def test_02_read_importer(self):
+        """Read an importer by its href."""
+        importer = self.client.get(self.importer['_href'])
+        self.assertEqual(self.importer['name'], importer['name'])
```
Ichimonji10: This test asserts that the attributes **returned** when the repository was created can be fetched again. It doesn't tell us that the attributes **sent** when the repository was created can be fetched again. This test is still useful as is, because it demonstrates that HTTP GET requests to importers don't blow up. But this test isn't as useful as it could be.

Note that `pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_users` is much better written than `pulp_smash.tests.pulp3.pulpcore.api_v3.test_repos`, for this reason.

Also, it would be even better if we could iterate through a bunch of attributes and verify that they're correctly set, instead of just spot-checking the name attribute.

pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+        self.assertEqual(
+            body['download_policy'],
+            self.importer['download_policy']
+        )
+        self.assertEqual(body['sync_mode'], self.importer['sync_mode'])
+
+    @selectors.skip_if(bool, 'importer', False)
+    def test_05_delete(self):
+        """Delete an importer."""
+        self.doCleanups()
```
Ichimonji10: `doCleanups()` calls all of the method calls that have been queued with `addCleanup`. But the queue of calls is empty here. The clean-up that was queued up in `test_04_fully_update` was executed when that test method finished.

pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+    def setUpClass(cls):
+        """Create class-wide variables.
+
+        In order to create an importer a repository has to be created first.
+        """
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client.request_kwargs['auth'] = get_auth()
+        cls.importer = {}
+        cls.repo = cls.client.post(REPO_PATH, gen_repo())
```
Ichimonji10: Having lots of persistent attributes usually makes it hard to keep track of what's going on. If there's some way to reduce the number of persistent variables, that would probably be good.

pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+from pulp_smash import api, config, selectors
+from pulp_smash.tests.pulp3.constants import FILE_IMPORTER_PATH, REPO_PATH
+from pulp_smash.tests.pulp3.file.api_v3.utils import (
+    gen_importer,
+    modify_importer_down_policy,
+    modify_importer_sync_mode,
+)
+from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
+from pulp_smash.tests.pulp3.utils import get_auth
+
```
Ichimonji10: This test will execute even if Pulp 2 is under test. It will also execute even if the file plug-in isn't installed. Checks like this must be implemented before this module can be merged.

pulp_smash/tests/pulp3/file/api_v3/test_crud_importers.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.repo['_href'])
+
+    @selectors.skip_if(bool, 'repo', False)
+    def test_01_create_importer(self):
+        """Create an importer."""
+        body = gen_importer()
+        body['repository'] = self.repo['_href']
```
Ichimonji10: Does a repository href need to be specified when creating an importer? If so, a `repo` or `repo_href` argument should probably be added to the `gen_importer` function.

Ichimonji10:
I took a stab at revising these tests as per my own comments, above. Please incorporate my changes into your work. See: https://github.com/Ichimonji10/pulp-smash/tree/pr-835

Note the two FIXME comments. I suggest reaching out to the developers to figure out why those errors are occurring. Perhaps credentials are intentionally omitted from GET requests for importers?

Ichimonji10:
I see you're going to be out on PTO for the rest of this week. I'll finish this.

Ichimonji10:
Merged as https://github.com/PulpQE/pulp-smash/commit/7a432cc4b9557b8eaacad764a9057c3891dc89dd, with a follow-up commit of https://github.com/PulpQE/pulp-smash/commit/986128bbb65c753a3ed7c66257d572aa6126e318. Also see https://github.com/PulpQE/pulp-smash/commit/185465396758f80824cdd694d27b39ead1d8bc2d.

PR #841, opened 2018-01-16
Skip tests affected by Pulp issue 3287

Ichimonji10:
This issue is wide-ranging:

> Pulp 2.15 nightlies broken on Fedora 27 due to a pulp_streamer import
> error

See: https://pulp.plan.io/issues/3287


Ichimonji10:
This PR may be unnecessary, as https://pulp.plan.io/issues/3287 appears to be a transient issue with the Fedora repositories, rather than an issue with Pulp itself.

kersommoura:
I took a look in this PR. If we decide to move forward, it is ready to be merged.
Based on the latest comments on the issue 3287, I do think that it will be a transient issue.

Ichimonji10:
I've updated the hosts I use for testing, and this issue no longer affects Fedora 27.

PR #849, opened 2018-01-24
Add rsync extra args test

simon-baatz:


Ichimonji10:
Excellent work.

The test appears to be functionally and semantically correct. To validate, I ran the test against a couple Pulp 2.15.1 nightly systems:

1. By default, the test skips. This is correct behaviour, as the corresponding Pulp PR hasn't been merged yet.
2. I commented out the `skipTest`, and the test failed. This is correct behaviour, for the same reason.
3. I commented out the `skipTest`, and the `--dry-run` publish and assertion, and the test succeeded. This is correct behaviour, as all other functionality is already in place.

Rebased and merged, with a follow-up commit of FIXME.

For basic information on follow-up commits, see the [review process](http://pulp-smash.readthedocs.io/en/latest/about.html#review-process) documentation. The goal is to get PRs merged faster and to reduce the amount of time spent fixing nitpicks, though at the cost of some extra junk in the commit log. If you'd rather fix nitpicky issues in the PR itself, let me know, and I'll treat future PRs differently.

Issue #839, opened 2018-01-05
Test that importer config can be updated susscessfully


preethit:
https://pulp.plan.io/issues/3210
Verify that we can 

    PUT pulp/api/v2/repositories/<id>/importers/<id>/

Ichimonji10:
This is done. See: https://github.com/PulpQE/pulp-smash/pull/835

PR #850, opened 2018-01-25
Test if importer config of ostree can be updated.

kersommoura:
Add tests to verify if the importer of ostree can be updated using HTTP
PUT.

Do the following:

 1. Create and sync a repository.
 2. Update the `importer` associated with the repository.
 3. Perform assertions about the just updated `importer`.

Closes: #839

### DIFF
--- a/pulp_smash/tests/pulp2/ostree/api_v2/test_crud.py
+++ b/pulp_smash/tests/pulp2/ostree/api_v2/test_crud.py
@@ -6,12 +6,15 @@
 .. _repositories:
    http://docs.pulpproject.org/en/latest/dev-guide/integration/rest-api/repo/cud.html
 """
+import unittest
+from random import choice, randrange
 from urllib.parse import urljoin
 
 from packaging.version import Version
 from requests.exceptions import HTTPError
 
-from pulp_smash import api, exceptions, selectors, utils
+from pulp_smash import api, config, exceptions, selectors, utils
+from pulp_smash.constants import OSTREE_FEED, OSTREE_BRANCHES
 from pulp_smash.tests.pulp2.constants import REPOSITORY_PATH
 from pulp_smash.tests.pulp2.ostree.utils import gen_repo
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/pulp2/ostree/api_v2/test_crud.py
```diff
@@ -326,3 +329,61 @@ def test_invalid_v3(self):
...
+        cfg = config.get_config()
+        if selectors.bug_is_untestable(3210, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/3210')
+        client = api.Client(cfg, api.json_handler)
+        body = gen_repo()
+        body['importer_config']['feed'] = OSTREE_FEED
+        body['importer_config']['branches'] = OSTREE_BRANCHES
+        repo = client.post(REPOSITORY_PATH, body)
+        self.addCleanup(client.delete, repo['_href'])
+        utils.sync_repo(cfg, repo)
```
Ichimonji10: If this can be dropped, please do so.

pulp_smash/tests/pulp2/ostree/api_v2/test_crud.py
```diff
@@ -326,3 +329,61 @@ def test_invalid_v3(self):
...
+            'depth': randint(0, 50),
+            'feed': utils.uuid4(),
+            'ssl_validation': choice((False, True)),
+        }
+        client.put(repo['importers'][0]['_href'], {
+            'importer_config': importer_config
+        })
+        repo = client.get(repo['_href'], params={'details': True})
+
+        self.assertEqual(
```
Ichimonji10: Use a loop here. For example:

```python
        for key, val in importer_config.items():
            with self.subTest(key=key):
                if val is None:
                    self.assertNotIn(key, repo['importers'][0]['config'])
                else:
                    self.assertEqual(val, repo['importers'][0]['config'][key])
```

kersommoura:
@Ichimonji10, your comments were addressed. Besides that, I was able to verify the issue using an older version of Pulp - 2.12.3a1, and the fix worked in newer versions of Pulp.

PR #854, opened 2018-01-30
Verify whether sync update repository version. 

kersommoura:
Do the following:

 1. Create a repository, and an importer.
 2. Sync the repository an arbitrary number of times.
 3. Verify that the repository version is equal to the previous number
    of syncs.

This test explores the design choice described [Sync of an unchanged
importer changes repository version](https://pulp.plan.io/issues/3308)

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
@@ -2,7 +2,8 @@
 """Tests that sync file plugin repositories."""
 import unittest
 from itertools import product
-from urllib.parse import urljoin
+from random import randint
+from urllib.parse import urljoin, urlsplit
 
 from pulp_smash import api, config
 from pulp_smash.constants import FILE_FEED_URL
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/pulp3/file/api_v3/test_sync.py
```diff
@@ -85,3 +86,70 @@ def do_test(self, download_policy, sync_mode):
...
+class SyncChangeRepoVersionTestCase(unittest.TestCase):
+    """Verify whether sync of repository updates repository version."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.cfg = config.get_config()
+
+    def test_all(self):
+        """Call :meth:`do_test` with varying arguments.
```
Ichimonji10: Is there any special reason to call `do_test` with each possible pairing of `download_policy` and `sync_mode`? Doing this will make the test case consume 2-6 times as much time. One could execute the test with any valid combination of `download_policy` and `sync_mode`, and still prove that repository versions correctly increment.

pulp_smash/tests/pulp3/file/api_v3/test_sync.py
```diff
@@ -85,3 +86,70 @@ def do_test(self, download_policy, sync_mode):
...
+        importer = client.post(FILE_IMPORTER_PATH, body)
+        self.addCleanup(client.delete, importer['_href'])
+
+        number_of_syncs = randint(1, 10)
+        for _ in range(number_of_syncs):
+            sync_repo(self.cfg, importer)
+
+        repo = client.get(repo['_href'])
+        path = urlsplit(repo['_latest_version_href']).path
+        latest_repo_version = path.split('/')[-2]
```
Ichimonji10: Cast this string to an integer. Without a type cast, the `assertEqual` below is comparing a string to an integer. I'm sure the semantics of doing so are defined by `assertEqual`, but why delve into that territory?

kersommoura:
@Ichimonji10, thanks for your review. All your comments were addressed.

pulp_smash/tests/pulp3/file/api_v3/test_sync.py
```diff
@@ -85,3 +86,43 @@ def do_test(self, download_policy, sync_mode):
...
+        2. Sync the repository an arbitrary number of times.
+        3. Verify that the repository version is equal to the previous number
+           of syncs.
+        """
+        cfg = config.get_config()
+        client = api.Client(cfg, api.json_handler)
+        client.request_kwargs['auth'] = get_auth()
+        repo = client.post(REPO_PATH, gen_repo())
+        self.addCleanup(client.delete, repo['_href'])
+        body = gen_importer(repo)
```
Ichimonji10: This function needs to return importers that have valid download policies before this test can be merged.

kersommoura: Logic related to download policy was added to [get_importer_down_policy](https://github.com/PulpQE/pulp-smash/blob/14c9dc7a1a0edb21bf91904d92b9431c06505e70/pulp_smash/tests/pulp3/file/api_v3/utils.py#L12).

Issue #565, opened 2017-02-26
Improve pulp_smash.api.poll_task


elyezer:
We've seen some cases where for some unknown reason the poll_task won't never timeout. In order to avoid that we could improve the poll_task to use a threading.Timer like:

```python
def poll_task(server_config, href):
    """Wait for a task and its children to complete. Yield response bodies.

    Poll the task at ``href``, waiting for the task to complete. When a
    response is received indicating that the task is complete, yield that
    response body and recursively poll each child task.

    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
    :param href: The path to a task you'd like to monitor recursively.
    :returns: An generator yielding response bodies.
    :raises pulp_smash.exceptions.TaskTimedOutError: If a task takes too
        long to complete.
    """
    timeout = 1800  # 1800s == 30m
    timer = threading.Timer(timeout, thread.interrupt_main)
    try:
        timer.start()
        while True:
            response = requests.get(
                urljoin(server_config.base_url, href),
                **server_config.get_requests_kwargs()
            )
            response.raise_for_status()
            attrs = response.json()
            if attrs['state'] in _TASK_END_STATES:
                # This task has completed. Yield its final state, then iterate
                # through each of its children and yield their final states.
                yield attrs
                for href in (task['_href'] for task in attrs['spawned_tasks']):
                    for final_task_state in poll_task(server_config, href):
                        yield final_task_state
                break
            sleep(5)
    except KeyboardInterrupt:
        raise exceptions.TaskTimedOutError(
            'Task {} is ongoing after {} seconds.'.format(href, timeout)
        )
    finally:
        timer.cancel()
```

The above approach is fine but now on Python 3 the `thread` module was renamed to `_thread` to encourage the usage of the `threading` instead. The only usage of the `thread` here is to interrupt the main thread after the timeout.

Another approach is to use a for loop:

```python
def poll_task(server_config, href):
    """Wait for a task and its children to complete. Yield response bodies.

    Poll the task at ``href``, waiting for the task to complete. When a
    response is received indicating that the task is complete, yield that
    response body and recursively poll each child task.

    :param server_config: A :class:`pulp_smash.config.ServerConfig` object.
    :param href: The path to a task you'd like to monitor recursively.
    :returns: An generator yielding response bodies.
    :raises pulp_smash.exceptions.TaskTimedOutError: If a task takes too
        long to complete.
    """
    # 360 * 5s == 1800s == 30m
[...]

PR #855, opened 2018-02-02
Poll asynchronous tasks more frequently

Ichimonji10:
Poll asynchronous tasks every two seconds, instead of every five
seconds. Many of the asynchronous tasks that Pulp executes complete
within just a few polling cycles, such as five, ten or fifteen seconds.
As a result, the long (five second) polling periods mean that a
significant amount of Pulp Smash's time is spent waiting for tasks to
complete. Increasing the polling rate should reduce the amount of time
spent waiting by a significant amount.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -395,10 +395,10 @@ def poll_task(server_config, href, pulp_system=None):
     """
     if not pulp_system:
         pulp_system = server_config.get_systems('api')[0]
-    # 360 * 5s == 1800s == 30m
+    # 900 * 2s == 1800s == 30m
     # NOTE: The timeout counter is synchronous. We query Pulp, then count down,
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I ran the following modules:

* package `pulp_smash.tests.pulp2.platform.api_v2`
* module `pulp_smash.tests.pulp2.rpm.api_v2.test_rsync_distributor`

I ran these modules against installations of Pulp nightly (to-be 2.16) on:

* Fedora 25
* Fedora 26
* Fedora 27
* RHEL 7

I found that:

* The runtimes for the `platform` module hardly changed.
* The runtimes for the rsync distributor module decreased from roughly 230 seconds to 170 seconds.

kersommoura:
Related issue: [565](https://github.com/PulpQE/pulp-smash/issues/565)

PR #857, opened 2018-02-07
Change how booleans are set

Ichimonji10:
In Fedora 26 and earlier, and on RHEL 7:

* `semanage` sets the default and current state of an SELinux policy.
* `setsebool` sets the current state of an SELinux policy.

In Fedora 27:

* `semanage` sets the default state of an SELinux policy.
* `setsebool` sets the current state of an SELinux policy.

This change in behaviour breaks tests that use `semanage` to modify the
current state of SELinux policies. Switch to using `setsebool`.

See: https://pulp.plan.io/issues/3347


Ichimonji10:
Tested against a matrix  of hosts, where the axes are:

* Pulp 2.15 stable, and Pulp 2.16 nightly
* Fedora 25, Fedora 26, Fedora 27, and RHEL 7

Issue #776, opened 2017-09-27
Distributions CRUD via REST API


asmacdo:
__As an authenticated REST API user, I can CRUD distributions__

Similar basic CRUD tests. Repositories, Importers, and Distributions all behave the same way with CRUD (synchronous Create and Read, async Update and Delete).

asmacdo:
`http OPTIONS http://pulp3.dev:8000/api/v3/repositories/my-repo/publishers/file/my-pub/distributions/`
```
HTTP/1.0 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 1118
Content-Type: application/json
Date: Wed, 27 Sep 2017 23:22:41 GMT
Server: WSGIServer/0.2 CPython/3.5.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "actions": {
        "POST": {
            "_href": {
                "label": " href",
                "read_only": true,
                "required": false,
                "type": "field"
            },
            "auto_updated": {
                "help_text": "The publication is updated automatically when the publisher has created a new publication",
                "label": "Auto updated",
                "read_only": false,
                "required": true,
                "type": "boolean"
            },
            "base_path": {
                "help_text": "The base (relative) path component of the published url.",
                "label": "Base path",
                "read_only": false,
                "required": true,
                "type": "string"
            },
            "http": {
                "help_text": "The publication is distributed using HTTP.",
                "label": "Http",
                "read_only": false,
                "required": true,
                "type": "boolean"
            },
            "https": {
                "help_text": "The publication is distributed using HTTPS.",
                "label": "Https",
                "read_only": false,
                "required": true,
                "type": "boolean"
            },
            "name": {
                "help_text": "The name of the distribution. Ex, `rawhide` and `stable`.",
                "label": "Name",
                "read_only": false,
                "required": true,
                "type": "string"
            },
            "publisher": {
                "label": "Publisher",
                "read_only": true,
                "required": false,
                "type": "field"
            }
        }
    },
    "description": "",
    "name": "Distribution List",
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ],
    "renders": [
        "application/json",
        "text/html"
    ]
}
```

asmacdo:
Distributions are not master detail, but otherwise the expected responses should mirror 752 (comment)

There can be an arbitrary number of distributions for each publisher.

asmacdo:
[...]

PR #860, opened 2018-02-12
Add tests to CRUD distributions.

kersommoura:
Add tests to CRUD distributions. A distribution is not associated within
a plugin, then the related tests were added as part of `pulp_core`.

See: #776


kersommoura:

pulp-file==0.0.1a12
pulpcore==3.0.0a20
pulpcore-common==3.0.0a13
pulpcore-plugin==0.0.1a15

PR #863, opened 2018-02-14
Add test to filter user by its username.

kersommoura:
Expand user test cases. Add an extra test to filter user by its
username.

See: https://pulp.plan.io/issues/3142

Closes: #748


kersommoura:

pulp-file==0.0.1a12
pulpcore==3.0.0a20
pulpcore-common==3.0.0a13
pulpcore-plugin==0.0.1a15

PR #864, opened 2018-02-14
Modify assertions in filter tests for repositories.

kersommoura:
Modify assertions performed on filtered repositories to be more generic.
It will comply with the same style of assertions used in equivalent test
scenarios.


kersommoura:

pulp-file==0.0.1a12
pulpcore==3.0.0a20
pulpcore-common==3.0.0a13
pulpcore-plugin==0.0.1a15

PR #873, opened 2018-02-19
Add `pulp-smash smoke-tests` CLI command

Ichimonji10:
This command prints the dotted name of each smoke test in Pulp Smash.
Sample usage:

    python -m unittest $(pulp-smash smoke-tests)

### DIFF
--- a/pulp_smash/pulp_smash_cli.py
+++ b/pulp_smash/pulp_smash_cli.py
@@ -1,10 +1,11 @@
 # coding=utf-8
 """The entry point for Pulp Smash's command line interface."""
 import json
+import unittest
 
 import click
 
-from pulp_smash import config, exceptions
+from pulp_smash import config, exceptions, utils
 from pulp_smash.config import PulpSmashConfig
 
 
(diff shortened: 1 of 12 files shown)


Ichimonji10:
Tested against a pair of Pulp 3 applications, running on F26 and F27 respectively.

Ichimonji10:
Let me add some unit tests.

pulp_smash/pulp_smash_cli.py
```diff
@@ -196,5 +197,57 @@ def settings_validate(ctx):
...
+@pulp_smash.command('smoke-tests')
+def smoke_tests():
+    """Print smoke tests, one per line.
+
+    Sample usage:
+
+        python -m unittest $(pulp-smash smoke-tests)
+    """
+    test_suite = unittest.TestSuite()
+    test_suite.addTests(unittest.TestLoader().discover('pulp_smash.tests'))
```
kersommoura: @Ichimonji10, Can we restricted the discover just for Pulp3 namespace?

Ichimonji10: What's the benefit in doing so?

kersommoura: It is just matter to avoid to parse the whole Pulp2 namespace looking for smoke tests, since right now we do not have any.  Besides that, I do not see any other.

Ichimonji10: I'd like to search the entire Pulp namespace by default. If efficency becomes a concern, then we can expand the user interface, so that the user can execute commands like:

```bash
pulp-smash smoke-tests pulp_smash.tests.pulp2
pulp-smash smoke-tests pulp_smash.tests.pulp3
pulp-smash smoke-tests pulp_smash.tests.pulp3.pulpcore
```

I do expect that we will mark some Pulp 2 tests as being smoke tests. When that happens, it'd be nice if they just showed up in the output of `pulp-smash smoke-tests` without us having to fiddle with the CLI machinery.

kersommoura: @Ichimonji10, thanks.

Issue #870, opened 2018-02-19
As an authenticated user, I can create a Content unit from one or more artifacts


kersommoura:
Add test to explore creation of content units to Pulp3.

See issue [2872](https://pulp.plan.io/issues/2872)

In order to create a content unit, an artifact has to exist.

Then HTTP POST to `/api/v3/content/file/` < content.json

content.json
```json 
  {
    "digest": "a6ad136cd5f329e3843d67dc41176cb0391e04fe23d001d00c4ec2fbc3fd9bd7",
    "path": "8eca5b27-4641-48cd-af73-bd3e6bbc4d89",
    "artifacts": {
      "8eca5b27-4641-48cd-af73-bd3e6bbc4d89": "/var/lib/pulp/artifact/a6/ad136cd5f329e3843d67dc41176cb0391e04fe23d001d00c4ec2fbc3fd9bd7"
    }
  }

```

PR #892, opened 2018-03-05
Add tests to CRUD content unit.

kersommoura:
In order to create a content unit at least one artifact has to exist in
Pulp3, then metadata related to this artifact will be used to create a content
unit.

An artifact is just data on disk, a content unit is an entry in a
database that points to an artifact and gives it a name.

Closes:#870

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
@@ -1,0 +1,89 @@
+# coding=utf-8
+"""Tests that perform actions over content unit."""
+import unittest
+
+from requests.exceptions import HTTPError
+
+from pulp_smash import api, config, selectors, utils
+from pulp_smash.constants import FILE_URL
+from pulp_smash.tests.pulp3.constants import ARTIFACTS_PATH, FILE_CONTENT_PATH
+from pulp_smash.tests.pulp3.file.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+from pulp_smash.tests.pulp3.utils import clean_artifacts, get_auth
+
+
+class ContentUnitTestCase(unittest.TestCase, utils.SmokeTest):
+    """CRUD content unit.
+
+    This test targets the following issues:
+
+    * `Pulp #2872 <https://pulp.plan.io/issues/2872>`_
+    * `Pulp Smash #870 <https://github.com/PulpQE/pulp-smash/issues/870>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variable."""
+        cls.cfg = config.get_config()
+        clean_artifacts(cls.cfg)
+        cls.content_unit = {}
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client.request_kwargs['auth'] = get_auth()
+        files = {'file': utils.http_get(FILE_URL)}
+        cls.artifact = cls.client.post(ARTIFACTS_PATH, files=files)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.artifact['_href'])
+
+    def test_01_create_content_unit(self):
+        """Create content unit."""
+        attrs = _gen_content_unit_attrs(self.artifact)
+        self.content_unit.update(self.client.post(FILE_CONTENT_PATH, attrs))
+        for key, val in attrs.items():
+            with self.subTest(key=key):
+                self.assertEqual(self.content_unit[key], val)
+
+    @selectors.skip_if(bool, 'content_unit', False)
+    def test_02_read_content_unit(self):
+        """Read a content unit by its href."""
+        content_unit = self.client.get(self.content_unit['_href'])
+        for key, val in self.content_unit.items():
+            with self.subTest(key=key):
+                self.assertEqual(content_unit[key], val)
+
+    @selectors.skip_if(bool, 'content_unit', False)
+    def test_03_partially_update(self):
+        """Attempt to update a content unit using HTTP PATCH.
+
+        This HTTP method is not supported and a HTTP exception is expected.
+        """
+        attrs = _gen_content_unit_attrs(self.artifact)
+        with self.assertRaises(HTTPError):
+            self.client.patch(self.content_unit['_href'], attrs)
+
+    @selectors.skip_if(bool, 'content_unit', False)
+    def test_03_fully_update(self):
+        """Attempt to update a content unit using HTTP PUT.
+
+        This HTTP method is not supported and a HTTP exception is expected.
+        """
+        attrs = _gen_content_unit_attrs(self.artifact)
+        with self.assertRaises(HTTPError):
+            self.client.put(self.content_unit['_href'], attrs)
+
+    @selectors.skip_if(bool, 'content_unit', False)
+    def test_04_delete_content_unit(self):
+        """Delete a content unit."""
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,94 @@
...
+def _gen_verbose_content_unit_attrs(artifact, path):
+    """Generate a dict with content unit attributes.
+
+    :param: artifact: A dict about the artifact.
+    :param: path: The path in which one the content unit will be available.
+    """
+    return {
+        'artifacts': {path: artifact['file']},
+        'digest': artifact['sha256'],
+        'path': path
```
daviddavis: Just a heads up: these fields are changing. It's going to be just `artifact` and `relative_path`. See https://github.com/pulp/pulp_file/pull/51.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+    def setUpClass(cls):
+        """Create class-wide variable."""
+        cls.artifact = {}
+        cls.cfg = config.get_config()
+        clean_artifacts(cls.cfg)
+        cls.content_unit = {}
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client.request_kwargs['auth'] = get_auth()
+        files = {'file': utils.http_get(FILE_URL)}
+        cls.artifact = cls.client.post(ARTIFACTS_PATH, files=files)
```
Ichimonji10: This variable is also defined on line 26, above. Define it just once.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+    * `Pulp #2872 <https://pulp.plan.io/issues/2872>`_
+    * `Pulp Smash #870 <https://github.com/PulpQE/pulp-smash/issues/870>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variable."""
+        cls.artifact = {}
+        cls.cfg = config.get_config()
+        clean_artifacts(cls.cfg)
```
Ichimonji10: Does the test break if old artifacts are left in place?

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+        cls.client.request_kwargs['auth'] = get_auth()
+        files = {'file': utils.http_get(FILE_URL)}
+        cls.artifact = cls.client.post(ARTIFACTS_PATH, files=files)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.artifact['_href'])
+
+    def test_01_create_content_units(self):
```
Ichimonji10: If this test creates one content unit, shouldn't it be called `test_01_create_content_unit`, without the trailing "s"?

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+        with self.assertRaises(HTTPError):
+            self.client.get(self.content_unit['_href'])
+
+
+def _gen_verbose_content_unit_attrs(artifact, relative_path):
+    """Generate a dict with content unit attributes.
+
+    :param: artifact: A dict about the artifact.
+    :param: relative_path: The relative path in which one the content unit
+     will be available.
```
Ichimonji10: Indent this by four lines.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+    """
```
Ichimonji10: Add a `:returns:` line, here.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+    return {'artifact': artifact['_href'], 'relative_path': relative_path}
```
Ichimonji10: Why let `relative_path` be passed in, when `utils.uuid4()` is a sane default? Let the method signature be:

```python
_gen_verbose_content_unit_attrs(artifact)
```

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @selectors.skip_if(bool, 'content_unit', False)
+    def test_04_delete_content_unit(self):
+        """Delete a content unit."""
+        self.client.delete(self.content_unit['_href'])
+        with self.assertRaises(HTTPError):
+            self.client.get(self.content_unit['_href'])
+
+
+def _gen_verbose_content_unit_attrs(artifact, relative_path):
```
Ichimonji10: What's the minimum set of attributes that can be used to create a content unit? An artifact href and a relative path? If so, this function doesn't generate a "verbose" dict of attributes.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
kersommoura: @Ichimonji10, the test does break.

kersommoura:
@Ichimonji10, thanks for your review. Your comments were addressed.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
kersommoura: The variable definition on line 26 was removed.

pulp_smash/tests/pulp3/file/api_v3/test_crud_content_unit.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean class-wide variable."""
+        cls.client.delete(cls.artifact['_href'])
+
+    def test_01_create_content_unit(self):
+        """Create content unit."""
+        attrs = _gen_content_unit_attrs(self.artifact)
+        type(self).content_unit = self.client.post(FILE_CONTENT_PATH, attrs)
```
Ichimonji10: This test doesn't verify that the `attrs` that are sent were accepted by Pulp. Try something like this:

```python
    def test_01_create_content_unit(self):
        """Create content unit."""
        attrs = _gen_content_unit_attrs(self.artifact)
        self.content_unit.update(self.client.post(FILE_CONTENT_PATH, attrs))
        for key, val in attrs.items():
            with self.subTest(key=key):
                self.assertEqual(self.content_unit[key], val)
```

Note that I call `self.content_unit.update()` instead of overwriting `type(self).content_unit`. Calling `update()` is safer, as it prevents certain classes of mistakes. Consider the following code:

```python
type(self).contennt_unit = self.client.post(FILE_CONTENT_PATH, attrs)
self.contennt_unit.update(self.client.post(FILE_CONTENT_PATH, attrs))
```

The former line of code will create a brand new attribute called `contennt_unit`, whereas the latter will throw an invalid attribute error.

kersommoura: done.

PR #894, opened 2018-03-06
Update content fields

daviddavis:
Updating content fields based on changes in:

https://github.com/pulp/pulp_file/pull/51


Ichimonji10:
See: https://github.com/pulp/pulp_file/pull/51

PR #896, opened 2018-03-07
Re-factor read_repo_content.

kersommoura:
Pulp3 allows mutliple repository versions. Adjust function
`read_repo_content` to reflect this.
Add the paramter `version` to the function `read_repo_content` thus
allowing read the content of a specific repository version. If `version`
is None, the latest repo version will be read.
This change will help future tests that require read content from the
same repo, but from different versions.


kersommoura:

pulp-file==0.0.1a18
pulpcore==3.0.0a25
pulpcore-common==3.0.0a14
pulpcore-plugin==0.0.1a20

PR #903, opened 2018-03-12
Attributes http and https no longer exist on distributions

daviddavis:


Ichimonji10:
There's a few other places in the code where references to distribution http and https show up. I'll poke around and see what I can find.

daviddavis:
Tested this change out locally against [my PR](https://github.com/pulp/pulp/pull/3368):

```
53 passed, 5 skipped in 144.24 seconds
```

Issue #897, opened 2018-03-07
As a user, I can publish a particular repository version.


daviddavis:
Currently, users publish by supplying `repository` to the publication endpoint. This new feature allows them to specify `repository_version` instead of the latest version.

1. Create a repo with at least 2 repository versions (e.g. versions 1 and 2).
2. Create a publication by supplying the non-latest `repository_version` (ie version 1).
3. With the new publication confirm that the repository_version attribute points to version 1 (and not version 2).

daviddavis:
https://pulp.plan.io/issues/3324

PR #909, opened 2018-03-27
Add test to verify publish of any repo version.

kersommoura:
Add function get_repo_versions to `pulp3/utils`.
Rename variable in `publish_repo`.

Do the following:

 1. Create a repository with at least 2 repository versions.
 2. Create a publication by supplying the latest `repository_version`.
 3. Assert that the publication `repository_version` attribute points to the latest repository version.
 4. Create a publication by supplying the non-latest `repository_version`.
 5. Assert that the publication `repository_version` attribute points to the supplied repository version.
 6. Assert that an exception is raised when providing two different repository versions to be published at same time.

Closes: #897

### DIFF
--- a/pulp_smash/tests/pulp3/utils.py
+++ b/pulp_smash/tests/pulp3/utils.py
@@ -3,7 +3,7 @@
 import random
 import unittest
 from copy import deepcopy
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlsplit
 
 from packaging.version import Version
 from requests.auth import AuthBase, HTTPBasicAuth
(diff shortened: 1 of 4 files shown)


pulp_smash/tests/pulp3/utils.py
```diff
@@ -289,3 +287,17 @@ def clean_content_units(cfg=None):
...
+
+
+def get_repo_versions(repo):
+    """Return hrefs of repository versions.
+
+    :param repo: A dict of information about the repository.
+    :returns: A sorted list with the hrefs of repository versions.
+    """
+    client = api.Client(config.get_config(), api.json_handler)
+    versions = client.get(repo['_versions_href'])['results']
```
Ichimonji10: This works well so long as only one page of results are returned. If there are more repository versions than can fit in a single page, this method will return incorrect results. This is OK in the sort term, but a generator that walks through all pages and yields hrefs should be written.

kersommoura: @Ichimonji10, that is a great improvement. I will address your comment.

PR #911, opened 2018-03-29
"Fix" test case affected by Pulp issue 3540

Ichimonji10:
Issue title: "When pulp_workers are restarted, /etc/default/pulp_workers
might be ignored."

See: https://pulp.plan.io/issues/3540


pcreech:
Yes, please!

Issue #871, opened 2018-02-19
As an authenticated user, I can delete any repository version.


kersommoura:
Add tests to explore functionality to delete any repository version - Pulp3.

See issue: [3219](https://pulp.plan.io/issues/3219)

This issue will be update with more details

It was suggested by @dkliban to use a large number of files, at least 10 files, and create a repo version for each file that was added to the repo.

daviddavis:
A couple things to check for:

1. Deleting a version (e.g. 1) doesn't affect the content of successive versions (e.g. 2, 3, ...)
2. The last version can be deleted
3. If I delete the last version (e.g. 3) and then create a new version, it should not have the previously deleted version # (e.g. 3). It should be 4.
4. If I delete a version and it has a publication, the publication should be removed too.

Issue #918, opened 2018-04-03
Let get_versions() handle pagination


Ichimonji10:
[`pulp_smash.pulp3.utils.get_versions()`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.pulp3.utils.html#pulp_smash.pulp3.utils.get_versions) works well so long as only one page of results is returned. If there are multiple pages of results, when the method will return incorrect results. This is OK in the short term, with a small test suite and simple tests, but it will silently produce incorrect results long-term. The method should be re-written to handle the case where multiple pages of results are returned. I suggest writing a generator that walks through all pages of results and yields them one by one, and letting `get_repo_versions()` use that generator.

Ichimonji10:
`get_artifact_paths()` should also handle pagination.

kersommoura:
Perhaps, add a new callback the api.Client, to handle pagination.

Ichimonji10:
This is done. The relevant work is spread out over about ten commits, in a branch in my repo. I'm submitting the PRs one at a time, to make reviews easier.

kersommoura:
@Ichimonji10, thanks.

Ichimonji10:
This issue has ben solved by a string of other commits. The most important change is that there is now a [`page_handler`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.api.html#pulp_smash.api.page_handler) callback function for paging through responses.

PR #919, opened 2018-04-03
Add tests to delete any repo version.

kersommoura:
Add 2 test cases.

First:

 1. Delete a version (e.g. 1) and verify doesn't affect the content of successive versions (e.g. 2, 3, ...).
 2. Assert that the last version can be deleted.
 3. Verify whether when the last version is deleted and then a new version is created, it should not have the previously deleted version content.

Second:

1. Test if delete a given repo version will delete its publication.

Closes:#871

### DIFF
--- a/pulp_smash/tests/pulp3/utils.py
+++ b/pulp_smash/tests/pulp3/utils.py
@@ -286,3 +286,34 @@ def get_repo_versions(repo):
         [version['_href'] for version in versions],
         key=lambda url: int(urlsplit(url).path.split('/')[-2])
     )
+
+
+def get_artifact_paths(repo, version_href=None):
+    """Return the paths of artifacts present in a given repository version.
+
+    :param repo: A dict of information about the repository.
+    :param version_href: The repository version to read.
+    :returns: A set with the paths of units present in a given repository.
+    """
+    return {
+        content_unit['artifact']  # file path and name
+        for content_unit in get_content(repo, version_href)['results']
+    }
+
+
+def delete_repo_version(repo, version_href=None):
+    """Delete a given repository version.
+
+    :param repo: A dict of information about the repository.
+    :param version_href: The repository version that should be
+        deleted.
+    :returns: A tuple. The tasks spawned by Pulp.
+    """
+    if version_href is None:
+        version_href = repo['_latest_version_href']
+    cfg = config.get_config()
+    client = api.Client(cfg, api.json_handler)
+    call_report = client.delete(version_href)
+    # As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
+    # returning multiple tasks, this may need to be re-written.
+    return tuple(api.poll_spawned_tasks(cfg, call_report))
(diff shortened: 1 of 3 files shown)


pulp_smash/tests/pulp3/utils.py
```diff
@@ -290,3 +290,34 @@ def get_repo_versions(repo):
...
     )
+
+
+def get_artifact_paths(repo, version_href=None):
+    """Return the paths of artifacts present in a given repository version.
+
+    :param repo: A dict of information about the repository.
+    :returns: A list with the paths of units present in a given repository.
+    """
+    return set([
```
Ichimonji10: Rather than creating a list and a set, create just a set:

```python
return {content_unit['artifact'] for content_unit in ...}
```

Set comprehension and dict comprehension syntax is similar. Here's samples of dict comprehension syntax:

```python
{key: val for key, val in ((1, 2), (3, 4), (5, 6))}
{item[0]: item[1] for item in ((1, 2), (3, 4), (5, 6))}
```

pulp_smash/tests/pulp3/utils.py
```diff
@@ -290,3 +290,34 @@ def get_repo_versions(repo):
...
+        a given repository version.
+    """
+    if version_href is None:
+        version_href = repo['_latest_version_href']
+    cfg = config.get_config()
+    client = api.Client(cfg, api.json_handler)
+    call_report = client.delete(version_href)
+    # As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
+    # returning multiple tasks, this may need to be re-written.
+    return next(api.poll_spawned_tasks(cfg, call_report))
```
Ichimonji10: See https://github.com/PulpQE/pulp-smash/issues/925.

Ichimonji10: Alternately, let this function `return tuple(api.poll_spawned_tasks(cfg, call_report))`.

kersommoura:
@Ichimonji10, I will address your comments today.

kersommoura:
@Ichimonji10, your comments were addressed.

pulp_smash/tests/pulp3/utils.py
```diff
@@ -286,3 +286,35 @@ def get_repo_versions(repo):
...
+def get_artifact_paths(repo, version_href=None):
+    """Return the paths of artifacts present in a given repository version.
+
+    :param repo: A dict of information about the repository.
+    :param version_href: The repository version to read.
+    :returns: A set with the paths of units present in a given repository.
+    """
+    return {
+        content_unit['artifact']  # file path and name
+        for content_unit in get_content(repo, version_href)['results']
```
Ichimonji10: See https://github.com/PulpQE/pulp-smash/issues/918#issuecomment-380129401.

pulp_smash/tests/pulp3/utils.py
```diff
@@ -286,3 +286,35 @@ def get_repo_versions(repo):
...
+
+
+def delete_repo_version(repo, version_href=None):
+    """Delete a given repository version.
+
+    :param repo: A dict of information about the repository.
+    :param version_href: The repository version that should be
+        deleted.
+    :returns: A tuple of information about the task created when deleting
+        a given repository version.
```
Ichimonji10: This method doesn't necessarily return a single task. It returns a tuple of tasks. Maybe this could read:

> A tuple. The tasks spawned by Pulp.

Ichimonji10:
Merged, with a follow-up commit of 22c3400fa4ab145f35aa1b7709f3fd8419050064.

Issue #908, opened 2018-03-21
As a user, when I delete an artifact, it is removed from the filesystem.


daviddavis:
1. Create an artifact and then delete it
2. Check that it's removed from the filesystem

https://pulp.plan.io/issues/3508

PR #920, opened 2018-04-03
Add test to verify artifact is removed from filesystem.

kersommoura:
Do the following:

1. Create and delete an artifact.
2. Assert that artifact file is not present on the filesystem.

Closes:#908


kersommoura:
@Ichimonji10, diff was applied. Thanks for your review.

daviddavis:
I think this should be a pulpcore test and not a pulp_file test. Artifacts are part of pulpcore:

https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/viewsets/content.py#L67

PR #924, opened 2018-04-04
Match changes in status API

dralley:


pulp_smash/tests/pulp3/pulpcore/api_v3/test_status.py
```diff
@@ -35,8 +36,12 @@ def test_02_data_content(self):
         """Verify whether few parameters are present on status page."""
         self.assertTrue(self.status['database_connection'])
         self.assertTrue(self.status['messaging_connection'])
-        self.assertTrue(self.status['known_workers'])
-        self.assertTrue(self.status['versions'])
+        self.assertIsInstance(list, self.status['online_workers'])
```
kersommoura: @dralley , please, update the order of parameters as per documentation. The same is valid for lines 39-41.

See : [isistance](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance)

dralley: fixed

Ichimonji10: By way of explanation, here's an example of the `isinstance` builtin in action:

```
>>> isinstance([], list)
True
>>> isinstance(list, [])
TypeError: isinstance() arg 2 must be a type or tuple of types
```

PR #927, opened 2018-04-04
Expand read test case for content units.

kersommoura:
Add an extra test case to test the GET method and the endpoint
`/api/v3/content/file/`, together with the query parameter
`relative_path`.


kersommoura:
Related issue: https://pulp.plan.io/issues/3084

Issue #926, opened 2018-04-04
Remove Content Unit deletion tests, and replace the delete Content Unit operations with orphan cleanup


dralley:
Pulp Issue #3445 (https://pulp.plan.io/issues/3445) removed the ability to delete content units, because it's inherently racy.  Content Units could be deleted while a using them is still running.

Therefore plugin tests (currently the file plugin) which test Content Unit CRUD are currently broken.  The Content Unit tests that test deletion should be removed, and the "tear down" routine should use orphan cleanup - once that functionality is completed (PR open to do so currently).

dralley:
Orphan cleanup PR is merged - this issue is now unblocked

dralley:
Related to: https://github.com/PulpQE/pulp-smash/issues/914

Issue #914, opened 2018-04-02
As a user, I can clean up both orphaned content units and orphaned artifacts


daviddavis:
## Test Case 1

1. Do a sync
3. Create a few orphan content units
2. Run a cleanup by calling `DELETE /api/v3/orphans/`

**Result:** Orphan content units should be cleaned up but the other content units created by sync should not.

## Test Case 2

1. Create an orphan artifact
2. Run a cleanup by calling `DELETE /api/v3/orphans/`

**Result:** The artifact and the file on the filesystem should be cleaned up.

## Test Case 3

1. Create an orphan content unit with an artifact
2. Run a cleanup by calling `DELETE /api/v3/orphans/`

**Result:** The content unit and the artifact should be cleaned up. Also, the artifact's file on the filesystem should be cleaned up too.

https://pulp.plan.io/issues/3442

asmacdo:
An orphaned content unit is a content unit that is not in any repository version. 
An orphaned artifact is an artifact that is not in any content units.

PR #928, opened 2018-04-06
Remove delete content units.

kersommoura:
In order to avoid a race condition it was removed the feature to remove
content units.

See: https://pulp.plan.io/issues/3445

A new endpoint was added `/api/v3/orphans/` and using the
DELETE method and this endpoint all orphan content units will be
removed. This endpoint also removes orphan artifact units.

Adjust pulp-smash to this new scenario.

Remove:
 * clean_content_units
 * clean_artifacts

Add:
 * delete_orphan_units

Closes:#926

Related:#914


Ichimonji10:
:100:

Issue #925, opened 2018-04-04
Be notified when Pulp 3 returns multiple tasks after certain calls


Ichimonji10:
The code base currently includes this kludge:

```python
# As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
# returning multiple tasks, this may need to be re-written.
last_task = next(api.poll_spawned_tasks(cfg, call_report))
return client.get(last_task['created_resources'][0])
```

What happens if `call_report` leads to multiple tasks? Every task after the first is ignored. That's bad. Even worse, I noticed that this same idiom is repeated in a recent PR.

Let's update each of these cases with code like this:

```python
# As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
# returning multiple tasks, this may need to be re-written.
tasks = tuple(api.poll_spawned_tasks(cfg, call_report))
if len(tasks) != 1:
    warnings.warn(
        'Multiple tasks were spawned in response to an API call. This is '
        'unexpected, and Pulp Smash may handle the response incorrectly',
        RuntimeWarning
    )
return client.get(tasks[-1]['created_resources'][0])
```

Or code like this:

```python
# As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
# returning multiple tasks, this may need to be re-written.
tasks = tuple(api.poll_spawned_tasks(cfg, call_report))
assert len(tasks) == 1
return client.get(tasks[-1]['created_resources'][0])
```

...or so on.

kersommoura:
See also: https://github.com/PulpQE/pulp-smash/issues/795

PR #932, opened 2018-04-09
Notify when Pulp 3 returns multiples tasks.

kersommoura:
Modify function `publish_repo` to issue a warning in case of number of
spawned tasks are different than one.

Adjust the `test_crd_publications` to use the `publish_repo` helper
function.

Closes:#925
Related:#795

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_crd_publications.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_crd_publications.py
@@ -19,7 +19,11 @@ from pulp_smash.tests.pulp3.file.api_v3.utils import (
 )
 from pulp_smash.tests.pulp3.file.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
 from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
-from pulp_smash.tests.pulp3.utils import get_auth, sync_repo
+from pulp_smash.tests.pulp3.utils import (
+    get_auth,
+    publish_repo,
+    sync_repo,
+)
 
 
 class PublicationsTestCase(unittest.TestCase, utils.SmokeTest):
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Perhaps both of these should be updated, not just one?

```console
$ git checkout upstream/master
$ git grep 'next(' pulp_smash/tests/pulp3/
pulp_smash/tests/pulp3/file/api_v3/test_crd_publications.py:        last_task = next(api.poll_spawned_tasks(self.cfg, call_report))
pulp_smash/tests/pulp3/utils.py:    last_task = next(api.poll_spawned_tasks(cfg, call_report))
```

pulp_smash/tests/pulp3/file/api_v3/test_crd_publications.py
```diff
@@ -61,6 +65,7 @@ def test_01_create_publication(self):
             urljoin(self.publisher['_href'], 'publish/'),
             {'repository': self.repo['_href']}
         )
+        verify_number_tasks(len(call_report))
```
Ichimonji10: If `len` must be used every time `verify_number_tasks` is called, then just call `len` inside `verify_number_tasks`. DRY.

pulp_smash/tests/pulp3/utils.py
```diff
@@ -186,8 +187,25 @@ def publish_repo(cfg, publisher, repo, version_href=None):
...
     # As of this writing, Pulp 3 only returns one task. If Pulp 3 starts
     # returning multiple tasks, this may need to be re-written.
-    last_task = next(api.poll_spawned_tasks(cfg, call_report))
-    return client.get(last_task['created_resources'][0])
+    verify_number_tasks(len(call_report))
+    tasks = next(api.poll_spawned_tasks(cfg, call_report))
+    return client.get(tasks['created_resources'][0])
+
+
+def verify_number_tasks(number_tasks):
```
Ichimonji10: It'd be very useful to print the contents of a malformed call report within this function.

Issue #893, opened 2018-03-06
As an authenticated user, I can access auto generated docs.


kersommoura:
As an authenticated user, I can access `api/v3/docs` and have access to the auto generated docs.

This require the installation of drf_openapi. See: [drf_open_api](https://docs.pulpproject.org/en/3.0/nightly/integration_guide/rest_api/index.html#pulpcore-rest-api)

daviddavis:
👍 

It might be nice to have two tests here. One against pulpcore and one against pulp_file (making sure that pulp_file routes show up when pulp_file is installed.)

kersommoura:
See  also: https://pulp.plan.io/issues/3539

PR #937, opened 2018-04-11
Add tests to auto generated docs.

kersommoura:
Add basic test coverage to the API auto generated docs.

Verify whether the API auto generated docs are available.

Closes:#893

### DIFF
--- a/pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
+++ b/pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
@@ -1,0 +1,49 @@
+# coding=utf-8
+"""Test related to the api docs page."""
+import unittest
+
+from requests.exceptions import HTTPError
+
+from pulp_smash import api, config, selectors, utils
+from pulp_smash.tests.pulp3.constants import API_DOCS_PATH
+from pulp_smash.tests.pulp3.pulpcore.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
+
+
+class ApiDocsTestCase(unittest.TestCase, utils.SmokeTest):
+    """Test whether API auto generated docs are available.
+
+    This test targets the following issues:
+
+    * `Pulp #3552 <https://pulp.plan.io/issues/3552>`_
+    * `Pulp Smash #893 <https://github.com/PulpQE/pulp-smash/issues/893>`_
+    """
+
+    def setUp(self):
+        """Create an API Client."""
+        cfg = config.get_config()
+        if selectors.bug_is_untestable(3552, cfg.pulp_version):
+            self.skipTest('https://pulp.plan.io/issues/3552')
+        self.client = api.Client(cfg)
+
+    def test_valid_credentials(self):
+        """Get API documentation with valid credentials.
+
+        Assert the API documentation is returned.
+        """
+        self.client.get(API_DOCS_PATH)
+
+    def test_no_credentials(self):
+        """Get API documentation with no credentials.
+
+        Assert the API documentation is returned.
+        """
+        del self.client.request_kwargs['auth']
+        self.client.get(API_DOCS_PATH)
+
+    def test_http_method(self):
+        """Get API documentation with an HTTP method other than GET.
+
+        Assert an error is returned.
+        """
+        with self.assertRaises(HTTPError):
+            self.client.post(API_DOCS_PATH)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -144,6 +144,7 @@ developers, not a gospel.
     api/pulp_smash.tests.pulp3.file.utils
     api/pulp_smash.tests.pulp3.pulpcore
     api/pulp_smash.tests.pulp3.pulpcore.api_v3
+    api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_api_docs
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_auth
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crd_artifacts
     api/pulp_smash.tests.pulp3.pulpcore.api_v3.test_crud_distributions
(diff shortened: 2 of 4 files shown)


pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
```diff
@@ -0,0 +1,39 @@
...
+    * `Pulp Smash #893 <https://github.com/PulpQE/pulp-smash/issues/893>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.cfg = config.get_config()
+        if selectors.bug_is_untestable(3552, cls.cfg.pulp_version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/3552')
+        cls.client = api.Client(cls.cfg, api.safe_handler)
```
Ichimonji10: `safe_handler` is the default response handler.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_01_access_api_docs(self):
+        """Test whether the API auto generated docs page is available."""
+        self.client.get(API_DOCS_PATH)
+
+    def test_02_http_method(self):
```
Ichimonji10: These methods don't have any ordering, do they? Drop the numbers from the test method names.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
```diff
@@ -0,0 +1,39 @@
...
+    This test targets the following issues:
+
+    * `Pulp #3552 <https://pulp.plan.io/issues/3552>`_
+    * `Pulp Smash #893 <https://github.com/PulpQE/pulp-smash/issues/893>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.cfg = config.get_config()
```
Ichimonji10: This variable is only used in this method. Make it a local variable, not a class variable. It can be made a class variable in the future if it's needed more broadly.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
```diff
@@ -0,0 +1,51 @@
...
+        """Test whether the API docs are available with bad credentials."""
+        self.client.request_kwargs['auth'] = HTTPBasicAuth(
+            utils.uuid4(),
+            utils.uuid4()
+        )
+        with self.assertRaises(HTTPError):
+            self.client.get(API_DOCS_PATH)
+
+    def test_access_api_docs_no_auth(self):
+        """Test whether the API docs page is available without auth."""
```
Ichimonji10: This test doesn't do what's stated. `self.client` has authentication credentials.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
```diff
@@ -0,0 +1,51 @@
...
+        cls.cfg = config.get_config()
+        if selectors.bug_is_untestable(3552, cls.cfg.pulp_version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/3552')
+
+    def setUp(self):
+        """Create an API Client."""
+        self.client = api.Client(self.cfg)
+
+    def test_access_api_docs_bad_auth(self):
+        """Test whether the API docs are available with bad credentials."""
```
Ichimonji10: Is this behaviour guaranteed somewhere? In other words, is it guaranteed that a request to a public API with bad credentials will produce an error? Or is that an implementation detail, where Pulp's behaviour might change in the future?

kersommoura: This behaviour may change in the future. 

See: https://pulp.plan.io/issues/3552#note-6

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
kersommoura: A new client is created for every test call. 

self.client.request_kwargs['auth']  is not used  to create authentication for each of these clients. Except in the case that bad credentials are used.

Ichimonji10: > A new client is created for every test call.

Correct.

> `self.client.request_kwargs['auth']` is not used to create authentication for each of these clients. Except in the case that bad credentials are used.

Really?

```python
>>> from pulp_smash import api, config
>>> client = api.Client(config.get_config())
>>> client.request_kwargs['auth']
('admin', 'admin')
```

pulp_smash/tests/pulp3/pulpcore/api_v3/test_api_docs.py
Ichimonji10: That note doesn't really address my question.

Ichimonji10:
Sample code:

```python
class ApiDocsTestCase(unittest.TestCase, utils.SmokeTest):
    """Test whether API auto generated docs are available.

    This test targets the following issues:

    * `Pulp #3552 <https://pulp.plan.io/issues/3552>`_
    * `Pulp Smash #893 <https://github.com/PulpQE/pulp-smash/issues/893>`_
    """

    def setUp(self):
        """Create an API Client."""
        cfg = config.get_config()
        if selectors.bug_is_untestable(3552, cfg.pulp_version):
            self.skipTest('https://pulp.plan.io/issues/3552')
        self.client = api.Client(cfg)

    def test_valid_credentials(self):
        """Get API documentation with valid credentials.

        Assert the API documentation is returned.
        """
        self.client.get(API_DOCS_PATH)

    def test_no_credentials(self):
        """Get API documentation with no credentials.

        Assert the API documentation is returned.
        """
        del self.client.request_kwargs['auth']
        self.client.get(API_DOCS_PATH)

    def test_http_method(self):
        """Get API documentation with an HTTP method other than GET.

        Assert an error is returned.
        """
            self.client.post(API_DOCS_PATH)
```

PR #940, opened 2018-04-12
Add pycodestyle to makefile.

kersommoura:
pycodestyle is listed as one of the dev requirements. But it is not being used.

Add pycodestyle to makefile, and as part of `make all`.

Add flag to `-ignore=E501` to pycodestyle, let pylint handles `line
too long error`.

Fix two minor issues found using pycodestyle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ help:
 	@echo "  docs-clean     to remove documentation"
 	@echo "  lint           to run all linters"
 	@echo "  lint-flake8    to run the flake8 linter"
+	@echo "  lint-pycode    to run the pycodestyle linter"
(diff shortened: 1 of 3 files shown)


Ichimonji10:
To recap what we talked about on Bluejeans:

* flake8 calls pycodestyle.
* flake8 and pylint both check for certain things, like line length. It'd be good if just one linter checked for those things. This can be accomplished by adding `F401,E501` to the make file.
* The `noqa` pragmas ignore too much stuff.
* This will drop all plain `noqa` pragmas from the code base: `git grep --files-with-matches noqa . | xargs sed -i 's/noqa //'`.

Issue #879, opened 2018-02-23
Multi-resource locking


daviddavis:
Not sure exactly how to test this out but basically, here's the use case:

1. Create a repo and a remote
2. Sync the repo and remote
3. Try to update the repo or the remote during the sync
4. The update shouldn't happen until after the sync

Another use case:

1. Create a repo and an remote
2. Update the remote to point to a different url
3. Immediately run a sync
4. The sync should fire after the update and sync from the second url

https://pulp.plan.io/issues/3186

PR #941, opened 2018-04-12
Add test to verify multi-resource locking.

kersommoura:
Do the following:

 1. Create a repository, and a remote.
 2. Update the remote to point to a different url.
 3. Immediately run a sync. The sync should fire after the update and
     sync from the second url.
 4. Assert that remote url was updated.
 5. Assert that the number of units present in the repository is
    according to the updated url.

Closes:#879

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
@@ -6,7 +6,11 @@ from random import randint
 from urllib.parse import urljoin, urlsplit
 
 from pulp_smash import api, config, utils
-from pulp_smash.constants import FILE_FEED_URL
+from pulp_smash.constants import (
+    FILE_FEED_COUNT,
+    FILE_FEED_URL,
+    FILE_LARGE_FEED_URL,
+)
 from pulp_smash.tests.pulp3.constants import (
     FILE_REMOTE_PATH,
     REMOTE_SYNC_MODE,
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Very good. The only substantive suggestion I can think of is to use subtests to make the assertions about the repo and the remote independent. There *shouldn't* be a case where one assertion passes and the other fails. But it's trivial to make the assertions independent, and if one of them starts failing, it will be very useful to know if the other assertion passed.

Separately, a little bit of whitespace would do wonders for readability.

Sample code:

```diff
diff --git a/pulp_smash/tests/pulp3/file/api_v3/test_sync.py b/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
index 5462e66..4456472 100644
--- a/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_sync.py
@@ -157,16 +157,27 @@ class MultiResourceLockingTestCase(unittest.TestCase, utils.SmokeTest):
         cfg = config.get_config()
         client = api.Client(cfg, api.json_handler)
         client.request_kwargs['auth'] = get_auth()
+
+        # create repo
         repo = client.post(REPO_PATH, gen_repo())
         self.addCleanup(client.delete, repo['_href'])
+
+        # create and update remote
         body = gen_remote()
         body['url'] = urljoin(FILE_LARGE_FEED_URL, 'PULP_MANIFEST')
         remote = client.post(FILE_REMOTE_PATH, body)
         self.addCleanup(client.delete, remote['_href'])
         url = {'url': urljoin(FILE_FEED_URL, 'PULP_MANIFEST')}
         client.patch(remote['_href'], url)
+        remote = client.get(remote['_href'])
+        with self.subTest(comment='verify remote url'):
+            self.assertEqual(remote['url'], url['url'])
+
+        # sync repo
         sync_repo(cfg, remote, repo)
         repo = client.get(repo['_href'])
-        remote = client.get(remote['_href'])
-        self.assertEqual(remote['url'], url['url'])
-        self.assertEqual(len(get_content(repo)['results']), FILE_FEED_COUNT)
+        with self.subTest(comment='verify repo content'):
+            self.assertEqual(
+                len(get_content(repo)['results']),
+                FILE_FEED_COUNT
+            )
```

Ichimonji10:
<ins>Never mind.</ins>

PR #942, opened 2018-04-12
Update linters and remove noqa from codebase.

kersommoura:
Update errors ignored by flake8. Add E501, and F401.

flake8 calls pycodestyle, but errors found by pycodestyle were ignored
due to the use of `noqa`.

Remove the majority of `noqa`  from the codebase. Remains nine `noqa` in
the whole codebase, all realted to same issue  `# noqa:E722`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,9 @@ docs-html:
 docs-clean:
 	@cd docs; $(MAKE) clean
 
+# E501 and F401 are ignored because Pylint performs similar checks.
 lint-flake8:
-	flake8 . --ignore D203 --exclude docs/_build
+	flake8 . --ignore D203,E501,F401 --exclude docs/_build
 
 lint-pylint:
 	pylint -j $(CPU_COUNT) --reports=n --disable=I \
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -365,7 +365,8 @@ class PulpSmashConfig(object):
         )
         self._xdg_config_dir = 'pulp_smash'
 
-    def __repr__(self):  # noqa
+    def __repr__(self):
+        """Create string representation of the object."""
         attrs = _public_attrs(self)
         attrs['pulp_version'] = type('')(attrs['pulp_version'])
         str_kwargs = ', '.join(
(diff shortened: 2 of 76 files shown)


setup.py
```diff
@@ -53,7 +53,6 @@
             'flake8',
             'flake8-docstrings',
             'flake8-quotes',
-            'pydocstyle',
```
Ichimonji10: Why is an unrelated dependency being dropped? This is py**doc**style, not py**code**style.

kersommoura: Fixing right now.

Makefile
```diff
@@ -36,7 +36,7 @@ docs-clean:
 	@cd docs; $(MAKE) clean
 
 lint-flake8:
-	flake8 . --ignore D203 --exclude docs/_build
+	flake8 . --ignore D203,E501,F401 --exclude docs/_build
```
Ichimonji10: Since you're messing around with the flake8 linter, please drop the reference to D203. See 76dfc72a0ef0dbfc2b6c55507614b2c3bb3e3c29

Makefile
```diff
@@ -36,7 +36,7 @@ docs-clean:
 	@cd docs; $(MAKE) clean
 
 lint-flake8:
```
Ichimonji10: Can a comment be added above this line explaining why E501 and F401 are being ignored? Perhaps something like this:

> E501 and F401 are ignored because Pylint performs similar checks.

Ichimonji10:
`git grep noqa` shows that there are still several bare `noqa` pragmas in the code base.

kersommoura:
@Ichimonji10, I removed almost all `noqa` from the codebase. I just kept the ones related to ` # noqa:E722`. There are nine of them in the whole codebase. All the others were removed.

Makefile
```diff
@@ -36,7 +36,8 @@ docs-clean:
...
-	flake8 . --ignore D203 --exclude docs/_build
+    # E501 and F401 are ignored because Pylint performs similar checks.
```
Ichimonji10: Move this comment up one line.

Also, note that make file syntax requires that make recipes (like this line of code) be preceded by one or more tabs, *not* spaces.

PR #958, opened 2018-04-20
Update assertions and variable name for workers.

kersommoura:
Update variable in the ``test_03_positive_filters``. Use
``last_heartbeat__gte`` instead of ``last_update__gte``.

Besides that, remove assertions related to ``last_heartbeat`` once this
parameter can be udpated between the query and assertion moment.

### DIFF
--- a/pulp_smash/tests/pulp3/pulpcore/api_v3/test_workers.py
+++ b/pulp_smash/tests/pulp3/pulpcore/api_v3/test_workers.py
@@ -45,8 +45,9 @@ class WorkersTestCase(unittest.TestCase, utils.SmokeTest):
         """Read a worker by its _href."""
         worker = self.client.get(self.worker['_href'])
         for key, val in self.worker.items():
-            with self.subTest(key=key):
-                self.assertEqual(worker[key], val)
+            if key != 'last_heartbeat':
+                with self.subTest(key=key):
+                    self.assertEqual(worker[key], val)
 
     @selectors.skip_if(bool, 'worker', False)
     def test_02_read_workers(self):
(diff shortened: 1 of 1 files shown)


pulp_smash/tests/pulp3/pulpcore/api_v3/test_workers.py
```diff
@@ -56,22 +57,24 @@ def test_02_read_workers(self):
...
+            'last_heartbeat__gte': self.worker['last_heartbeat'],
             'name': self.worker['name'],
             'online': self.worker['online'],
             'missing': self.worker['missing'],
         })
         self.assertEqual(len(page['results']), 1)
         for key, val in self.worker.items():
-            with self.subTest(key=key):
-                self.assertEqual(page['results'][0][key], val)
+            if key != 'last_heartbeat':
```
dkliban: what about last_updated? i think that value also changes when the heartbeat chanages.

kersommoura: @dkliban looking into right now.

kersommoura: When reading a worker I got the following response.

``` text
{'_href': 'http://f27.p3.source:8000/api/v3/workers/35059a9f-20ac-41a8-92e6-41f1fb5af7a4/',
 'last_heartbeat': '2018-04-18T15:21:26.181917Z',
 'missing': True,
 'name': 'reserved_resource_worker_1@localhost',
 'online': False}
```
I could not identify the mentioned variable.

PR #959, opened 2018-04-20
[DON'T MERGE] Reduces number of filters in worker test case

dkliban:


kersommoura:
See: https://github.com/PulpQE/pulp-smash/pull/962

Ichimonji10:
What's the status of this PR? Are there improvements that can be made in the immediate future so that this can be merged, or should this be closed until the bug at hand is figured out?

PR #960, opened 2018-04-20
Switch status API to RQ

bmbouter:
The attribute name in the status API changed when RQ with Redis replaced
AMQP brokers like RabbitMQ.

This PR is meant to be used with these core and pulp_file PRs:

https://github.com/pulp/pulp/pull/3454
https://github.com/pulp/pulp_file/pull/72

### DIFF
--- a/pulp_smash/tests/pulp3/pulpcore/api_v3/test_status.py
+++ b/pulp_smash/tests/pulp3/pulpcore/api_v3/test_status.py
@@ -21,7 +21,7 @@ STATUS = {
             'type': 'object',
             'properties': {'connected': {'type': 'boolean'}},
         },
-        'messaging_connection': {
+        'redis_connection': {
             'type': 'object',
             'properties': {'connected': {'type': 'boolean'}},
         },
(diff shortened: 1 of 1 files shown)


dralley:
This PR should be renamed and the scope expanded a bit, I think.  The settings file generated by ```pulp-smash settings create``` should be modified, along with all the code that reads it.

kersommoura:
@dralley , please, can you file an smash issue providing more details, and outlining the requested changes?

dralley:
@kersommoura https://github.com/PulpQE/pulp-smash/issues/965

bmbouter:
No problem at all I can rebase. I'll do that soon. Thank you @Ichimonji10. I'm also considering relabeling it back to 'tasking_connection' or something like that which was a feedback point from another PR. The thinking is that RQ can support other backends (at some point for instance etcd is on their roadmap). I'll make that change with my rebase too.

bmbouter:
@Ichimonji10 I pushed a rebased change. What else needs to be done in this PR to make it ready for merging?

PR #961, opened 2018-04-20
Add more info to worker filter error

daviddavis:


daviddavis:
@kersommoura I think it would be beneficial to have this change regardless of whether we skip the test or not. The current error `2 != 1` is not very helpful.

PR #967, opened 2018-04-23
Remove sync_mode and download_policy from Remote

asmacdo:
https://github.com/pulp/pulp/pull/3450

These settings were removed from the Remote in pulpcore:
https://pulp.plan.io/issues/3492

Going forward, these settings will be passed as parameters to the
plugins. Currently however, the file plugin does not make use of either
of them, so we have removed the fields from sync calls altogether.


bmbouter:
@asmacdo I agree with these changes. This looks right based on the recent core changes.

PR #968, opened 2018-04-23
Consolidate and flatten test logic

Ichimonji10:
Use a single variable for conceptually identical test code, and flatten
test logic.

See: 8c948e996fd8a6e61cd84e18a7f1cb945603a7e1


Ichimonji10:
Tested against from-source installs of Pulp on F26 and F27.

PR #969, opened 2018-04-23
Move API root

asmacdo:


kersommoura:
@asmacdo. what about content? Do you need to updated more constants?

asmacdo:
@kersommoura, I'm glad you noticed that, I probably should have mentioned why we don't need to change that also.

Smash uses the `distribution.base_url` to construct the content urls. 
https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/pulp3/file/api_v3/test_download_content.py#L97

The `distribution.base_url` is correct after the pulpcore change, so pulp-smash doesn't even need to know what the content app root is.

PR #970, opened 2018-04-23
Refactor test case for status API

Ichimonji10:
Make test methods more independent. Make assertions more strict,
primarily by using a schema.


Ichimonji10:
Tested against a pair of Pulp 3 applications, on F26 and F27 respectively.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_status.py
```diff
@@ -21,35 +58,41 @@ class StatusTestCase(unittest.TestCase, utils.SmokeTest):
...
         with self.assertRaises(HTTPError):
             self.client.post(STATUS_PATH)
+
+    def verify_get_response(self, status):
+        """Verify the response to an HTTP GET call.
+
+        Verify that several attributes and have the correct type or value.
+        """
+        validate(status, STATUS)
+        self.assertTrue(status['database_connection']['connected'])
```
Ichimonji10: This is an example of an assertion that's being made more strict *without* making use of a schema. Formerly, this assertion read as follows:

```python
self.assertTrue(status['database_connection'])
```

kersommoura: @Ichimonji10, pretty neat. Thanks for updating this.

PR #973, opened 2018-04-24
Add tests to the task system.

kersommoura:
Add basic test coverage to the task system. Add tests related to
filters, read, cancel and delete a task.

Closes:#754


pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,88 @@
...
+            'finished_at': utils.uuid4(),
+            'started_at': utils.uuid4()})
+        self.assertEqual(len(page['results']), 0)
+
+    @selectors.skip_if(bool, 'task', False)
+    def test_05_cancel_tasks(self):
+        """Cancel tasks."""
+        for state in ('running', 'waiting'):
+            for task in self.filter_tasks({'state': state})['results']:
+                self.client.post(urljoin(task['_href'], 'cancel/'))
```
daviddavis: I tested this out and I don't think any tasks are being canceled. I added this to the end of the test to confirm:

```python
self.assertGreater(len(self.filter_tasks({'state': 'canceled'})['results']), 0)
```

kersommoura: @daviddavis, thanks for you suggestion. I improved the assertions.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+    def test_03_read_tasks(self):
+        """Read a task by its worker."""
+        with self.subTest('invalid worker name'):
+            with self.assertRaises(HTTPError):
+                self.filter_tasks({'worker': utils.uuid4()})
+        with self.subTest('valid worker name'):
+            page = self.filter_tasks({'worker': self.task['worker']})
+            self.assertNotEqual(len(page['results']), 0)
+
+    def test_04_filter_invalid_date(self):
```
Ichimonji10: How about a complementary test method that filters by a valid date?

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+        page = self.filter_tasks({
+            'finished_at': utils.uuid4(),
+            'started_at': utils.uuid4()})
+        self.assertEqual(len(page['results']), 0)
+
+    def test_05_cancel_tasks(self):
+        """Cancel tasks."""
+        href_canceled_tasks = []
+        for state in ('running', 'waiting'):
+            for task in self.filter_tasks({'state': state})['results']:
```
Ichimonji10: While this test method does potentially accomplish the goal of cancelling a task and seeing what happens, it could cause some collateral damage. What if we manage to implement parallelized tests? In that case, this would mess up other test cases. Far better to figure out a way to launch a long-running task from within this test method and cancel it. Or, if that can't be done, it's probably best to not perform this test.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+                href_canceled_tasks.append(task['_href'])
+                self.client.post(urljoin(task['_href'], 'cancel/'))
+        for task_href in href_canceled_tasks:
+            self.assertEqual(self.client.get(task_href)['state'], 'canceled')
+
+    def test_06_delete_tasks(self):
+        """Delete tasks."""
+        href_deleted_tasks = []
+        for state in ('canceled', 'completed', 'failed', 'skipped'):
+            for task in self.filter_tasks({'state': state})['results']:
```
Ichimonji10: Again, I worry about collateral damage. How about deleting just the task(s) spawned by this test method? This change in direction does mean less thorough test coverage. But I think the solution is to figure out how to make this test case better.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+        """Read a task by its _href."""
+        task = self.client.get(self.task['_href'])
+        for key, val in self.task.items():
+            if key in _DYNAMIC_TASKS_ATTRS:
+                continue
+            with self.subTest(key=key):
+                self.assertEqual(task[key], val)
+
+    @selectors.skip_if(bool, 'task', False)
+    def test_03_read_tasks(self):
```
Ichimonji10: The `03`, 04`, etc in the test method names suggests that the test case might break if the test methods are re-ordered. But as far as I can tell, that's not the case. Can the test methods be named so as to reflect the requirements of the test case? For example:

```
test_01_create_task
test_02_read_href
test_02_read_valid_name
test_02_read_invalid_name
test_02_read_valid_date
test_02_read_invalid_date
test_03_cancel_task
test_04_delete_task
```

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
kersommoura: @Ichimonji10, I was talking with @daviddavis about this right now. A possible solution is create a large content to be synced, then run `sync_repo` few times. And attempt to cancel the last task. Assuming that the other ones still ongoing. Perhaps, a fixture with files with GB of data can be used for such.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
kersommoura: I tried it before, but tests that involve time is hard to foreseen results. I will try to come up with another test to accomplish this.

kersommoura:
@Ichimonji10, I addressed your comments. 

I removed the test related to cancel a task. I will file a smash issue to add this test later on.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
Ichimonji10: The task created earlier in this test case should work, no?

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,88 @@
...
+    @selectors.skip_if(bool, 'task', False)
+    def test_02_read_valid_date(self):
+        """Read a task by a valid date."""
+        page = self.filter_tasks({'started_at': self.task['started_at']})
+        self.assertGreaterEqual(len(page['results']), 1, page['results'])
+
+    @selectors.skip_if(bool, 'task', False)
+    def test_03_delete_tasks(self):
+        """Delete a task."""
+        state = self.client.get(self.task['_href'])['state']
```
Ichimonji10: Is there any reason to re-read this task? `test_01_create_task` waits for this task to complete. (`json_handler` calls the appropriate polling functions.) I added some debugging code and found that this holds true:

```python
self.task == self.client.get(self.task['_href'])
```

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,88 @@
...
+        """Read a task using a valid worker name."""
+        page = self.filter_tasks({'worker': self.task['worker']})
+        self.assertNotEqual(len(page['results']), 0, page['results'])
+
+    def test_02_read_invalid_date(self):
+        """Read a task by an invalid date."""
+        page = self.filter_tasks({
+            'finished_at': utils.uuid4(),
+            'started_at': utils.uuid4()})
+        self.assertEqual(len(page['results']), 0)
```
Ichimonji10: Can I get a third argument here, please? :smiley:

```python
self.assertEqual(len(page['results']), 0, page['results'])
```

kersommoura:
@Ichimonji10, done.

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,88 @@
...
+"""Test that operations can be performed over tasks."""
+import unittest
+
+from requests import HTTPError
+
+from pulp_smash import api, config, selectors, utils
+from pulp_smash.tests.pulp3.constants import REPO_PATH, TASKS_PATH
+from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
+from pulp_smash.tests.pulp3.utils import get_auth
+from pulp_smash.tests.pulp3.pulpcore.utils import set_up_module as setUpModule  # noqa pylint:disable=unused-import
```
Ichimonji10: Should `noqa` be here?

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_02_read_valid_date(self):
+        """Read a task by a valid date."""
+        page = self.filter_tasks({'started_at': self.task['started_at']})
+        self.assertGreaterEqual(len(page['results']), 1, page['results'])
+
+    @selectors.skip_if(bool, 'task', False)
+    def test_03_delete_tasks(self):
+        """Delete a task."""
+        if self.task['state'] in (
+                'canceled', 'completed', 'failed', 'skipped'):
```
Ichimonji10: Rather than hard-coding this constant here, how about referencing the existing constant?

Also, `self.task` should already be in a completed state. For it to be otherwise would indicate a problem either in Pulp's tasking system or Pulp Smash's logic for interacting with it. How about code like this?

```python
    @selectors.skip_if(bool, 'task', False)
    def test_03_delete_tasks(self):
        """Delete a task."""
        # If this assertion fails, then either Pulp's tasking system or Pulp
        # Smash's code for interacting with the tasking system has a bug.
        self.assertIn(self.task['state'], _P3_TASK_END_STATES)
        self.client.delete(self.task['_href'])
            self.client.get(self.task['_href'])
```

pulp_smash/tests/pulp3/pulpcore/api_v3/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+from pulp_smash import api, config, selectors, utils
+from pulp_smash.tests.pulp3.constants import REPO_PATH, TASKS_PATH
+from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
+from pulp_smash.tests.pulp3.utils import get_auth
+from pulp_smash.tests.pulp3.pulpcore.utils import set_up_module as setUpModule  # pylint:disable=unused-import
+
+_DYNAMIC_TASKS_ATTRS = ('finished_at',)
+"""Task attributes that are dynamically set by Pulp, not set by a user."""
+
+_P3_TASK_END_STATES = ('canceled', 'completed', 'failed', 'skipped',)
```
Ichimonji10: This can be imported from `pulp_smash.api`. I know it's a Bad Idea to import private variables, but given that this module specifically tests the tasking system, the benefit of having a single variable seems to outweigh the improper-ness of accessing private data. The variable could also be made public, if you prefer.

kersommoura: @Ichimonji10, what about move to Pulp3 constants?

PR #974, opened 2018-04-24
Updating file content path

daviddavis:
Per https://pulp.plan.io/issues/3601.


Ichimonji10:
This breaks tests when executed against a freshly provisioned VM. Is there a corresponding Pulp PR?

daviddavis:
So the change is for pulp_file only. The PR is the issue linked above but here's a link:

https://github.com/pulp/pulp_file/pull/81

PR #975, opened 2018-04-25
Rsync tests should succeed even if selinux is disabled

dekimsey:
The rsync tests check for `setsebool`, but don't check if selinux is
enabled before attempting to flip booleans. This generates errors in the
test suite.

Noticed when running test suite in docker with an OSX host.


dekimsey:
Q: Is it problematic to call `config.get_config()` at the module namespace?

dekimsey:
Thansk @Ichimonji10. Sorry for the churn on this, I missed the flake/pep8 tests.

Ichimonji10:
Is it an intentional design goal that Docker hosts not make use of SELinux? Or is the lack of SELinux in Docker hosts a bug?

dekimsey:
> Is it an intentional design goal that Docker containers not make use of SELinux? Or is the lack of SELinux in Docker containers a bug, for which an issue can be filed?

It is my understanding that docker does support selinux on a Linux host. The issue here is on an OSX host, selinux isn't available. I don't know selinux would ever function in a container. 

The best test I could come up with was simply seeing if it was disabled, as the tests are meaningless at that point. If it's set to permissive, they should pass correctly.

dekimsey:
> It's OK to call config.get_config() anywhere.

Hrmm, Travis tests are failing due to being unable to find settings file. Am I doing something incorrect in how I'm getting the cli.Client object?

Ichimonji10:
> > It's OK to call config.get_config() anywhere.
>
> Hrmm, Travis tests are failing due to being unable to find settings file. Am I doing something incorrect in how I'm getting the cli.Client object?

Yes. And I should've been more precise in my original reply.

As written, `config.get_config()` is executed when the module is imported. This is bad. Code should be executed when the user wants it to, not at import time. This is why this idiom shows up in many applications:

```python
def main():
    do_interesting_stuff

if __name__ == '__main__':
    exit(main())
```

The unittest test runner (and any reasonably compatible clone, such as py.test) execute `setUpModule()` very early during the test execution process. It's probably best to insert code which inspects the runtime environment into such a method. That method can then overwrite or update a global variable. Something like this:

```python
_SELINUX_ENABLED = None

def setUpModule():
    global _SELINUX_ENABLED
    _SELINUX_ENABLED = cli.Client(config.get_config()).run(...)
```

Ichimonji10:
> It is my understanding that docker does support selinux on a Linux host. The issue here is on an OSX host, selinux isn't available. I don't know selinux would ever function in an OSX-hosted container.

Gotcha.

As far as I know, Pulp 2 specifically provides a guarantee about SELinux compatibility, and it's geared toward enterprise environments, so I'm wary of any action that might weaken our testing on that front. I'm open to changes that improve compatibility with different environments, but let's be careful about it. 

One solution could be to skip all SELinux tests if Pulp is running inside a container, and if SELinux is disabled. This would allow us to keep current test behaviour on current test platforms. It broadens compatibility without weakening our current testing guarantees.

Here's a simple Bash one-liner that checks to see if a Docker container is in use:

```bash
awk -F : '{print $3}' < /proc/1/cgroup | grep docker
```

This can be translated into something suitable for use within Pulp Smash. (Pulp Smash's CLI client doesn't support Bash tricks like piping without some distasteful gymnastics. It behaves much like Python's `subprocess.run`. Within this test suite, such actions should be done within Python code.)

dekimsey:
Acknowledged, it was due to my use of the decorator to skip the tests. The decorator's `skipIf` is tested early enough that setUpModule doesn't have the right value.

I've switched to simply raising the SkipTest exception instead similar to the `utils.set_up_module` function.

Thanks for the recommendation regarding `make`. I was using the vagrant psmash to test my workflow.

pulp_smash/tests/pulp2/platform/cli/test_selinux.py
```diff
@@ -31,6 +31,12 @@
 """A single line of output from ``ps``."""
 
 
+def setUpModule():  # pylint:disable=invalid-name
+    """Skip tests if the either default requires or selinux are unavailable."""
+    set_up_module()
+    require_selinux()
```
Ichimonji10: Let's say Pulp 2 is installed on Fedora, and for whatever reason, SELinux is disabled. In this case, this module of tests will skip, and neither the Pulp team nor the Pulp QE team will know about the lack of SELinux testing. Given the Pulp team's commitment to supporting SELinux, this omission is dangerous. This is especially true given the rich history of SELinux issues on Pulp 2.

If SELinux tests are to be skipped, please make sure to do so under narrowly defined circumstances. Would the sample code I've provided in https://github.com/PulpQE/pulp-smash/pull/975#issuecomment-384394011 work?

pulp_smash/tests/pulp2/rpm/api_v2/utils.py
```diff
@@ -418,6 +418,12 @@ def set_pulp_manage_rsync(cfg, boolean):
         client.run(sudo + ('which', 'setsebool'))
     except exceptions.CalledProcessError:
         return None
+
+    # setsebool may be available, but selinux could be disabled. Docker on OSX
+    enforcing = client.run(['getenforce']).stdout.strip()
+    if enforcing.lower() == 'disabled':
+        return None
```
Ichimonji10: Beware that this new code substantially changes the behaviour of this function in ways that run contrary to the docstring. But see the other comment for more relevant suggestsions for improvement.

pulp_smash/tests/pulp2/platform/cli/test_selinux.py
dekimsey: Ah, I didn't see your latest comment regarding cgroups. I'll adjust the test to check for the container.

pulp_smash/config.py
```diff
@@ -352,13 +357,14 @@ class PulpSmashConfig(object):
         http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
     """
 
-    def __init__(self, pulp_auth=None, pulp_version=None, systems=None):
+    def __init__(self, pulp_auth=None, pulp_version=None, systems=None, selinux_enabled=None):
```
dekimsey: All of these values default to None. But the pydoc for the class suggests they default to useful values. The current default-enforcing behavior is actually in `.read()`

pulp_smash/pulp_smash_cli.py
```diff
@@ -41,7 +41,7 @@ def settings(ctx):
 
 @settings.command('create')
 @click.pass_context
-def settings_create(ctx):
+def settings_create(ctx):  # pylint:disable=too-many-locals
```
dekimsey: We are now one too many locals, should we make helper function per major config key (pulp, systems)?

Ichimonji10: I'm willing to temporarily mask `too-many-locals`, so long as an issue is filed with a plan for fixing this. One helper function per major config key is a good idea. The Pulp version (2 or 3) may make a difference as to the layout of `systems` in the future, anyway.

dekimsey:
While reviewing the changeset, I noticed there is a [DisableSELinuxMixin](https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/tests/pulp2/rpm/api_v2/utils.py#L146). I don't know if the class should use `cfg.selinux_enabled`. When testing, I didn't run into any issues, but that may just be because it's not being invoked.

Ichimonji10:
This PR lets one declare that SELinux is disabled. If SELinux is declared as being disabled, then the `DisableSELinuxMixin` should never be invoked. And if SELinux is declared as being enabled, then the `DisableSELinuxMixin` can reasonably assume that SELinux is enabled. I can imagine some cases where this "reasonable assumption" won't hold true, but I think it's fine to rely on this assumption for now. Better to make an incremental improvement than revisit those assumptions in this PR.

pulp_smash/config.py
Ichimonji10: The strategy adopted here is good. The weird discrepancy between documented constructor behaviour and actual constructor behaviour can be fixed at a later time. See: https://github.com/PulpQE/pulp-smash/issues/981

tests/test_config.py
```diff
@@ -197,6 +198,7 @@ def test_public_attrs(self):
         """Assert that public attributes have correct values."""
         attrs = config._public_attrs(self.cfg)  # pylint:disable=W0212
         attrs['pulp_version'] = type('')(attrs['pulp_version'])
+        attrs['selinux_enabled'] = type(True)(attrs['selinux_enabled'])
```
Ichimonji10: `type(True)` can be replaced with `bool`. The `type('')` on the line above is a hold-over from when both Python 2 and Python 3 were supported, and `''` might be either a byte or a string.

dekimsey:
> If SELinux is declared as being disabled, then the DisableSELinuxMixin should never be invoked.  

That's what I figured, I'll need to test for this. I'm not sure how to do that.

To test for this, would it be acceptable having `maybe_disable_selinux` raise an Assertion if it's called when selinux is disabled by config?

tests/test_config.py
dekimsey: Ack. I wasn't sure, I thought the construct was odd.

Ichimonji10:
Excellent work. I have just one request, and it's a bit of a pain: rename `PulpSmashConfig.selinux_enabled` to `PulpSmashConfig.pulp_selinux_enabled`. This lets the new variable abide by the same naming scheme as `PulpSmashConfig.pulp_auth` and `pulp_version`.

To understand the naming scheme, consider how one gets at the data within an in-memory config file:

```
>>> import json
>>> with open('/home/ichimonji10/.config/pulp_smash/settings.json') as handle:
...     config_file = json.load(handle)
...
>>> config_file['pulp']['auth']
['admin', 'admin']
>>> config_file['pulp']['version']
'2.16.1'
>>> config_file['pulp']['selinux_enabled']
True
```

A `PulpSmashConfig` object represents the entire configuration file structure. Just as one would execute `config_file['pulp']['auth']`, one can execute `pulp_smash_config.pulp_auth`.

If this is a pain, let me know, and avoid pushing to this branch for a little while. I can create a diff and write a trivial shell script that will apply the diff to this PR.

dekimsey:
> rename PulpSmashConfig.selinux_enabled to PulpSmashConfig.pulp_selinux_enabled.

Will do

Ichimonji10:
This change is backward-compatible with current configuration files. If `selinux_enabled` is not present in a `settings.json` file, then Pulp Smash will default to setting the option to "true" when parsing the file in memory.

Ichimonji10:
Assuming this gets updated and merged, I have a branch ready to go which gets rid of `type('')` calls. See: https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:type

kersommoura:
I have been following this thread. It has been great to see those technical issues addressed.

Ichimonji10:
I've created a branch with the requested diff applied. See https://github.com/PulpQE/pulp-smash/compare/master...Ichimonji10:pr-975. I can verify that `python -m unittest pulp_smash.tests.platform.cli.test_selinux` behaves correctly when "selinux enabled" is:

* present in the config file and true. (Tests run.)
* present in the config file and false. (Tests skip.)
* absent from the config file. (Tests run.)

Ichimonji10:
Pending any last-minute issues, we're going to merge this PR on Monday, with a follow-up commit implementing the diff.

dekimsey:
@Ichimonji10 Sorry I missed the updates to this PR. I saw the approval a few days ago and stopped paying close attention. I check my email somewhat infrequently for my own sanity. Did you want to have these changes squashed? It all looks good to me.

What diff did you want applied? I believe I enabled granting maintainers write access to the PR.

I'll submit a PR to get the default enabled flag in [pulp/devel](https://github.com/pulp/devel/blob/master/scripts/vagrant-setup.sh#L66) fixed to `selinux enabled`.

Ichimonji10:
> I saw the approval a few days ago and stopped paying close attention.

I goofed.

> Did you want to have these changes squashed? It all looks good to me.

This PR has just one commit, so nothing needs to be squashed.

> What diff did you want applied?

https://github.com/PulpQE/pulp-smash/pull/975#pullrequestreview-116084274

> I believe I enabled granting maintainers write access to the PR.

We have a policy of not directly modifying commits. All modifications are done as separate commits. It results in a messier commit history, but it helps avoid some of the "I didn't actually write that" complaints.

dekimsey:
Requested diff is applied.

Issue #987, opened 2018-05-01
Let gen_remote() require a URL


Ichimonji10:
1. Describe what you expected to happen.

[`pulp_smash.tests.pulp3.file.api_v3.utils.gen_remote`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.pulp3.file.api_v3.utils.html#pulp_smash.tests.pulp3.file.api_v3.utils.gen_remote) returns a valid dict for use in creating a remote.

2. Describe what actually happened.

The returned dict is incomplete. A URL must be injected into the returned dict.

3. Describe what version of Pulp and what OS is affected by the issue.

Pulp 3.

4. State steps to re-create the issue.

Execute a call like the following:

```python
client.post(FILE_REMOTE_PATH, gen_remote())  # Boom!
```

This will fail. It would be better if `gen_remote` required a URL:

```python
client.post(FILE_REMOTE_PATH, gen_remote('valid url'))
```

In those cases where one wishes to send a malformed request, the returned dict can be saved:

```python
body = gen_remote('bogus url')
del body['url']
client.post(FILE_REMOTE_PATH, body)  # Boom!
```

5. If there is a related pulp.plan.io issue, link to it, and set the "smash test" field.

6. Choose proper labels. See: [Smash Labels](https://pulp-smash.readthedocs.io/en/latest/about.html#labels)

Done.

PR #988, opened 2018-05-01
Add url as parameter to gen_remote.

kersommoura:
Add an extra parameter to `gen_remote` function. In order to create a
remote a URL is mandatory. Update this function to be according to this.

See: https://pulp.plan.io/issues/3395

Closes:#987


pulp_smash/tests/pulp3/file/api_v3/utils.py
```diff
@@ -4,11 +4,12 @@
...
-    """Return a semi-random dict for use in creating an remote."""
-    return {
-        'name': utils.uuid4(),
-    }
+def gen_remote(url):
+    """Return a semi-random dict for use in creating an remote.
+
+    :param url: The URL of an external content source.
+    """
+    return {'name': utils.uuid4(), 'url': url, }
```
Ichimonji10: This could be written more cleanly without the trailing comma and whitespace:

```python
    return {'name': utils.uuid4(), 'url': url}
```

Issue #990, opened 2018-05-03
get_repo_versions has a misleading name


Ichimonji10:
[`pulp_smash.tests.pulp3.utils.get_repo_versions`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.pulp3.utils.html#pulp_smash.tests.pulp3.utils.get_repo_versions) doesn't return an iterable of repository versions. Instead, it returns a sorted list of repository version hrefs. Do one of the following:

* Rename the function to `get_repo_versions_hrefs`.
* Let the function return an iterable of repository versions.

Ichimonji10:
I'd also consider renaming the function so that `repo` is no longer in its name, e.g. `get_version_hrefs`. The function accepts a `repo` as an argument, which makes the `repo` in its name redundant. There are several other functions in this module which follow the same naming convention, including `get_added_content`, `get_artifact_paths`, and `get_content`.

On the flip side, there are also several functions which *do* include `repo` in their name, including `delete_repo_version`, `get_repo_versions`, and `publish_repo`. This inconsistency is something that should probably be taken care of before we split apart Pulp Smash and make an API.

kersommoura:
@Ichimonji10, I will work on it. Thanks.

Ichimonji10:
:+1: The just-added [`pulp_smash.tests.pulp3.file.api_v3.test_repo_version.FilterRepoVersionTestCase`](http://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.tests.pulp3.file.api_v3.test_repo_version.html#pulp_smash.tests.pulp3.file.api_v3.test_repo_version.FilterRepoVersionTestCase) might benefit from this change. If the function above is configured to return an iterable of repository versions, its `get_repo_versions_attr` function could be simplified.

PR #992, opened 2018-05-07
Rename functions of utils pulp3.

kersommoura:
Few functions had the repo in their names. Simplify their names,
removing the word repo from it.

Rename the following functions:

 1. `get_repo_versions` = `get_version_hrefs`
 2. `sync_repo` = `sync`
 3. `publish_repo` = `publish`
 4. `delete_repo_version` = `delete_version`

Update the logic for `get_version_hrefs`.

Closes:#990

### DIFF
--- a/pulp_smash/tests/pulp3/utils.py
+++ b/pulp_smash/tests/pulp3/utils.py
@@ -151,7 +151,7 @@ def get_plugins(cfg=None):
     }
 
 
-def sync_repo(cfg, remote, repo):
+def sync(cfg, remote, repo):
     """Sync a repository.
 
     :param pulp_smash.config.PulpSmashConfig cfg: Information about the Pulp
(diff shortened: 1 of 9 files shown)


pulp_smash/tests/pulp3/utils.py
```diff
@@ -262,18 +262,19 @@ def delete_orphans(cfg=None):
...
-    return sorted(
-        [version['_href'] for version in versions],
-        key=lambda url: int(urlsplit(url).path.split('/')[-2])
-    )
+    attributes = [
+        version_href['_href']
+        for version_href in client.get(repo['_versions_href'])['results']
+    ]
+    attributes.sort(key=lambda url: int(urlsplit(url).path.split('/')[-2]))
+    return attributes
```
Ichimonji10: +1 Sorting in-place takes about the same amount of time, while using less memory. This avoids unnecessary cache line evictions. (I say this while also recognizing that performance is less important than making code work or making code correct.)

PR #993, opened 2018-05-07
Replace get_version_hrefs() with get_versions()

Ichimonji10:
High-level general-purpose utilities are most useful when they return
complete or near-complete responses from Pulp, instead of attributes
extracted from Pulp's responses. The `get_version_hrefs` function
violates this design principle: it fetches the versions of a repository,
and then returns just the hrefs of each version.

Drop this function in favor of a `get_versions` function, which returns
the complete list of repository versions returned by Pulp. This lets the
function be used in more places, such as in the `get_repo_versions_attr`
function.

In addition, let the revamped function accept a `params` argument, which
can be used to filter which repositories Pulp returns. This lets it
replace the `filter_repo_version` function.

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_orphans.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_orphans.py
@@ -25,7 +25,7 @@ from pulp_smash.tests.pulp3.utils import (
     delete_version,
     get_auth,
     get_content,
-    get_version_hrefs,
+    get_versions,
     sync,
 )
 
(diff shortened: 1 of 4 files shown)


Ichimonji10:
Labeled as [DO NOT MERGE], because I haven't executed the updated tests against a live Pulp system. I've opened the PR anyway, so as to keep track of my work.

Ichimonji10:
I've added debugging information to a few assertions, as per a request.

PR #997, opened 2018-05-10
Add populate_pulp()

Ichimonji10:
This function syncs file content into Pulp, then deletes the repository
and remote and repository used for the task. In a fresh Pulp 3 system,
the result is that orphaned file contents will be present.

Within Pulp Smash, modules may import from siblings or parents, but not
children. This helps to prevent circular imports. Unfortunately,
populate_pulp() imports gen_remote() from a child module. Move
gen_remote() up the namespace hierarchy.

### DIFF
--- a/pulp_smash/tests/pulp3/file/utils.py
+++ b/pulp_smash/tests/pulp3/file/utils.py
@@ -1,9 +1,51 @@
 # coding=utf-8
 """Utilities for tests for the file plugin."""
-from pulp_smash.tests.pulp3 import utils
+from urllib.parse import urljoin
+
+from pulp_smash import api
+from pulp_smash.constants import FILE_FEED_URL
+from pulp_smash.tests.pulp3.constants import (
+    FILE_CONTENT_PATH,
+    FILE_REMOTE_PATH,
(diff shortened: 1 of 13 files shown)


Ichimonji10:
Tested against a pair of Pulp 3 applications.

Ichimonji10:
`populate_pulp` relies on the file plugin being installed, so I've placed it into `pulp_smash.tests.pulp3.file.utils`, not `pulp_smash.tests.pulp3.utils`.

pulp_smash/tests/pulp3/file/utils.py
```diff
@@ -1,9 +1,58 @@
...
+)
+from pulp_smash.tests.pulp3.pulpcore.utils import gen_repo
+from pulp_smash.tests.pulp3.utils import (
+    require_pulp_3,
+    require_pulp_plugins,
+    sync,
+)
+
+
+def gen_remote(url):
```
kersommoura: As we talked, please, move the gen_remote to `pulp3/utils.py` since it is not plugin dependent.

Ichimonji10:
Updated.

PR #998, opened 2018-05-10
Move functions higher up namespace

Ichimonji10:
Move gen_distribution() and gen_repo() up the namespace hierarchy. This
lets tests for different plugins avoid importing from each other. In
this particular case, file plugin tests no longer need to import from
the pulp core tests. This is especially useful if plugins are going to
be pulled out of Pulp Smash and pushed into individual plugins.


dralley:
This is a good change, thanks!

I don't know if perhaps this is already being planned (I have no insight into what the plan is, yet), but one thing that might be done in the future is make the testing helper utilities available at ```pulp_smash.pulp3```, while the tests themselves remain in ```pulp_smash.tests.pulpX```

Ichimonji10:
I strongly expect that we'll blow away `pulp_smash.tests.*`. The details are up for debate, but your proposition is very reasonable.

PR #999, opened 2018-05-10
Drop astroid from dependency list

Ichimonji10:
Pylint depends on astroid. Astroid was added to the dependency list some
time back because of a bug which forced us to pin dependencies. However,
that bug has since been fixed.

Dependency pinning: 5e14a0e66c811ff4f02e67e440a3da42c99c814b

Dependency un-pinning: d07605ff6e331d67ab70d808822e71eee25a6e9c

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ setup(
             'flake8-quotes',
             'pydocstyle',
             'pylint',
-            'astroid',
             # For `make test-coverage`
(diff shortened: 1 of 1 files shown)


kersommoura:
Rebased and closed as https://github.com/PulpQE/pulp-smash/commit/d868b0e0497a0e4a4ad940bb6acc8e788900ac70

PR #1001, opened 2018-05-11
Drop compatibility with old config file schema

Ichimonji10:
The old configuration file schema was deprecated over 14 months ago, on
March 8, 2017. Retaining compatibility with the old configuration file
schema is user-friendly, but hinders additional improvements. Given the
high likelihood that Pulp Smash will become a full-fledged framework in
the near future, now seems like a good time to make a push for having
the best interfaces possible.

See: 5f8838fb7432da9b7cfcb5775a120711d65042f9

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -11,9 +11,8 @@ import collections
 import itertools
 import json
 import os
-import warnings
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


kersommoura:
Rebased and merged as c0fdd818a2ee30d935faeb505483061faaa7ee57

Issue #981, opened 2018-04-27
PulpSmashConfig.__init__'s behaviour doesn't match documentation


Ichimonji10:
1. Describe what you expected to happen.

`PulpSmashConfig` doesn't correctly document the default values for its constructor's arguments. Here's `__init__`'s signature:

```python
__init__(self, pulp_auth=None, pulp_version=None, systems=None)
```

And here's the class's documentation:

```
:param pulp_auth: A two-tuple. Credentials to use when communicating with
    the server. For example: ``('username', 'password')``.
:param pulp_version: A string, such as '1.2' or '0.8.rc3'. Defaults to
    '1!0' (epoch 1, version 0). Must be compatible with the `packaging`_
    library's ``packaging.version.Version`` class.
:param pulp_smash.config.PulpSystem systems: A list of the hosts comprising
    a Pulp application.
```

These are my conclusions:

* I have no idea what `PulpSmashConfig().pulp_auth` defaults to.
* I expect that `PulpSmashConfig().pulp_version == Version('1!0')`
* I have no idea what `PulpSmashConfig().systems` defaults to.

2. Describe what actually happened.

What actually happens when `PulpSmashConfig()` is executed? This:

```
>>> from pulp_smash.config import PulpSmashConfig
>>> cfg = PulpSmashConfig()
>>> cfg.pulp_auth is cfg.pulp_version is None
True
>>> cfg.systems
[]
```

This doesn't match my expectations. As a fix, I would expect the following:

* `PulpSmashConfig` clearly states the default value of all attributes.
* `PulpSmashConfig` abides by the stated default values.

It may also be a good idea to move all "default value" logic from `PulpSmashConfig.read` into `PulpSmashConfig.__init__`. DRY an all that.

3. Describe what version of Pulp and what OS is affected by the issue.

n/a

4. State steps to re-create the issue.

See step 2.

5. If there is a related pulp.plan.io issue, link to it, and set the "smash test" field.

n/a

6. Choose proper labels. See: [Smash Labels](https://pulp-smash.readthedocs.io/en/latest/about.html#labels)

Done.

PR #1004, opened 2018-05-15
Rename "systems" to "hosts" in config file

Ichimonji10:
The configuration file lets one declare information about a Pulp
application and the computers that it's deployed on. A good name to use
for these computers is "hosts." Consider a definition from Wiktionary:

> Any computer attached to a network.

However, the configuration file calls these hosts "systems." Consider a
definition from Wiktionary:

> The set of hardware and software operating in a computer.

"host" seems more appropriate than "system" for several reasons. First,
it makes more semantic sense to talk about computers than to talk about
the hardware and software in a computer. Second, Pulp makes services
available to a network, even if just the local network. If no network is
available, then Pulp can't function. It seems appropriate to use
terminology which reflects this focus.

Rename "systems" to "hosts" in the configuration file. This resolves a
long-standing discrepancy within the code base itself, where both terms
are used to reference the same concept.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -61,7 +61,7 @@ configuration file format is. Consider the example below:
             "version": "2.12.2",
             "selinux enable": true
         },
-        "systems": [
+        "hosts": [
             {
                 "hostname": "pulp.example.com",
                 "roles": {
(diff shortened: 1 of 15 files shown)


Ichimonji10:
This is another good example of a case where there has been some internal technical debt whose cost is low now, but which will rise in cost when the interface is made more public. I can clarify confusion for one new dev at a time without too much issue, but not so easily for a dozen new users.

Things  are going to get a little more substantial after this point. I have a fix for #981 two commits in the future, and as a bonus, the fix should make the code more robust and reduce the number of pointless object instances which need to be created at runtime. Some of the other recently-filed issues should also be easy to fix.

PR #1005, opened 2018-05-16
Re-order PulpSmashConfig.__init__ parameters

Ichimonji10:
Let all application-wide parameters occur before per-host parameters. As
a bonus, this ordering is the same as is shown in documentation.

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -316,12 +316,17 @@ class PulpSmashConfig():
         http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
     """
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I think it might be useful to add a star into the argument list, thus forcing following arguments to be passed by name. Something like this:

```python
def __init__(self, pulp_auth, pulp_version, pulp_selinux_enabled, *, hosts):
```

See [PEP 3102](https://www.python.org/dev/peps/pep-3102/). This would let us add additional system-wide arguments such as `pulp_fips` without breaking existing client code.

Ichimonji10:
Rebased.

PR #1013, opened 2018-05-18
Handle Pulp issue 2334 more gracefully

Ichimonji10:
The Python `DuplicateUploadsTestCase` is affected by Pulp issue 2334.
Rather than skipping the test case outright, and rather than hard-coding
a work-around into Pulp Smash, let the test case conditionally adapt to
the bug.

See: https://pulp.plan.io/issues/2334

See: 84f088d825746f4723e1d8e29be9486c2e75fe58


Ichimonji10:
I've set [Pulp #2334](https://pulp.plan.io/issues/2334) back to the ASSIGNED state. Tested against Pulp 2.16 nightly.

PR #1014, opened 2018-05-18
Revamp PulpSmashConfig object

Ichimonji10:
The `PulpSmashConfig` object has a weird constructor: all arguments are
optional. As a result, it's excessively easy to create a new object that
has an invalid state. For example, this will raise an exception, as
necessary information is not available:

    config.PulpSmashConfig().get_base_url()

From a design perspective, this is bad. No amount of linting can help
discover improperly written code when the underlying code doesn't even
define constructor parameters as required.

Why is this done? Because certain methods, like `read`, are instance
methods. This code needs to work:

    config.PulpSmashConfig().read()

What's to be done? The obvious solution is to change methods which don't
require instance data from instance to class or static methods. This
makes it possible to define constructor parameters as required. And it
opens the door to more aggressive input validation in the future, if so
desired. To summarize: this change leads to more efficient *and* more
correct code.

Given how fundamental this change is, it's also a good time to review
other aspects of `PulpSmashConfig`'s design. And as it happens, there
are a few other improvements to be made:

* There are two types of paths that are especially useful for users of
  the `PulpSmashConfig` class: a path to which one might save settings,
  and a path from which one might load settings. The
  `default_config_file_path` and `get_config_file_path` methods provide
  this information, respectively. These method names aren't terribly
  obvious. Rename them to `get_save_path` and `get_load_path`,
  respectively.
* The `read` method verifies that the file being read from disk is a
  file. This is an overly aggressive check. It prevents interesting use
  cases such as the file on disk being a symlink.
* It's quite possible that a corresponding `write` method might be
  written in the future. But the names "read" and "write" don't convey
  quite the right semantics. The names "load" and "save" are more
  appropriate. Think about playing a video game: one loads and saves a
  game; one does not read and write a game.
* The `services_for_roles` method has a parameter named `roles`. This is
  redundant. Change the method name to `get_services`. This change is in
  line with some other changes that have been made recently.

Fix: https://github.com/PulpQE/pulp-smash/issues/981

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -201,6 +201,8 @@ class Client():
     >>> from pulp_smash.config import PulpSmashConfig
     >>> cfg = config.PulpSmashConfig(
     ...     pulp_auth=('username', 'password'),
+    ...     pulp_version='1!0',
+    ...     pulp_selinux_enabled=True,
     ...     hosts=[
     ...         config.PulpHost(
     ...             hostname='example.com',
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -131,14 +131,10 @@ class Client():  # pylint:disable=too-few-public-methods
     """A convenience object for working with a CLI.
 
     This class provides the ability to execute shell commands on either the
-    local host or a remote host. Here is a pedagogic usage example:
+    local host or a remote host. Here is a typical usage example:
 
     >>> from pulp_smash import cli, config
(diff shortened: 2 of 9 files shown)


Ichimonji10:
Tested by executing a subset of tests, with `python -m unittest discover pulp_smash.tests.pulp2.platform`.

Ichimonji10:
Rebased.

PR #1019, opened 2018-05-22
Amend test to feed valid date to the filter

dralley:


pulp_smash/tests/pulp3/pulpcore/api_v3/test_workers.py
```diff
@@ -90,7 +91,7 @@ def test_03_positive_filters(self):
     def test_04_negative_filters(self):
         """Read a worker with a query that does not match any worker."""
         page = self.client.get(WORKER_PATH, params={
-            'last_heartbeat': utils.uuid4(),
```
dralley: We should use a valid date, but one that will still not match any worker.  datetime.now() + 1 day should be good

PR #1020, opened 2018-05-22
Pass filename when uploading eggs on Python 2.11+

Ichimonji10:
As of Pulp 2.11, the Python plugin requires a filename when uploading an
egg. Update the Python upload test accordingly.

See: https://pulp.plan.io/issues/2334


Ichimonji10:
Tested against Pulp 2.16.1 stable.

Ichimonji10:
Rebased.

Issue #906, opened 2018-03-15
Validation of distribution base_paths


daviddavis:
Ensure that base paths on distributions are validated. Some things to try:

* Base paths should be unique
* Base paths should form valid urls. We're using [Django's URLValidator](https://docs.djangoproject.com/en/2.0/ref/validators/#urlvalidator) for this. One thing that fails is base paths with spaces.
* Base paths shouldn't begin or end with slashes ("/").

https://pulp.plan.io/issues/3412

PR #1023, opened 2018-05-23
Ensure that base_path on distributions are validated.

kersommoura:
Add tests to ensure that `base_path` on distributions are validated.
Assert that `base_path` should be unique, without spaces, and should not
begin or end with slashes.

Closes:#906


pulp_smash/tests/pulp3/pulpcore/api_v3/test_crud_distributions.py
```diff
@@ -78,3 +79,63 @@ def test_05_delete(self):
...
         with self.assertRaises(HTTPError):
             self.client.get(self.distribution['_href'])
+
+
+class DistributionBasePathTestCase(unittest.TestCase):
+    """Test possible values for ``base_path`` on a distribution.
+
+    This test targets the following issues:
+
+    * `Pulp #3142 <https://pulp.plan.io/issues/3142>`_
```
Ichimonji10: This is a typo. This should link to issue 3412.

kersommoura: It is. Fixing right now.

Ichimonji10:
Tested against Pulp 3 from-source installations on Fedora 26 and Fedora 27.

PR #1025, opened 2018-05-24
Deprecate BaseAPITestCase

Ichimonji10:
It encourages fragile clean-up logic and slows down tests by
unnecessarily deleting orphans.


Ichimonji10:
Great idea!

PR #1030, opened 2018-05-25
Move P3_TASK_END_STATES to pulp3 tools namespace.

kersommoura:
Move `P3_TASK_END_STATES` to `pulp3/constants` and adjust where
necessary. This will avoid a constant related to Pulp 3 to live in the
`pulp_smash.api` module.


Ichimonji10:
I generally think that imports should only go up the namespace. For example, it's OK for `pulp_smash.pulp3.*` to import from `pulp_smash.*`, but not the other way around. We could make an exception for this case, though. `pulp_smash.pulp3.constants` shouldn't be importing from `pulp_smash.*`. Thoughts?

kersommoura:
My idea was to move all constants related to pulp 3 to the same namespace.  

Since we are using absolute paths, I did not  take in consideration different levels as you mentioned. 

I will read more about it. We can re-evaluate this pr.

Ichimonji10:
Consolidating constants is a good idea. I'm just mulling over whether this could bite us.

Maybe this constant could be defined as `pulp_smash.api._P3_TASK_END_STATES` and imported into `pulp_smash.pulp3.constants` as `P3_TASK_END_STATES`? Internally, this would ensure that circular imports. And externally, API consumers would get a logical namespace.

kersommoura:
This will solve the problem. Mainly the one that a test is currently importing constants related to pulp 3 from different locations.

I will read more about imports and package levels just to understand it.

I can update the pr with the solution the you mentioned.

kersommoura:
Updated.

PR #1031, opened 2018-05-25
Move require_selinux to pulp_smash namespace.

kersommoura:
Function `require_selinux` can be re-used in other tests later on.
Besides that, the function is not related to a specific Pulp
version.


Ichimonji10:
I'm actually hoping to remove all mentions of unittest from the Pulp Smash framework itself, given that clients may use any test runner that they wish.

As an aside, the `require_selinux` function can be greatly simplified, now that the `PulpSmashConfig` object has been revamped. I think it can be written like so:

```
def require_selinux():
    if not config.get_config().pulp_selinux_enabled:
        raise unittest.SkipTest("Selinux tests disabled by user's settings.json")
```

Given how simple this function is, placing it in the framework doesn't provide a huge amount of value.

PR #1039, opened 2018-05-29
Adjust the call of `super` to Python 3 syntax.

kersommoura:
Pulp-Smash is a Python 3 project. Adjust the calls to `super` to Python
3 syntax.

Standardize the use of `super` in the code base.


kersommoura:
Rebased, and fixed conflicts.

PR #1041, opened 2018-05-29
Update OPENSUSE_FEED_URL

Ichimonji10:
The repodata available at `OPENSUSE_FEED_URL` has changed, and none of
the packages listed in `*updateinfo.xml*` include
`<relogin_suggested>True</relogin_suggested>` or
`<restart_suggested>True</restart_suggested>` tags. As a result,
`pulp_smash.tests.pulp2.rpm.api_v2.test_updateinfo.OpenSuseErrataTestCase`
fails.

Update URL to point at a new OpenSUSE repository that does have those
tags. This is a band-aid solution. The new repository URL will likely
become invalid or change in the future. A more permanent fix is to add
something to the Pulp Fixtures repository.


Ichimonji10:
Tested against Fedora 26.

Ichimonji10:
Re-tested, and still passing.

PR #1043, opened 2018-05-29
Update URL to Puppet forge

Ichimonji10:
The Puppet forge has moved from forge.puppetlabs.com to
forge.puppet.com. Update the corresponding constant appropriately. In
addition, using HTTPS instead of HTTP is a good idea in general.


Ichimonji10:
Tested against Pulp 2.16 nightly on Fedora 26.

kersommoura:
Rebased and merged as 27ee154b5f269cabf776a68a4cc73ce1915c0532

PR #1044, opened 2018-05-29
Update Puppet module referenced by Pulp Smash

Ichimonji10:
According to https://pulp.plan.io/issues/3692#note-2, the pulp/pulp
module has been removed from the Puppet forge. Use a different module
for test purposes.


Ichimonji10:
Tested against Pulp 2.16 nightly on Fedora 26.

PR #1045, opened 2018-05-30
Disable positive worker filtering test

dralley:
This test, also, is falling victim to the indeterministic worker
filtering issue (Python 3.5 only).


pulp_smash/tests/pulp3/pulpcore/api_v3/test_workers.py
```diff
@@ -90,6 +90,8 @@ def test_03_positive_filters(self):
     @selectors.skip_if(bool, 'worker', False)
     def test_04_negative_filters(self):
         """Read a worker with a query that does not match any worker."""
+        if not selectors.bug_is_fixed(3586, self.cfg.pulp_version):
+            raise unittest.SkipTest('https://pulp.plan.io/issues/3586')
```
Ichimonji10: The `skipTest` method is the more canonical way to skip a test method. But it does the same thing, so this is fine:

```python
    def skipTest(self, reason):
        """Skip this test."""
        raise SkipTest(reason)
```

dralley: Cool, thanks.  I just copied it from the test above - should I change it in both places?

PR #1046, opened 2018-05-30
Update `coverage-run` make target

Ichimonji10:
Let the `coverage-run` make target include everything in the
`pulp_smash` directory except for the `pulp_smash/tests` directory. This
rule-based approach is far better than naming every individual module to
be included in the coverage report.


kersommoura:
Rebased and merged as c2ab8fe7781c3ae7f4a579f92ce960243a566d09

PR #1048, opened 2018-06-01
Ensure only certain content is in Pulp

Ichimonji10:
Among other things, the `AutoDistributionTestCase` adds content to Pulp,
randomly selects one of the content units in Pulp, and fetches that
content unit from the current Pulp Fixtures host with roughly the
following logic:

    urljoin(FILE_URL, relative_path_of_chosen_unit)

This works well *if* the chosen unit is available at `FILE_URL`. But
what if some content is chosen, where that content isn't available at
`FILE_URL`? A 404 will result!

As it happens, this is entirely possible. The test case doesn't bother
to clean existing orphans before adding known valid content to Pulp. As
a result, this test case might choose some content that isn't available
at `FILE_URL`. Fix this issue by deleting orphans near the beginning of
the test case.


kersommoura:
Rebased and merged as 7ac375d17c51d7700f7e99da7a8a50b9b7c9632d

PR #1053, opened 2018-06-04
Make api.Client constructor more straightforward

Ichimonji10:
Use straightforward if/else logic to handle all "default to none"
api.Client constructor parameters.


Ichimonji10:

```bash
done
```

```bash
```

PR #1054, opened 2018-06-07
Overhaul task polling functions

Ichimonji10:
Do the following:

* Let the `poll_spawned_tasks` function use a `pulp_smash.api.Client`
  object when polling Pulp's API, rather than directly using the
  underlying Requests library. This makes polling code more concise, and
  it makes some other simplifications possible.
* Don't set a default value for the `pulp_host` parameter. If a user
  passes no value for this parameter, just pass the `None` through to
  `pulp_smash.api.Client`, which will provide a default value if needed.
  In short: this change makes the logic for providing a default value
  for `pulp_host` more DRY.


Ichimonji10:

```bash
done
```

Pulp 3 testing is in progress.

kersommoura:
Tested against Pulp3 installed from source.

``` python
 python -m unittest discover -f pulp_smash.tests.pulp3
```

PR #1056, opened 2018-06-07
Rewrite response handlers

Ichimonji10:
Currently, API response handlers accept two arguments: a
`PulpSmashConfig` object and a `requests.Response` object. Here's a
typical method signature:

    pulp_smash.api.json_handler(cfg, response)

This mostly works fine. But it does mean that the response handlers lose
some context about the state of the calling `pulp_smash.api.Client`
objects. For example, response handlers don't know if the calling API
client has been instructed to poll a specific
`pulp_smash.config.PulpHost`, as defined by the parent's `pulp_host`
attribute.

Re-write the API response handlers, so that the API client object which
is calling them is now passed to them. As an example of how this is
useful, the `_handle_202` function now knows which Pulp host it should
communicate with when polling asynchronous tasks.

This improved level of access also means that response handlers have a
greater ability to screw things up. Rewrite and expand the docstring on
`pulp_smash.api.Client`, in an attempt to ensure that users know
everything they need to know.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -75,12 +75,12 @@ def _check_tasks(tasks, task_errors):
                 raise exceptions.TaskReportError(msg, task)
 
 
-def _handle_202(cfg, response):
+def _handle_202(cfg, response, pulp_host):
     """Check for an HTTP 202 response and handle it appropriately."""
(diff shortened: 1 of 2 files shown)


Ichimonji10:

```bash
done
```

```bash
```

Issue #1055, opened 2018-06-07
Resources should have unique names


kersommoura:
In Pulp 3 resources that have the attribute `name` should be unique. For instance, if a distribution is created with `name` foo, it is not allowed to create another distribution with the same name.

resources = 'distributions', 'publishers', 'remotes', 'repositories'

1. Create a resource.
2. Use the `name` of previous create resource to create a new resource.
3. Assert that an HTTP error was raised.

Related issue: https://pulp.plan.io/issues/3748

PR #1060, opened 2018-06-08
Add test to assert that distribution has unique `name` attribute

ragabala:
Assuming that in Pulp 3 distribution name is a unique attribute, this test
tries to create a new distribution reusing the same name of one created before
and expects for HTTPError meaning the distribution cannot be created if name is
not unique.

See #1055


ragabala:
@Ichimonji10 Hey please review this PR 

Thanks in Advance!

PR #1062, opened 2018-06-08
Make all `gen_` functions to accept `**kwargs`

rochacbruno:
Making `gen_` functions to accept `kwargs` allows the customization
of generated dict. e.g: `repo = gen_repo(name='foo')`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -167,6 +167,7 @@ developers, not a gospel.
     api/tests.test_cli
     api/tests.test_config
     api/tests.test_pulp2_utils
+    api/tests.test_pulp3_utils
     api/tests.test_pulp_smash_cli
     api/tests.test_selectors
     api/tests.test_utils
--- a/docs/api/tests.test_pulp3_utils.rst
+++ b/docs/api/tests.test_pulp3_utils.rst
@@ -1,0 +1,6 @@
+`tests.test_pulp3_utils`
+========================
+
+Location: :doc:`/index` → :doc:`/api` → :doc:`/api/tests.test_pulp3_utils`
+
+.. automodule:: tests.test_pulp3_utils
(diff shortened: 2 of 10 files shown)


rochacbruno:
@Ichimonji10 Python 3.4 doesn't have the `**kwargs` support for dicts. I can change it to use `.update` instead but the code will be more verbose.

Do we really need Python 3.4 support?

Ichimonji10:
Supporting Python 3.4 lets Pulp Smash run natively on RHEL 7. Being able to use `{..., **kwargs}` is nice, but it seems less important than supporting RHEL 7. Please use `.update`.

dralley:
I thought we are using SCL to get Python 3 on RHEL?  If so, can we not simply upgrade the SCL to Python 3.6, which is what is being used in ```pulp/devel```?

Or does RHEL 7 include Python 3 natively now?  I might be behind the times.

Ichimonji10:
The RHEL 7 repositories provide Python 3.4. The SCLs provide Python 3.5 and 3.6, but configuring and using SCLs is a non-trivial task.

dralley:
Cool, I didn't know 3.4 was available nowadays.  Thanks.

Ichimonji10:
By the way, you're correct that we *are* using SCLs to provide Python 3.6 on RHEL. But that's because Pulp 3 requires Python 3.5 or newer.

rochacbruno:
@Ichimonji10 made it compatible with Python 3.4, Travis is passing (the coverage decrease is unrelated)

Ichimonji10:
LGTM. Please squash the two commits into one. The second commit just modifies how the first commit accomplishes its goal.

rochacbruno:
@Ichimonji10 done! I was used to having it squashed by the "squash and merge" button!

kersommoura:
Rebased and closed as 13df4753536227258a44a856c4fab2eb5d5ce5d5

Ichimonji10:
> @Ichimonji10 done! I was used to having it squashed by the "squash and merge" button!

Fair enough. The commit message wouldn't have made much sense in that case, though. :laughing:

PR #1063, opened 2018-06-12
Add new API response handler called page_handler

Ichimonji10:
This response handler is similar to `json_handler`, but:

* It is only compatible with Pulp 3.
* If it detects that the response is a page, it will walk through all
  pages of results, collect the results into a single list, and return
  that list.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -88,6 +88,17 @@ def _handle_202(cfg, response, pulp_host):
             _check_tasks(tasks, ('error',))
 
 
+def _walk_pages(cfg, page, pulp_host):
+    """Walk through pages, yielding the "results" in each page."""
+    client = Client(cfg, json_handler, pulp_host=pulp_host)
+    while True:
+        yield page['results']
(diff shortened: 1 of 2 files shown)


Ichimonji10:

```bash
done
```

```bash
```

kersommoura:
Related to https://github.com/PulpQE/pulp-smash/issues/918

Ichimonji10:
Adding unit tests.

Ichimonji10:
Unit tests added. Re-tested against Pulp 2 and 3 VMs.

kersommoura:
I explored the new `response_handler`. For instance, for Pulp 3, I replaced the `json_handler` to `page_hander`.

Besides that, I forced the pagination to be triggered changing a parameter in Pulp 3 settings, and I ran a few other tests using `json_handler` and `page_handler`.

The new `response_handler` will be very helpful for Pulp 3 tests.

Good job @Ichimonji10 !

Ichimonji10:
Good catch. Updated.

PR #1068, opened 2018-06-14
Let get_content() handle pagination

Ichimonji10:
Let the function return all content, even if there is more than can fit
in one page of results.

See: https://github.com/PulpQE/pulp-smash/issues/918


Ichimonji10:

```bash
done
```

```bash
```

PR #1071, opened 2018-06-14
Let Pulp 3 test_crd_publications use page_handler

Ichimonji10:
Let these tests correctly handle the case where there are so many search
results for a given query that they don't all fit into a single page.


Ichimonji10:

```bash
done
```

```bash
```

kersommoura:
Related to #918

Issue #1033, opened 2018-05-25
Drop unittest from non-test code


Ichimonji10:
Let's search through all non-test code (e.g. `pulp_smash.utils`) and drop references to unittest. We want to let third parties run functional tests using any test runner of their choice, and referencing unittest harms our value proposition.

In cases where we currently raise a `SkipTest` exception, we could try raising a custom exception instead.

Ichimonji10:
As of commit 14973198d144f76a027ffedad7419a64a6a5deeb:

```console
$ git grep --max-depth 2 unittest pulp_smash
pulp_smash/pulp2/utils.py:import unittest
pulp_smash/pulp2/utils.py:class BaseAPICrudTestCase(unittest.TestCase):
pulp_smash/pulp2/utils.py:            'Avoid using BaseAPICrudTestCase. It is coupled to the unittest '
pulp_smash/pulp2/utils.py:            raise unittest.SkipTest('Abstract base class.')
pulp_smash/pulp2/utils.py:class BaseAPITestCase(unittest.TestCase):
pulp_smash/pulp2/utils.py:        https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.addCleanup
pulp_smash/pulp2/utils.py:            'Avoid using BaseAPITestCase. It is coupled to the unittest test '
pulp_smash/pulp2/utils.py:            'logic. If you are using unittest, delete orphans only as needed, '
pulp_smash/selectors.py:    >>> import unittest
pulp_smash/selectors.py:    >>> class MyTestCase(unittest.TestCase):
pulp_smash/selectors.py:    ...     @selectors.require('2.7', unittest.SkipTest)
pulp_smash/selectors.py:    >>> from unittest import SkipTest
pulp_smash/selectors.py:    >>> unittest_require = partial(require, exc=SkipTest)
pulp_smash/selectors.py:            """Wrap a (unittest test) method."""
pulp_smash/selectors.py:    >>> import unittest
pulp_smash/selectors.py:    >>> class MyTestCase(unittest.TestCase):
pulp_smash/selectors.py:    ...     @skip_if(bool, 'my_var', False, unittest.SkipTest)
pulp_smash/selectors.py:    ...     @skip_if(bool, 'my_var', False, unittest.SkipTest)
pulp_smash/selectors.py:    >>> from unittest import SkipTest
pulp_smash/selectors.py:    >>> unittest_skip_if = partial(skip_if, exc=SkipTest)
[...]

PR #1074, opened 2018-06-15
Drop `pulp-smash smoke-tests`, plus backing code

Ichimonji10:
The `pulp-smash smoke-tests` command searches through test cases
packaged with Pulp Smash itself, and prints the dotted path to all test
cases that inherit from the `SmokeTest` class. This is useful, because a
user can run a constant subset of all tests, no matter how large the
test suite grows:

    python -m unittest $(pulp-smash smoke-tests)

This works so long as tests live within a known namespace, and so long
as tests inherit from `unittest.TestCase`. However, Pulp Smash is being
turned into a toolkit, where tests will live in namespaces that the
toolkit doesn't know about, and where test cases may be compatible with
any number of test runner frameworks. Thus, it doesn't make sense to
keep this command around. Drop it.

### DIFF
--- a/pulp_smash/pulp_smash_cli.py
+++ b/pulp_smash/pulp_smash_cli.py
@@ -1,11 +1,10 @@
 # coding=utf-8
 """The entry point for Pulp Smash's command line interface."""
 import json
-import unittest
 
 import click
 
-from pulp_smash import config, exceptions, utils
+from pulp_smash import config, exceptions
 from pulp_smash.config import PulpSmashConfig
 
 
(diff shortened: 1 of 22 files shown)


dralley:
@Ichimonji10 I don't know of anyone using smoke tests currently

kersommoura:
This piece of code was very good to show how run a subset of tests. Perhaps, we can add this as part of the documentation.

kersommoura:
Rebased and merged as 8090d6032eaf10fbd33a56caf2524470c890b5c6

PR #1075, opened 2018-06-15
Drop script that runs Pulp Smash functional tests

Ichimonji10:
The functional tests that are bundled with Pulp Smash in the
`pulp_smash.tests` directory are being moved out of the Pulp Smash
package. Drop a script whose sole purpose is to run these
soon-to-be-gone tests.


Ichimonji10:
The JJB job definitions don't include any references to this script.

PR #1079, opened 2018-06-20
Drop unittest from @require

Ichimonji10:
The `@require` decorator is designed to skip tests under certain
circumstances. This is easiest to do if some assumptions are made about
which test runner is being used. For example, if one assumes that a
unittest-compatible test runner is likely to be used, then it makes
sense for the function to raise `unittest.SkipTest`.

code base. Add an `exc` parameter to the function, which is the

### DIFF
--- a/pulp_smash/selectors.py
+++ b/pulp_smash/selectors.py
@@ -170,44 +170,52 @@ def bug_is_fixed(bug_id, pulp_version):
     return False
 
 
-def require(version_string):
+def require(ver, exc):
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested with `python -m unittest pulp_smash.tests.pulp2.platform.api_v2.test_repository.CreateSuccessTestCase` with the version number set to 2.16.2, 2.7, 2.6, and 2.1.

PR #1080, opened 2018-06-20
Drop unittest from @skip_if

Ichimonji10:
The `@skip_if` decorator is designed to skip tests under certain
circumstances. This is easiest to do if some assumptions are made about
which test runner is being used. For example, if one assumes that a
unittest-compatible test runner is likely to be used, then it makes
sense for the function to raise `unittest.SkipTest`.

code base. Add an `exc` parameter to the function, which is the

### DIFF
--- a/pulp_smash/tests/pulp2/docker/utils.py
+++ b/pulp_smash/tests/pulp2/docker/utils.py
@@ -2,10 +2,12 @@
 """Utilities for Docker tests."""
 import os
 import json
+from functools import partial
+from unittest import SkipTest
 
 from packaging.version import Version
 
-from pulp_smash import cli, utils
+from pulp_smash import cli, selectors, utils
 from pulp_smash.constants import (
     DOCKER_UPSTREAM_NAME,
     DOCKER_UPSTREAM_NAME_NOLIST,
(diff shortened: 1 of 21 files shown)


Ichimonji10:
Tested with:

```bash
python -m unittest discover pulp_smash.tests.pulp2.docker
python -m unittest discover pulp_smash.tests.pulp2.python
```

pulp_smash/tests/pulp2/docker/utils.py
```diff
@@ -57,3 +59,11 @@ def write_manifest_list(cfg, manifest_list):
         )
     )
     return file_path, dir_path
+
+
+skip_if = partial(selectors.skip_if, exc=SkipTest)  # pylint:disable=invalid-name
```
rochacbruno: Any special reason for repeating the definition of this `partial` multiple times instead of having a single `selectors.unittest_skip_if` and reuse it?

Ichimonji10: I'd like to make sure that Pulp Smash can be used with many different test runners. One way of advancing that goal is to make sure that the tools it provides reference unittest as little as possible.

PR #1081, opened 2018-06-20
Drop unittest from Pulp 2 require_* utils

Ichimonji10:
The various `require_*` methods in `pulp_smash.pulp2.utils` are designed
to skip tests under certain circumstances. This is easiest to do if some
assumptions are made about which test runner is being used. For example,
if one assumes that a unittest-compatible test runner is likely to be
used, then it makes sense for the functions to raise
`unittest.SkipTest`.

code base. Add an `exc` parameter to the functions, which is the

### DIFF
--- a/pulp_smash/pulp2/utils.py
+++ b/pulp_smash/pulp2/utils.py
@@ -308,46 +308,58 @@ def pulp_admin_login(cfg):
     ))
 
 
-def require_issue_3159():
+def require_issue_3159(exc):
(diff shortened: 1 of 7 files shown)


Ichimonji10:
Tested with:

```bash
python -m unittest discover pulp_smash.tests.pulp2.platform
python -m unittest discover pulp_smash.tests.pulp2.python
```

Issue #1066, opened 2018-06-12
CVE verification: Ensure last_override_config does not leak sensitive info


bmbouter:
This Pulp CVE describes [issue 3521](https://pulp.plan.io/issues/3521) whereby the previous run's override config can be read as the attribute `last_override_config` on both importers and distributors. The decision (on the issue) was made to remove this attribute altogether due to it's security implications.

Without the fix, when you set an override_config, you'll be able to read that data when viewing the importer or distributor again. With this fix that data will always show up as `'last_override_config': {}`.

This should be verified for both Importers *and* Distributors.

Details on how I tested this during development (with commands) is here: https://pulp.plan.io/issues/3521#note-15

This is a backwards compatible change for 2.16.2 and a release note was added for it.

Issue #763, opened 2017-08-22
Replace parent class BaseAPITestCase by unittest.TestCase


kersommoura:
Few tests fail due to time-out that happening during the execution of tear down.
To reduce the number of calls to `client.delete(ORPHANS_PATH)`, and try to reduce the number of failures. Replace the BaseAPITestCase by unittest.TestCase. 
```
    @classmethod
    def tearDownClass(cls):
        """Delete all resources named by ``resources``."""
        client = api.Client(cls.cfg)
        for resource in cls.resources:
            client.delete(resource)
        client.delete(ORPHANS_PATH)
```

Basically, it will be necessary pass class by class, replace the parent class and do the necessary re-factors.

See also: [#762](https://github.com/PulpQE/pulp-smash/issues/762)

Ichimonji10:
https://github.com/PulpQE/pulp-smash/pull/1025 deprecated `BaseAPITestCase`.

PR #1088, opened 2018-06-25
Test last_override_config, drop a BaseAPITestCase

Ichimonji10:
Rewrite `pulp_smash.tests.pulp2.docker.api_v2.test_sync`:

* Drop `BaseAPITestCase` from the module.
* Assert that `last_override_config` is an empty dict.
* Restructure the tests to be more independent.

See: https://github.com/PulpQE/pulp-smash/issues/763

Fix: https://github.com/PulpQE/pulp-smash/issues/1066


Ichimonji10:
Tested against Pulp 2.16.2 beta on F26, F27, and RHEL 7.

PR #1089, opened 2018-06-25
Change signature of upload_import_erratum()

Ichimonji10:
Update the `upload_import_erratum` function so that it accepts an entire
dict of information about a repository, instead of just a repository
href. This has two benefits:

* Variable passing is easier. Rather than worrying about which bits of
  information are required where, one can typically just pass the
  information returned by Pulp itself.
* The function is now more like other, similar functions, which makes
  usage more intuitive.


Ichimonji10:
Tested against Pulp 2.16.2 beta on F26, F27, and RHEL 7.

PR #1092, opened 2018-06-28
Make PaginationTestCase use small pages

Ichimonji10:
The `PaginationTestCase` verifies that Pulp properly pages results. For
example, if a client GETs the list of repositories and there are more
than 100 results, then Pulp will return only 100 results, plus a link to
the next page of results. There's also a "previous" link. Pulp Smash
assumes that each page will contain 100 elements. This assumption was
valid until today, when the `page_size` parameter was introduced.

Make Pulp Smash take advantage of this parameter, by requesting
relatively small `page_size`. This lets test cases be much faster. This
is of critical importance to the Pulp 3 development team, which runs
tests on Travis hosts, which may have limited CPU resources.

Unfortunately, this change does reduce test coverage. Pagination is
still tested for the following use cases:

* GET a list of repositories.
* GET a list of repository versions.

It is no longer tested for the following use cases:

* GET a list of repository content.
* GET a list of content added to a repository.
* GET a list of content removed from a repository.

Re-adding this test coverage will require a solution such as making the
underlying functions (like `get_content`) more capable, duplicating some
URL-building logic into the test case, or simply being OK with dropping
test coverage.

### DIFF
--- a/pulp_smash/tests/pulp3/file/api_v3/test_pagination.py
+++ b/pulp_smash/tests/pulp3/file/api_v3/test_pagination.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 """Tests related to pagination."""
 import unittest
-from random import randint, sample
+from random import sample
 from threading import Lock, Thread
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
@dralley

Ichimonji10:
Tested against Pulp 3 from-source on Fedora 27.

Ichimonji10:
@CodeHeeler

PR #1094, opened 2018-07-02
Make assertion on prose more lax

Ichimonji10:
`UploadInvalidRPMTestCase` verifies that a certain string is in a prose
error message. That prose error message has changed in the current
development branch of Pulp. Make the test case slightly more lax, so
that it passes on either version of Pulp, all without reducing the
integrity of the test case by an appreciable amount.


Ichimonji10:
This fixes a test failure in Jenkins.

PR #1096, opened 2018-07-03
Skip issues affected by Pulp issue 3816

jeremy-audet:
Issue title:

> Pulp removes current metadata when remove_old_repodata_threshold = 0

See: https://pulp.plan.io/issues/3816


Ichimonji10:
Tested against Pulp 2.16 and Pulp master (to-be 2.17). The test executes on the former and skips on the latter.

PR #1097, opened 2018-07-05
Fix wrong url, magic number

dralley:


pulp_smash/tests/pulp3/file/api_v3/test_download_content.py
```diff
@@ -88,13 +88,15 @@ def test_all(self):
         # Pick a file, and download it from both Pulp Fixtures…
         unit_path = choice(get_content_unit_paths(repo))
         fixtures_hash = hashlib.sha256(
-            utils.http_get(urljoin(FILE_URL, unit_path))
```
dralley: This worked inadvertently, because ```urljoin()``` replaced the filename segment with ```unit_path```

.gitignore
```diff
@@ -14,3 +14,4 @@ __pycache__/
 # Coverage information
 /.coverage
 /htmlcov/
+.pytest_cache/
```
Ichimonji10: This `.gitignore` file only ignores files that might be generated by Pulp Smash or required tooling, such as Python, Sphinx, or coverage. Adhering to this distinction helps avoid the "but what about X cool tool?" discussions. I can add comments to this file to make this clear, if it helps.

Files generated by tools that the user has elected to use can be ignored by the user like so:

```sh
mkdir -p ~/.config/git
cat ~/.config/git/excludesfile <<EOF
# Created by pytest
.pytest_cache/
EOF
touch ~/.config/git/config
git config --global core.excludesfile ~/.config/git/excludesfile
```

Otherwise, this PR looks great.

dralley:
@Ichimonji10 addressed

PR #1099, opened 2018-07-07
Adding logic to skip repoview checks for FIPS enabled machine

ragabala:
Since repoview logic doesn't work in FIPS enabled machine (Issue 3775)
skip logic for repoview checks are implemented for this commit

refer (https://pulp.plan.io/issues/3775)

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -10,7 +10,7 @@ from urllib.parse import urlparse
 
 import requests
 
-from pulp_smash import cli
+from pulp_smash import cli, exceptions
 
 # A mapping between URLs and SHA 256 checksums. Used by get_sha256_checksum().
 _CHECKSUM_CACHE = {}
@@ -84,6 +84,41 @@ def os_is_f27(cfg, pulp_host=None):
         '/etc/redhat-release',
     ))
     return response.returncode == 0
+
+
+def fips_is_supported(cfg, pulp_host=None):
+    """Return ``True`` if the server supports Fips, or ``False`` otherwise.
+
+    :param pulp_smash.config.PulpSmashConfig cfg: Information about the system
+        being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target,
+        instead of the default chosen by :class: pulp_smash.cli.Client`.
+    :return: True of False
+    """
+    try:
+        cli.Client(cfg, pulp_host=pulp_host).run((
+            'sysctl',
+            'crypto.fips_enabled'
+        ))
+    except exceptions.CalledProcessError:
+        return False
+    return True
+
+
+def fips_is_enabled(cfg, pulp_host=None):
+    """Return ``True`` if the Fips is enabled in server, or ``False`` otherwise.
+
+    :param pulp_smash.config.PulpSmashConfig cfg: Information about the system
+        being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target,
+        instead of the default chosen by :class: pulp_smash.cli.Client`.
+    :return: True of False
+    """
+    return cli.Client(cfg, pulp_host=pulp_host).run((
+        'sysctl',
+        '--values',
+        'crypto.fips_enabled'
+    )).stdout.strip() == '1'
 
 
 def uuid4():
--- a/pulp_smash/tests/pulp2/rpm/api_v2/test_repoview.py
+++ b/pulp_smash/tests/pulp2/rpm/api_v2/test_repoview.py
@@ -39,6 +39,8 @@ class RepoviewTestCase(unittest.TestCase):
         cfg = config.get_config()
         if cfg.pulp_version < Version('2.9'):
             self.skipTest('https://pulp.plan.io/issues/189')
+        if utils.fips_is_supported(cfg) and utils.fips_is_enabled(cfg):
+            self.skipTest('https://pulp.plan.io/issues/3775')
 
         # Create a repo, and add content
         client = api.Client(cfg)


pulp_smash/utils.py
```diff
@@ -86,6 +86,23 @@ def os_is_f27(cfg, pulp_host=None):
...
+    :param pulp_smash.config.PulpSmashConfig cfg: Information about the system
+        being targeted.
+    :param pulp_host: A :class:`pulp_smash.config.PulpHost` to target,
+        instead of the default chosen by :class:`pulp_smash.cli.Client`.
+    :returns: True or false.
+    """
+    response = cli.Client(cfg, cli.echo_handler, pulp_host).run((
+        'grep',
+        '1',
+        '/proc/sys/crypto/fips_enabled'
```
Ichimonji10: There's a bunch of subtle problems inherent in checking the value of `/proc/sys/crypto/fips_enabled` in this manner. In order of decreasing importance:

* What would happen if this file contained a value of "10?" This method would return true. However, we don't know the meaning of "10," so why would we go ahead and state that FIPS is enabled when we don't know whether this is true or not?
* Using grep in this manner makes it impossible to distinguish between several different error conditions. These include:

  * `/proc/sys/crypto/fips_enabled` doesn't exist.
  * `/proc/sys/crypto/fips_enabled` exists but is unreadable.
  * `/proc/sys/crypto/fips_enabled` exists and is readable, but doesn't contain "1."

* Some of the files in `/proc/sys` are deprecated. The `sysctl` executable prevents access to deprecated files in this hierarchy. Directly accessing these files bypasses these protections.

pulp_smash/utils.py
```diff
@@ -86,6 +86,23 @@ def os_is_f27(cfg, pulp_host=None):
     return response.returncode == 0
 
 
+def os_is_fips(cfg, pulp_host=None):
```
Ichimonji10: It's impossible for a function to "be FIPS." It would be nicer to have a pair of methods called `fips_is_supported` and `fips_is_enabled`.

Ichimonji10:
Try defining a pair of methods that can be used like so:

```python
if utils.fips_is_supported(cfg) and utils.fips_is_enabled(cfg):
    self.skipTest('https://pulp.plan.io/issues/3775')
```

These methods can be implemented rougly as follows:

```python
def fips_is_supported(cfg, pulp_host=None):
    try:
        cli.Client(cfg, pulp_host=pulp_host).run(('sysctl', 'crypto.fips_enabled'))
    except SomeException:
        return False
    return True

def fips_is_enabled(cfg, pulp_host=None):
    return (cli
        .Client(cfg, pulp_host=pulp_host)
        .run(('sysctl', '--values', 'crypto.fips_enabled'))
        .stdout
        .strip()) == '1'
```

ragabala:
Hi @Ichimonji10 ,

Thanks for those wonderful pointers. As a part of refactoring, I saw that the cli run method doesn't throw any exceptions for 
* /proc/sys/crypto/fips_enabled doesn't exist.
* /proc/sys/crypto/fips_enabled exists but is unreadable.
* /proc/sys/crypto/fips_enabled exists and is readable, but doesn't contain "1."

However it threw return code as non zero and also throws message in the std_err. Hence I have merged the above two methods as a single one. Have also applied your logic of comparing it against the value rather than using grep. Please let me know if we can have this merged.

pulp_smash/utils.py
```diff
@@ -86,6 +86,26 @@ def os_is_f27(cfg, pulp_host=None):
     return response.returncode == 0
 
 
+def fips_is_enabled(cfg, pulp_host=None):
```
Ichimonji10: Please implement as two separate methods, e.g. `fips_is_supported` and `fips_is_enabled`. My developer spidey senses are tingling.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
...
 
 
+def fips_is_supported(cfg, pulp_host=None):
+    """Return ``True`` if the server has crypto fips option available or ``False`` otherwise
+    :param cfg: Information about the system being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target, instead of the default chosen by :class: pulp_smash.cli.Client`.
+    :return: True of False
+    """
+    try:
+       cli.Client(cfg, cli.echo_handler, pulp_host).run((
```
Ichimonji10: Why use `cli.echo_handler`? This callback always returns the responses it processes without checking for non-zero return codes.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
...
+            'sysctl',
+            'crypto.fips_enabled'
+        ))
+    except Exception:
```
Ichimonji10: Don't catch `Exception`! Catch the exception that the `cli.Client` raises when it discovers a non-zero exit code.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
...
+    return True
+
+
+def fips_is_enabled(cfg, pulp_host=None):
+    """Return ``True`` if the server has crypto fips enabled or ``False`` otherwise
+    :param cfg: Information about the system being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target, instead of the default chosen by :class: pulp_smash.cli.Client`.
+    :return: True of False
+    """
+    response = cli.Client(cfg, cli.echo_handler, pulp_host).run((
```
Ichimonji10: Again, don't use `cli.echo_handler`.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
...
+        'sysctl',
+        '--values',
+        'crypto.fips_enabled'
+    ))
+    return response.returncode == 0 and response.stdout.strip() == '1'
```
Ichimonji10: Don't bother checking `response.returncode`. Let the callback response handler worry about checking the response return code. As an aside, this makes it possible to drop the `response` variable and implement this `fips_is_enabled` as a single expression.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
     return response.returncode == 0
 
 
+def fips_is_supported(cfg, pulp_host=None):
+    """Return ``True`` if the server has crypto fips option available or ``False`` otherwise
+    :param cfg: Information about the system being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target, instead of the default chosen by :class: pulp_smash.cli.Client`.
```
Ichimonji10: Wrap text.

pulp_smash/utils.py
```diff
@@ -86,6 +86,36 @@ def os_is_f27(cfg, pulp_host=None):
...
+        ))
+    except Exception:
+        return False
+    return True
+
+
+def fips_is_enabled(cfg, pulp_host=None):
+    """Return ``True`` if the server has crypto fips enabled or ``False`` otherwise
+    :param cfg: Information about the system being targeted
+    :param pulp_host: A :class : `pulp_smash.config.PulpHost` to target, instead of the default chosen by :class: pulp_smash.cli.Client`.
```
Ichimonji10: Wrap text.

pulp_smash/utils.py
```diff
@@ -9,7 +9,7 @@
 from urllib.parse import urlparse
 
 import requests
-
```
Ichimonji10: Please leave this blank line here. From [PEP 8](https://www.python.org/dev/peps/pep-0008/):

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
>
> Imports should be grouped in the following order:
>
> * Standard library imports.
> * Related third party imports.
> * Local application/library specific imports.
>
> You should put a blank line between each group of imports.

pulp_smash/utils.py
```diff
@@ -9,7 +9,7 @@
...
+from pulp_smash import exceptions
 from pulp_smash import cli
```
Ichimonji10: This can be written as:

    from pulp_smash import cli, exceptions

No need to have separate imports.

Issue #980, opened 2018-04-27
pulp_smash_cli.settings_create is at capacity, generating "too-many-locals" warnings


dekimsey:
# Issue 

1. Describe what you expected to happen.

pulp_smash lint tests should pass cleanly

2. Describe what actually happened.

Python lint tests warn (well, after #975 is merged) due to too many locals in the `pulp_smash.pulp_smash_cli.settings_create` function. This change is directly caused by the addition of the `selinux_enabled` flag in #975.

3. Describe what version of Pulp and what OS is affected by the issue.

n/a

4. State steps to re-create the issue.

```
(pulp-smash) [vagrant@pulp2 pulp-smash]$ make lint
flake8 . --ignore E501,F401 --exclude docs/_build
pylint -j 4 --reports=n --disable=I \
        docs/conf.py \
        pulp_smash \
        scripts/run_functional_tests.py \
        setup.py \
        tests
No config file found, using default configuration
************* Module pulp_smash.pulp_smash_cli
R: 44, 0: Too many local variables (16/15) (too-many-locals)

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```
5. If there is a related pulp.plan.io issue, link to it, and set the "smash test" field.

n/a

6. Choose proper labels. See: [Smash Labels](https://pulp-smash.readthedocs.io/en/latest/about.html#labels)

Plan?

# Proposed Solution

Briefly [discussed](https://github.com/PulpQE/pulp-smash/pull/975#pullrequestreview-115732635), but basically we may want to split out the settings_create into smaller functions to build up the larger configuration. (pulp, server, role, etc).

Issue #965, opened 2018-04-23
Modify Pulp Smash for the Pulp 3 transition to RQ 


dralley:
Pulp 3 is planning to switch from tasking using Celery to tasking using RQ.  This means that Pulp will no longer require a message broker like Qpid or RabbitMQ.

This won't require too much modification to Pulp Smash, but the main ones I can think of are:

* There's no need to ask whether the user is using Qpid or RabbitMQ, when using ```pulp-smash settings create``` with Pulp 3

* There's no need for the ```amqp_broker``` section in the ```settings.json```

* The status API page will change.

There is a PR open to make these changes (https://github.com/PulpQE/pulp-smash/pull/960), and this issue might not necessarily be an exhaustive list.

kersommoura:
Besides that, the options `celerybeat`, `mongod`, and `squid` should be evaluate as well.

PR #1101, opened 2018-07-09
Make settings machinery support Pulp 3

Ichimonji10:
When applied, this commit will redefine what a "valid" configuration
file is. Existing configuration files will continue to be valid. But in
addition, entirely new (and similar-looking) configuration files may
also be written.

The old configuration files are suitable for use when testing Pulp 2.
They let one declare information like "host pulp.example.com is running
rabbitmq." The new configuration files are suitable for testing Pulp 3.
They lew one declare information like "host pulp.example.com is running
redis."

In addition, this commit will update two pieces of machinery that are
related to the configuration file. Specifically, this commit will
overhaul the entire CLI interface (as accessed by the `pulp-smash`
executable) and re-implement the `PulpSmashConfig.get_services` method.

Fix: https://github.com/PulpQE/pulp-smash/issues/965

Fix: https://github.com/PulpQE/pulp-smash/issues/980

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -89,7 +89,8 @@ you declare properties of the individual hosts that comprise the Pulp
 application.
 
 Each host must fulfill the "shell" role. In addition, the hosts must
-collectively fulfill the :obj:`pulp_smash.config.REQUIRED_ROLES`.
+collectively fulfill either :obj:`pulp_smash.config.P2_REQUIRED_ROLES` or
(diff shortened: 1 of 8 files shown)


Ichimonji10:
Currently testing against Pulp 2.17.3 nightly with `python -m unittest discover --failfast pulp_smash.tests`.

Tested against Pulp 3 from-source with `python -m unittest discover --failfast pulp_smash.tests`.

Issue #939, opened 2018-04-11
Add authentication as part of API Client


kersommoura:
To override the default authentication system when creating an API Client, the following statement has to be included.

``` python     
api_client = api.Client(cls.cfg, api.json_handler)
api_client.request_kwargs['auth'] = get_auth()
```
This scenario is more common for Pulp3 tests. Add authentication as part of the API Client will avoid these extra lines of code.

Ichimonji10:
We write code as listed above in many tests. But curiously, it's entirely unnecessary. Authentication credentials are already present in API clients, and the `get_auth` function is used to randomly assign either HTTP BASIC or certificate-based credentials to the API client, the idea being that randomly using either kind of credentials may give better test coverage than always using HTTP BASIC credentials.

I'd like to at least rename `get_auth` to `get_random_auth`, which better reflects what's going on. If posible, I'd love to drop many of the `api_client.request_kwargs['auth'] = get_auth()` lines of code from the code base, so that others who read our test code don't get tricked into thinking that this is somehow necessary.

I'm not sure how to solve this issue, but it deserves attention before Pulp Smash becomes a framework.

PR #1102, opened 2018-07-10
Move code for JWT auth out of top level

Ichimonji10:
JWT authentication isn't currently supported by Pulp 3. For more on
this, see Pulp issue 3248. Given this, it's problematic that there is
JWT-related functionality in the top-level namespace of
`pulp_smash.pulp3.utils`. Top-level utilities should be broadly usable,
among other things. A top-level utility that can't even be used is dumb.

Drop all JWT-related code from `pulp_smash.pulp3.utils`. Move some of it
to module `test_auth`. As a bonus, rewrite the tests in that module for
greater robustness and clarity.

See: https://pulp.plan.io/issues/3248

Fix: https://github.com/PulpQE/pulp-smash/issues/939


Ichimonji10:
Tested with `python -m unittest discover --failfast pulp_smash.tests.pulp3.pulpcore` against a Pulp 3 VM with pulpcore and pulp_file installed.

Issue #1100, opened 2018-07-09
Replace the os_is_* functions


Ichimonji10:
The `os_is_*` functions set a bad precedent. There's an enormous number of combinations of system facts that one might want to discover, and having `os_is_*` functions is a recipe for an explosion in number of functions. A framework should provide more generalized tools that individual users can adapt to their needs.

A good solution would be to provide `get_os_release` or `get_redhat_release` functions that would return a dict of key/value pairs. Plugins could then write code like:

```python
def os_is_f27():
    os_release = get_os_release()
    return os_release['ID'] == 'fedora' and os_release['VERSION_ID'] == '27'
```

Or, if the implementation of that becomes problematic, something like this would still be much nicer:

```python
def os_is_f27():
    return get_os_release_id() == 'fedora' and get_os_release_version_id == '27'
```

PR #1103, opened 2018-07-10
Replace os_is_* functions

Ichimonji10:
The `pulp_smash.utils.os_is_f2{6,7}` functions are not suitable for
inclusion within a top-level toolkit. There's an enormous number of
combinations of system facts that one might want to discover, and having
`os_is_*` functions is a recipe for an explosion in number of functions.
A framework should provide more generalized tools that individual users
can adapt to their needs.

Replace these functions with functions that report information from
`/etc/os-release`. This is more suitable for use within a toolkit: there
is a much more constrained set of values that one may wish to get from
`/etc/os-release`.

These functions are still somewhat susceptible to a surface-area
explosion. If users wish to get more and more information from
`/etc/os-release`, then it would be best to create a function which
returns a dict where the keys and values correspond to the keys and
values in `/etc/os-release`. Thankfully, the new functions which were
just added are forward compatible with such a design. If these happens,
the just-added functions could be rewritten like so:

```python
def get_os_release_id(cfg, pulp_host=None):
    raise DeprecationWarning(...)
    return get_os_release(cfg, pulp_host)['ID']
```

Fix: https://github.com/PulpQE/pulp-smash/issues/1100

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -58,19 +58,23 @@ class GetSha256ChecksumTestCase(unittest.TestCase):
         self.assertEqual(checksums[0], checksums[2])
 
 
-class OsIsF26TestCase(unittest.TestCase):
-    """Test :func:`pulp_smash.utils.os_is_f26`."""
+class GetOsReleaseTestCase(unittest.TestCase):
+    """Test the ``get_os_release_*`` functions.
 
-    def test_returncode_zero(self):
-        """Assert true is returned if the CLI command returns zero."""
-        with mock.patch.object(cli, 'Client') as client:
-            client.return_value.run.return_value.returncode = 0
-            response = utils.os_is_f26(mock.Mock())
-        self.assertTrue(response)
+    These tests are very simple: they just make sure that the string returned
+    by the used :class:`pulp_smash.cli.Client` object is stripped and returned.
+    """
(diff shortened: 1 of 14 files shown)


Ichimonji10:
Tested with `python -m unittest discover --failfast pulp_smash.tests.pulp3.pulpcore`.

tests/test_utils.py
```diff
@@ -58,19 +58,23 @@ def test_all(self):
...
-            response = utils.os_is_f26(mock.Mock())
-        self.assertTrue(response)
+            client.return_value.run.return_value.stdout = ' fedora '
+            response = utils.get_os_release_id(mock.Mock())
+        self.assertEqual(response, 'fedora')
 
-    def test_returncode_nonzero(self):
-        """Assert false is returned if the CLI command returns non-zero."""
+    def test_get_os_release_version_id(self):
+        """Test :func:`pulp_smash.utils.get_os_release_id`."""
```
ragabala: This docstring should be """Test :func:`pulp_smash.utils.get_os_release_version_id`."""

Ichimonji10:
Updated.

ragabala:
Thanks @Ichimonji10 , This looks good to me :smiley:

Issue #1104, opened 2018-07-10
Fix cross-plugin imports


Ichimonji10:
As a result of https://github.com/PulpQE/pulp-smash/pull/1098, some modules in `pulp_smash.tests.pulp3.pulpcore` now import from `pulp_smash.tests.pulp3.file`. This is problematic, as the tests in each of the different plugins will go into different repos, and will not be able to import from each other. The easiest solution is to move and/or copy some utilities from `pulp_smash.tests.pulp3.file.*` to `pulp_smash.tests.pulp3.pulpcore.*`. More elegant would be if one plugin could try to import code from the other. One way or another, this needs to be fixed.

PR #1108, opened 2018-07-16
Fix cross-plugin imports

Ichimonji10:
`pulp_smash.tests.pulp3.pulpcore` should not import from
`pulp_smash.tests.pulp3.file`. This is because these two different
packages will soon be moved into separate repositories that don't have
access to each others' namespaces.

Fix: https://github.com/PulpQE/pulp-smash/issues/1104


Ichimonji10:
Tested with `python -m unittest discover pulp_smash.tests.pulp3.pulpcore.api_v3.plugin_involved`.

dralley:
LGTM

Issue #1035, opened 2018-05-25
Improve unit test coverage percentage


Ichimonji10:
Pulp Smash is being converted from a test suite to a framework for writing tests. As a result, breaking changes in the code are far more problematic. Unit tests are perfect for detecting interface changes. Let's raise unit test coverage to at least 90%.

kersommoura:
As I suggestion we should use `make test-coverage` to see areas that require unittests.

PR #1110, opened 2018-07-19
Adding Unit Tests for testing Skip If selectors

ragabala:
Adding Unit tests for checking the Skip If selectors in pulp-smash. This is done as a part of increasing code coverage in pulp-smash test suite.

refer  #1035

### DIFF
--- a/tests/test_selectors.py
+++ b/tests/test_selectors.py
@@ -106,3 +106,33 @@ class BugIsFixedTestCase(unittest.TestCase):
             get_bug.side_effect = requests.exceptions.ConnectionError
             with self.assertWarns(RuntimeWarning):
                 selectors.bug_is_fixed(None, ver)
+
+
+class SkipIfConditionTestCase(unittest.TestCase):
+    """Test :meth: `pulp_smash.selectors.skip_if`."""
+
+    class DummyTestClass:
+        """A Dummy Test Class."""
+
+        def __init__(self):
+            """Set ``self.element`` for use by ``@skip_if``."""
+            self.element = 'test'
+
+        @selectors.skip_if(bool, 'element', True, unittest.SkipTest)
+        def test_should_skip(self):
+            """``skip_if`` should skip this method."""
+            pass
+
+        @selectors.skip_if(bool, 'element', False, unittest.SkipTest)
+        def test_should_run(self):
+            """``skip_if`` should not skip this method."""
+            pass
+
+    def test_skip_true(self):
+        """Make ``@skip_if`` raise an exception."""
+        with self.assertRaises(unittest.SkipTest):
+            self.DummyTestClass().test_should_skip()
+
+    def test_skip_false(self):
+        """Make ``@skip_if`` continue without raising an exception."""
+        self.assertIsNone(self.DummyTestClass().test_should_run())


tests/test_selectors.py
```diff
@@ -106,3 +106,35 @@ def test_connection_error(self):
...
+
+            def __init__(self):
+                self.element = 'test'
+
+            @selectors.skip_if(bool, 'element', True, unittest.SkipTest)
+            def dummy(self):  # pylint:disable=missing-docstring,no-self-use
+                print("doesn't print")
+                return True
+
+        self.assertRaises(unittest.SkipTest, DummyClass().dummy)
```
Ichimonji10: Clever. I didn't know one could pass a callable to `assertRaises`.

ragabala:
Hi @Ichimonji10 @rochacbruno ,

Have made the requested changes. Please let me know if this works fine

tests/test_selectors.py
```diff
@@ -106,3 +106,30 @@ def test_connection_error(self):
             get_bug.side_effect = requests.exceptions.ConnectionError
             with self.assertWarns(RuntimeWarning):
                 selectors.bug_is_fixed(None, ver)
+
+
+class SkipIfConditionTestCase(unittest.TestCase):
+    """Test :meth: `pulp_smash.selectors.skip_if`."""
+
+    class _DummyClass:  # pylint:disable=missing-docstring,too-few-public-methods
```
Ichimonji10: We have no idea what sort of private attributes `SkipIfConditionTestCase` inherits from `unittest.TestCase`. Adding an additional private attribute should be done with care to avoid name conflicts.

tests/test_selectors.py
```diff
@@ -106,3 +106,30 @@ def test_connection_error(self):
...
+
+        def __init__(self):
+            self.element = 'test'
+
+        @selectors.skip_if(bool, 'element', True, unittest.SkipTest)
+        def dummy_should_skip(self):  # pylint:disable=missing-docstring,no-self-use
```
Ichimonji10: `pylint:disable` pragmas should be used with care, because - by definition - they ignore warning signs!

tests/test_selectors.py
```diff
@@ -106,3 +106,30 @@ def test_connection_error(self):
...
+            pass
+
+        @selectors.skip_if(bool, 'element', False, unittest.SkipTest)
+        def dummy_should_run(self, bool_val):  # pylint:disable=missing-docstring,no-self-use
```
Ichimonji10: Why does this method need to accept a `bool_val` argument? So that the `test_skip_false` method can make an assertion, right? I understand, but having an assertion just for having an assertion isn't helpful. Sometimes, the mere fact that a method executes is itself the test.

tests/test_selectors.py
```diff
@@ -106,3 +106,30 @@ def test_connection_error(self):
...
+            return bool_val
+
+    def test_skip_true(self):
+        """Assert the skip_if decorator skips test if conditions is satisfied."""
+        self.assertRaises(unittest.SkipTest, self._DummyClass().dummy_should_skip)
+
+    def test_skip_false(self):
+        """Assert the skip_if decorator doesn't skips test if conditions is not satisfied."""
+        for test_val in (True, False):
```
Ichimonji10: This for loop doesn't help to prove that `@skip_if` works correctly, and it does make the test method more complicated.

ragabala:
Hi @rochacbruno,
Can you please check if this can be merged. Did the required changes that @Ichimonji10 asked.

Thanks !!!

PR #1112, opened 2018-07-19
Improve definition of CalledProcessError

Ichimonji10:
The `CalledProcessError` exception assumes that it is passed four
arguments. This is a recipe for bugs. Make its `__init__` require four
arguments.


Ichimonji10:
This fixes the linting error that's failing all of the other PRs.

Issue #1111, opened 2018-07-19
Test 'docker pull' from Crane


dkliban:
1. Create a docker repository with a docker_importer that has the following values: feed=https://registry-1.docker.io, upstream-name=busybox

2. Sync repository 

3. Publish repository

4. Perform calls needed to simulate docker pull.

ragabala:
Hi ,

These are requests that were done to download the docker content from pulp
`curl -L -k -X GET <pulp-host-name>:5000/v2/mydocker2/manifests/latest`
response output example : 
```
'fsLayers': [{'blobSum': 'sha256: ...'},
  {'blobSum': 'sha256:...'},
  {'blobSum': 'sha256:...'},
  ...
'}],
```
This lists a json with information about all the blobs that are to be downloaded as a part of the image.
 
to download each of these blobs to build the image,  the following URL is hit recursively with the blob shasum that were returned as a part of the first request.
`curl -L -X GET <pulp-host-name>:5000/v2/mydocker2/blobs/sha256:<shasum> >> docker.image`

response : The response for the above url contains **binary object**

Ichimonji10:
Is this fixed by https://github.com/PulpQE/Pulp-2-Tests/pull/4 ?

rochacbruno:
@ragabala can we close this?

ragabala:
Thanks @rochacbruno 

Closing this one as we have automated and tested this.

PR #1113, opened 2018-07-19
Adding Test Case for testing docker pull again Crane

ragabala:
This testcase is added to ensure that the docker pull on crane
repositories work. This test case is based on emulating the docker pull
requests and calling the same urls and checking whether content gets
downloaded.

Refer `pulp plan #3638`
Also Refer #1111


ragabala:
Closing this as this has to be implemented in pulp 2 tests

PR #1127, opened 2018-08-11
Minor cleanup

dralley:
This is mostly a preference thing, but it makes more sense this way
personally.


pulp_smash/pulp3/utils.py
```diff
@@ -22,7 +22,7 @@ def require_pulp_3(exc):
         https://docs.python.org/3/library/functools.html#functools.partial
     """
     cfg = config.get_config()
-    if cfg.pulp_version < Version('3') or cfg.pulp_version >= Version('4'):
```
dralley: In Python, you can chain comparisons like this https://docs.python.org/3.5/reference/expressions.html#comparisons

PR #1130, opened 2018-08-27
Add a utility function for getting the "content_summary" from Pulp

dralley:
Pulp has a "content_summary" field on repository versions which provides an easy way to see how many of each content type are in the repository.  It looks like (e.g.) {'file': 5, 'rpm': 10, 'errata': 2}. This utility function provides an easy way to get these counts without manually taking a len() of the list content, and is also necessary for plugins which have multiple content types, such as the RPM plugin with both RPMs and Errata.  It avoids the issue of all different types of content being mixed together when taking counts.


kersommoura:
@dralley, please, add a commit message.

dralley:
@kersommoura done

PR #1133, opened 2018-09-09
Make Pulp 3 sync() function accept `**kwargs`

gmbnomis:
As (some) plugins support "mirror" parameters now, allow to call sync() with
customized parameters.

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -64,7 +64,7 @@ def get_plugins(cfg=None):
     return {version['component'] for version in status['versions']}
 
 
-def sync(cfg, remote, repo):
+def sync(cfg, remote, repo, **kwargs):
     """Sync a repository.
 
     :param pulp_smash.config.PulpSmashConfig cfg: Information about the Pulp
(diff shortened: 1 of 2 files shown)


pulp_smash/pulp3/utils.py
```diff
@@ -76,9 +76,9 @@ def sync(cfg, remote, repo):
         a call report.
     """
     client = api.Client(cfg, api.json_handler)
-    return client.post(
-        urljoin(remote['_href'], 'sync/'), {'repository': repo['_href']}
-    )
+    data = {'repository': repo['_href']}
+    data.update(kwargs)
```
rochacbruno: Can you please include in the docstring a sentence saying that `kwargs` will contribute to `repository` data.

```py
"""
...
:param kwargs: Keyed arguments to be merged in to repository data.
"""
```

pulp_smash/pulp3/utils.py
```diff
@@ -76,9 +76,9 @@ def sync(cfg, remote, repo):
...
+    return client.post(urljoin(remote['_href'], 'sync/'), data)
```
rochacbruno: pulp-smash coverage is decreasing so it would be nice if we can have a simple test case for this function added to https://github.com/PulpQE/pulp-smash/blob/master/tests/test_pulp3_utils.py

No need to perform the post (so we can use one of the mock examples) and assert that the right data is being posted.

gmbnomis:
@rochacbruno Thanks for the quick feedback! I pushed a new version that should address your comments.

PR #1146, opened 2018-10-25
Add ServiceManager.is_active to check if a worker is active.

rochacbruno:
Demanded by pulp/pulp#3720 to check whether a pulp worker is running on systemd.

example:

```py
if not ServiceManager(...).is_active(['pulp_worker@*']):
    raise unittest.SkipTest('Requires pulp workers running on systemd')
```

It is using the return code by `service status` or `systemctl is-active`
if != 0 then the methods returns False.

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -359,6 +359,22 @@ class BaseServiceManager(metaclass=ABCMeta):
         cmd = ('systemctl', 'restart') + tuple(services)
         return (client.run(cmd, sudo=True),)
 
+    @staticmethod
+    def _is_active_sysv(client, services):
+        with contextlib.suppress(exceptions.CalledProcessError):
+            return tuple((
+                client.run(('service', service, 'status'), sudo=True)
+                for service in services
+            ))
+        return False
+
+    @staticmethod
+    def _is_active_systemd(client, services):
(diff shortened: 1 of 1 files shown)


pulp_smash/cli.py
```diff
@@ -359,6 +359,22 @@ def _restart_systemd(client, services):
...
+                client.run(('service', service, 'status'), sudo=True)
+                for service in services
+            ))
+        return False
+
+    @staticmethod
+    def _is_active_systemd(client, services):
+        with contextlib.suppress(exceptions.CalledProcessError):
+            cmd = ('systemctl', 'is-active') + tuple(services)
+            return (client.run(cmd, sudo=True),)
```
rochacbruno: This will return exit code `0` if any of the services are active examples 

```bash
$ systemctl is-active pulp_worker@*
$ echo $?
0  # if any of @{1..m} is running

$systemctl is-active pulp_worker@1 pulp_worker@2 foobar
$ echo $?
0  # if any of above is running

$ systemctl is-active foobar
$ echo $?
3  # specific service error because foobar is not active, is this case we suppress the error and return False
```

kersommoura:
@rochacbruno, as we talk later we can update those methods.  And support `systemd` only.

PR #1149, opened 2018-10-27
Added `raise_if_unsupported` on Packagemanager

rochacbruno:
Related to pulp/pulp_file#129

```python
pm = PackageManager(cfg)
pm.raise_if_unsupported(unittest.SkipTest, 'Test requires yum/dnf')
# will raise and skip if not yum or dnf
# else test continues
pm.install('foobar')
```

or

```python
pm = PackageManager(cfg, raise_if_unsupported=(unittest.SkipTest, 'Test requires yum'))
# will raise and skip if not yum or dnf
# else test continues
pm.install('foobar')
```

The main idea is to replace this:

```py
try:
    PackageManager(cgf).name
except exceptions.NoKnownPackagemanager:
    raise unittest.SkipTest("This test requires dnf/yum")
```

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -707,13 +707,22 @@ class PackageManager:
 
     :param pulp_smash.config.PulpSmashConfig cfg: Information about the target
         host.
+    :param tuple raise_if_unsupported: a tuple of Exception and optional
+        string message to force raise_if_unsupported on initialization::
+
+          pm = PackageManager(cfg, (unittest.SkipTest, 'Test requires yum'))
+          # will raise and skip if unsupported package manager
+
+        The optional is calling `pm.raise_if_unsupported` explicitly.
     """
 
-    def __init__(self, cfg):
+    def __init__(self, cfg, raise_if_unsupported=None):
         """Initialize a new object."""
         self._cfg = cfg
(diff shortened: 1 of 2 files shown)


pulp_smash/cli.py
```diff
@@ -722,6 +731,25 @@ def name(self):
             self._name = self._get_package_manager(self._cfg)
         return self._name
 
+    def raise_for_status(self, exc, message='Unsupported package manager'):
```
kersommoura: I would like to suggest for us to update this function name. `raise_for_status` implies that a certain status was verified and that  is the reason for failure. Perhaps, I am so used to this function name in the context of a HTTP request. Just a suggestion.

rochacbruno: @kersommoura naming is hard :( 

```
This is inspired by `requests.raise_for_status` and in this
        case status can be `Known` or `Not Known` package manager.
```

I named `raise_for_status` as we already know what the function does because of requests, in this case the implicit status would be `supported` and `unsupported` package manager.

I had no other idea to name this function, do you have a suggestion?

- raise_for_support
- raise_if_unsupported
- raise_on_error

rochacbruno:
Going to name it `PackageManager.raise_if_unsupported(Exception, message)`

PR #1152, opened 2018-11-26
Fix pylint 2.2.0 errors

kersommoura:
pylint newest version 2.2.0  doesn't allow `simplifiable-if-expression`. This
commit adds an ignore statement in the MakeFile.

See: https://pypi.org/project/pylint/2.2.0/


kersommoura:
@ragabala, I fixed the newly introduced errors.

PR #1154, opened 2018-11-28
API Changes

dralley:
The schema looks like this:

```
"_content": {
    "file": "/pulp/api/v3/content/file/files/?repository_version=..."
},
"_content_added": {
    "file": "/pulp/api/v3/content/file/files/?repository_version_added=..."
},
"_content_removed": {}
```

Change the get_*content() functions to return a dictionary of types
matched to the respective list of content of that type.

Fix the get_artifact_paths() function to use those new functions
properly.

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 """Utility functions for Pulp 3 tests."""
+from collections import defaultdict
 import warnings
 from urllib.parse import urljoin, urlsplit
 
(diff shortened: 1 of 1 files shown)


dralley:
@kersommoura , PyLint doesn't seem to distinguish between constants and functions defined as variables.  What would you recommend I do, put in ignores for the warnings or rewrite this code somehow?

dralley:
The rationale for using defaultdicts instead of normal dicts in those few places is so that you don't have to deal with a KeyError when using the get_*_content functions with a type that doesn't exist in the resultant dict.  

For example you might want to do this

```
self.assertEqual(len(get_added_content(repo)['python']), 0)
```

If there are no packages in the added content, then this would raise a KeyError.  But with defaultdict it returns an empty list, and the test works fine.

kersommoura:
@dralley, I will take a look into this right now.

kersommoura:
Those changes will change how pulp-smash parses some responses from Pulp 3. However, those changes are necessary since Pulp API itself is changing as well.

Those changes were discussed on pulp-dev mail list.

kersommoura:
See: http://pylint.pycqa.org/en/latest/user_guide/options.html#default-behavior

kersommoura:
Those changes were tested with the new API layout. 

See: https://github.com/pulp/pulp_file/pull/133/files

kersommoura:
@dralley, I reviewed those changes related to the functional tests. I left minor comments.

PR #1155, opened 2018-12-05
Remove 'notes' field from gen_repo()

dralley:
Notes is being removed until we can better evaluate the design.


dralley:
This can be merged whenever you're ready

PR #1156, opened 2018-12-10
Add download policies constant

bmbouter:
https://pulp.plan.io/issues/4243
re #4243

### DIFF
--- a/pulp_smash/pulp3/constants.py
+++ b/pulp_smash/pulp3/constants.py
@@ -20,6 +20,8 @@ CONTENT_PATH = urljoin(BASE_PATH, 'content/')
 
 DISTRIBUTION_PATH = urljoin(BASE_PATH, 'distributions/')
 
+DOWNLOAD_POLICIES = ['immediate', 'on_demand', 'streamed']
+
 JWT_PATH = urljoin(BASE_PATH, 'jwt/')
 
 MEDIA_PATH = '/var/lib/pulp'


pulp_smash/pulp3/constants.py
```diff
@@ -20,6 +20,8 @@
 
 DISTRIBUTION_PATH = urljoin(BASE_PATH, 'distributions/')
 
+DOWNLOAD_POLICIES = ['immediate', 'on_demand', 'streamed']
```
kersommoura: Did we rename `cahe-olny`?
```python
DOWNLOAD_POLICIES = ['cache_only', 'immediate', 'on_demand']
```
Constant on [Pulp file](https://github.com/pulp/pulp_file/blob/c7a90288e96d45faac191809cc9be8cca94a2117/pulp_file/tests/functional/constants.py#L12)

bmbouter: I did so in another PR but we can do this in any order/way you recommend.

https://github.com/pulp/pulp_file/pull/140/files#diff-2d4578fd30e72ae31e7baef653596f7fL12

bmbouter:
@pulp/QE I'd like to get this merged and released. I just realized it's failing these tests here: https://github.com/ATIX-AG/pulp_gem/pull/14

bmbouter:
Oh this is merged already, my page wasn't refreshed. When can this be released to pypi?

rochacbruno:
@bmbouter released

pulp_smash/pulp3/constants.py
rochacbruno: https://github.com/PulpQE/pulp-smash/issues/1157

bmbouter:
@rochacbruno Thank you!

Issue #1157, opened 2018-12-14
Constants should be imutable


rochacbruno:
Here and on plugin repositories we see some mutable variables defined in constant files.

like this:

https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/pulp3/constants.py#L23

```py
DOWNLOAD_POLICIES = ['immediate', 'on_demand', 'streamed']
```

It is problematic as on test cases those data can be easily changed and because of that we see sometimes the use of the `copy()` method.

Ideally on `constants.py` across any repo we must define sequences as `tuple` and key/pairs as `MappingProxyType` so we never allow that data to be modified.

When needed the tester should create a new list based on the tuple or copy the Mapping as a new dictionary

kersommoura:
This issue was already addressed. See: https://github.com/PulpQE/pulp-smash/pull/1158

PR #1158, opened 2018-12-17
Update Pulp 3 constant to be a tuple

kersommoura:
Update the constant `DOWNLOAD_POLICIES` to be a tuple.

See: https://github.com/PulpQE/pulp-smash/issues/1157


bherrin3:
Might wanna get @bmbouter to ensure there wasn't a reason he made this a list instead of a tuple.

Else, :+1:

bmbouter:
No reason to use a list over a tuple. I used a list continuing what was already present in the plugin code I ported it from. +1 to a tuple.

PR #1159, opened 2018-12-17
Revert "API Changes"

kersommoura:
Reverts PulpQE/pulp-smash#1154


kersommoura:
I did not see that there were 2 commits on the previous PR. We always merge one commit.

PR #1160, opened 2018-12-17
API Changes

dralley:
The schema looks like this:

"_content": {
    "file": "/pulp/api/v3/content/file/files/?repository_version=..."
},
"_content_added": {
    "file": "/pulp/api/v3/content/file/files/?repository_version_added=..."
},
"_content_removed": {}

Change the get_*content() functions to return a dictionary of types
matched to the respective list of content of that type.

Fix the get_artifact_paths() function to use those new functions
properly.


kersommoura:
See: https://github.com/PulpQE/pulp-smash/pull/1154

The previous PR was reverted since it was composed by 2 commits.

PR #1166, opened 2019-02-06
Introduced the Pulp 3 'content' role.

rochacbruno:
Added `content` as a valid role for config hosts
Added helper methods in `cfg` to get content_host Base URL.
Added `content` as a new option on `pulp-smash settings create`

Note: In cases where `content` role is not defined then the `api`
host is assumed as the content host.

https://docs.pulpproject.org/en/3.0/nightly/components.html#content-serving-application

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -54,6 +54,14 @@ def _get_pulp_3_api_role():
     return api_role
 
 
+def _get_pulp_3_content_role():
+    """Return a schema definition for the Pulp 3 "content" role."""
+    content_role = _get_pulp_3_api_role()
+    content_role['required'].remove('service')
+    content_role['properties']['service'] = {'type': 'string'}
+    return content_role
+
+
 # `get_config` uses this as a cache. It is intentionally a global. This design
 # lets us do interesting things like flush the cache at run time or completely
 # avoid a config file by fetching values from the UI.
@@ -90,6 +98,14 @@ P3_REQUIRED_ROLES = {
     'shell',
 }
 """The set of roles that must be present in a functional Pulp 3 application."""
+
+P3_OPTIONAL_ROLES = {
+    'content',
+}
+"""Additional roles that can be present in a Pulp 3 application."""
+
+P3_ROLES = P3_REQUIRED_ROLES.union(P3_OPTIONAL_ROLES)
+"""The set of all roles that can be present in a Pulp 3 application."""
 
 JSON_CONFIG_SCHEMA = {
     'title': 'Pulp Smash configuration file',
@@ -186,6 +202,9 @@ JSON_CONFIG_SCHEMA = {
                     'type': 'object',
                     'properties': {
                         'api': {'$ref': '#/definitions/pulp 3 api role'},
+                        'content': {
+                            '$ref': '#/definitions/pulp 3 content role'
+                        },
                         'pulp resource manager': {
                             '$ref': '#/definitions/pulp resource manager role'
                         },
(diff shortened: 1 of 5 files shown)


pulp_smash/config.py
```diff
@@ -73,6 +81,7 @@ def _get_pulp_3_api_role():
 P2_OPTIONAL_ROLES = {
     'pulp cli',
     'squid',
+    'content',
```
kersommoura: Is this one supposed to be part of Pulp 2? Can we update this to make a bit clear about what Pulp version?

kersommoura: I will get this code locally and review. Just a minor comment when skimming through the code.

rochacbruno: @kersommoura naming problem, both pulp2 and pulp3 uses the same `P2_` prefixed variables.

kersommoura: @rochacbruno, as we talked this change will require changes in the CLI as well.

Makefile
```diff
@@ -49,7 +49,7 @@ lint: lint-flake8 lint-pylint
 # E501 and F401 are ignored because Pylint performs similar checks.
 # W504 ignored since it requires line breaks after binary operators.
 lint-flake8:
-	flake8 . --ignore E501,F401,W504 --exclude docs/_build
+	flake8 . --ignore E501,F401,W504 --exclude docs/_build,.eggs
```
rochacbruno: Python 3.7 is creating a `/.eggs` directory for --editable installations and we do not want to lint that code.

tests/test_cli.py
```diff
@@ -80,7 +80,7 @@ def test_check_returncode_nonzero(self):
     def test_can_eval(self):
         """Assert ``__repr__()`` can be parsed by ``eval()``."""
         string = repr(cli.CompletedProcess(**self.kwargs))
-        from pulp_smash.cli import CompletedProcess  # pylint:disable=unused-import
+        from pulp_smash.cli import CompletedProcess  # pylint:disable=unused-import,unused-variable
```
rochacbruno: pylint was requiring this category to be disabled on Python 3.7

tests/test_pulp_smash_cli.py
```diff
@@ -45,6 +45,8 @@ def test_missing_settings_file(self):
 class SettingsCreateTestCase(BasePulpSmashCliTestCase):
     """Test ``pulp_smash.pulp_smash_cli.settings_create`` command."""
 
+    maxDiff = None
```
rochacbruno: We want to see the full diff for Assertion errors

Makefile
kersommoura: What about we created a dedicated PR for this fix, since it is a different concern?

pulp_smash/config.py
```diff
@@ -91,6 +99,16 @@ def _get_pulp_3_api_role():
 }
 """The set of roles that must be present in a functional Pulp 3 application."""
 
+P3_OPTIONAL_ROLES = {
+    'pulp cli',
+    'squid',
```
kersommoura: Pulp 3 will not use squid. Lazy implementation is different than Pulp 2.

pulp_smash/config.py
```diff
@@ -91,6 +99,16 @@ def _get_pulp_3_api_role():
 }
 """The set of roles that must be present in a functional Pulp 3 application."""
 
+P3_OPTIONAL_ROLES = {
+    'pulp cli',
```
kersommoura: There is no CLI at the moment as well.

PR #1171, opened 2019-02-19
Adds `async_request` support to HTTP client.

rochacbruno:
Using asyncio to enable methods like `client.async_post`

Drops Python 3.4 as it does not support async/await

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: python
 python:
-- 3.4
 - 3.5
 - 3.6
 # Python 3.7+ is only available on Xenial, due to dependency requirements. For
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -7,8 +7,12 @@ handling of href paths and HTTP 202 status codes. This module provides a
 customizable client that makes it easier to work with the API in a safe and
 concise manner.
 """
+import asyncio
+import concurrent.futures
 import warnings
+from functools import partialmethod
 from time import sleep
+
 from urllib.parse import urljoin, urlparse
 
 import requests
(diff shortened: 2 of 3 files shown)


rochacbruno:
There was an attempt to use grequest to make things more simplerm but grequests (gevent) cannot work together with plumbum so dropped in favor of asyncIO

```py
class Client:
    ...
    
    def async_request(self, method, args_list, **kwargs):
        """Perform async requests using grequests.

        :param method: The HTTP method ex: GET, POST, DELETE
        :param args_list: A list of args to be passed to grequests.request
            ex: {'url': 'http://foo', 'json': {}, ...}
        :param kwargs: Common kwargs to be passed to all requests.
        """
        future_requests = []
        for arg in args_list:

            if 'url' not in arg:
                raise RuntimeError('url must be defined for each request')

            request_kwargs = self.request_kwargs.copy()
            request_kwargs['url'] = urljoin(
                request_kwargs['url'], arg.pop('url')
            )
            request_kwargs.update(arg)
            request_kwargs.update(kwargs)
            future_requests.append(
                grequests.request(method, **request_kwargs)
            )

        return [
            self.response_handler(self, response)
            for response in grequests.map(future_requests)
        ]

    async_delete = partialmethod(async_request, 'DELETE')
    async_get = partialmethod(async_request, 'GET')
    async_head = partialmethod(async_request, 'HEAD')
    async_options = partialmethod(async_request, 'OPTIONS')
    async_patch = partialmethod(async_request, 'PATCH')
    async_post = partialmethod(async_request, 'POST')
    async_put = partialmethod(async_request, 'PUT')
```

rochacbruno:
If we add external library `paco` we can keep Python 3.4

https://github.com/h2non/paco#examples

PR #1174, opened 2019-02-27
Updates pulp-smash for repo layout changes

bmbouter:
Updates pulp-smash for repo layout changes

This is designed to go along with these changes in Pulp 3.0

~https://github.com/pulp/pulp/pull/3896~
https://github.com/pulp/pulpcore/pull/2

The response format changed you we have to adjust how the utils parse
the response data for tests.

https://pulp.plan.io/issues/4283
re #4283

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -159,16 +159,16 @@ def _build_content_fetcher(content_field):
         repo_version = client.get(version_href)
 
         content = defaultdict(list)
-        for content_type, content_url in repo_version[content_field].items():
-            typed_content = client.get(content_url)
+        for content_type, content_dict in repo_version['content_summary'][content_field].items():
+            typed_content = client.get(content_dict['href'])
             content[content_type] = typed_content
         return content
     return inner
 
 
-get_content = _build_content_fetcher('content_hrefs')  # pylint:disable=invalid-name
-get_added_content = _build_content_fetcher('content_added_hrefs')  # pylint:disable=invalid-name
-get_removed_content = _build_content_fetcher('content_removed_hrefs')  # pylint:disable=invalid-name
+get_content = _build_content_fetcher('present')  # pylint:disable=invalid-name
+get_added_content = _build_content_fetcher('added')  # pylint:disable=invalid-name
+get_removed_content = _build_content_fetcher('removed')  # pylint:disable=invalid-name
 
 
 def _build_summary_fetcher(summary_field):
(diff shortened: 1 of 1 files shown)


pulp_smash/pulp3/utils.py
```diff
@@ -159,16 +159,16 @@ def inner(repo, version_href=None):
...
+        for content_type, content_dict in repo_version['content_summary'][content_field].items():
+            typed_content = client.get(content_dict['href'])
             content[content_type] = typed_content
         return content
     return inner
 
 
-get_content = _build_content_fetcher('content_hrefs')  # pylint:disable=invalid-name
-get_added_content = _build_content_fetcher('content_added_hrefs')  # pylint:disable=invalid-name
-get_removed_content = _build_content_fetcher('content_removed_hrefs')  # pylint:disable=invalid-name
```
dralley: You will want to change lines 174-205, also.  We should probably re-think the API to better suit the new layout, but for the time being let's just keep backwards compatibility.

bmbouter: fixed, ty

pulp_smash/pulp3/utils.py
```diff
@@ -196,13 +196,16 @@ def inner(repo, version_href=None):
             return {}
 
         client = api.Client(config.get_config(), api.page_handler)
-        return client.get(version_href)[summary_field]
+        to_return = client.get(version_href)['content_summary'][summary_field]
+        for key in to_return:
+            to_return[key].pop('href')  # Figure out how to assert on changing hrefs in response
```
rochacbruno: @bmbouter I don't get why we need to remove the href from response here, can you elaborate a little more why this is being popped out?

bmbouter: Well it really shouldn't be but it's the best I can offer before I go away for 2 weeks.

So the response formats look like this:  
screenshot from 2019-02-25 17-54-43

The `href` field changes depending on how many test have run before it so it can't easily be added to the fixture data. What do you think is the right way to resolve this issue in pulp smash?

dralley: Previously Pulp had a bunch of different fields on a repository named

* content_summary
* content_added_summary
* content_removed_summary
* content_hrefs
* content_added_hrefs
* content_removed_hrefs

These are all being merged together to look like this

```
"content_summary: {
    'added': {
        'file.file': {
            'count': 5,
            'href': '/pulp/api/v3/content/file/files/?repository_version_added=/pulp/api/v3/repositories/1/versions/1/'
        }
    },
    'present': {
        'file.file': {
            'count': 5,
            'href': '/pulp/api/v3/content/file/files/?repository_version=/pulp/api/v3/repositories/1/versions/1/'
        }
    },
    'removed': {}
}
```

Since the pulp_smash API is going to stay the same (for now at least), it's a hack to make it look the same as it did before.

bmbouter: Recapping our convo, I'm going to update this function to fully preserve the previous data layout and file a pulp-smash issue to handle creating a new util function with the new interface so users can port over.

bmbouter: That issue is here:  https://github.com/PulpQE/pulp-smash/pull/1175

PR #1179, opened 2019-03-06
Update get_artifact_paths() to be more correct

dralley:
It needs to be able to handle content types with no artifact associated.


kersommoura:
Can we add a short commit message, and a related issue, if any?

dralley:
@kersommoura  No related issue, but I added a message.

Issue #1178, opened 2019-03-06
Create new create_distribution method


CodeHeeler:
This PR for pulp3 changed distribution creation, updates, and deletion to be asynchronous: https://github.com/pulp/pulpcore/pull/6

The result is that these actions return a task now instead of a distribution object.  Anywhere tests performed these actions needed additional code added that was largely the same as you see here:
https://github.com/pulp/pulpcore/pull/6/files#diff-5400f9161b5ae065143722f44303c5e1R97

This work could be deduplicated by writing a new create_distribution method in pulp-smash that handles the returned task object and then yields the actual distribution for consumption by the tests.

rochacbruno:
I would give it an async'ish name

 ```python3
def await_distribution(task: TaskObject) -> dict:
```

rochacbruno:
https://github.com/pulp/pulpcore/pull/6#discussion_r263019901

rochacbruno:
This are the existing functions we can use to create that generic function 

https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L485-L551

PR #1181, opened 2019-03-14
Add util function create_distribution for Pulp 3

kersommoura:
Creating a distribution was modified to be an async task, returning a
202, and task reference. Add a function to ease the creation of
distributions.

See: https://github.com/pulp/pulpcore/pull/6

ref: https://github.com/PulpQE/pulp-smash/issues/1178

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -8,7 +8,11 @@ from urllib.parse import urljoin, urlsplit
 from packaging.version import Version
 
 from pulp_smash import api, config, utils
-from pulp_smash.pulp3.constants import ORPHANS_PATH, STATUS_PATH
+from pulp_smash.pulp3.constants import (
+    DISTRIBUTION_PATH,
+    ORPHANS_PATH,
+    STATUS_PATH,
+)
 
 
 def require_pulp_3(exc):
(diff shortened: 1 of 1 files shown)


kersommoura:
@rochacbruno, I think a long term solution will be to change our task system to use async code as you suggested on #1178. However this will require a massive overhaul, for the time being I decided to use what we already have in place. 

We already have  a function called `gen_distribution`, I decided to name the new one `create_distribution`. If you have any better naming idea. This change was already tested locally. I will send a PR to pulpcore requesting this one to test once more.

rochacbruno:
@kersommoura 

From the Pulp-Smash user's point of view, I believe we are introducing a source of confusion.

I explain:

To create a repository we do:

`api.post(REPO_PATH, body)`

And the above pattern follows for all other entities.

It is natural that when using Pulp-Smash to automate tasks with speed, we assume that to create distribution we must do:

`api.post(DISTRO_PATH, body)`

But specifically for distribution we have something that goes out of the way.

`pulp3.utils.create_distribution(cfg, body)`

How will a Pulp-Smash user know that only distribution has this difference?

I suggest that we follow a more Framework'ish thinking and move on to O.O to create entities in Pulp-Smash.

Example:

```python
from pulp_smash.entities import Repo, Distribution, Publication

repo = Repo.create(_generate=True)
distribution = Distribution.create(**body)
publication = Publication.create(repo=repo, ...)
```

Then we can encapsulate the logic behind that `create` (and other CRUD) methods for each of those entities.

It is really  a long term solution that we can work on, but I am concerned on how to solve that problem right now with an easier solution.

I suggest we put a good information on some readme or docs about this change, or even add a `warning` when a `post` request is made directly with the DISTRO_PATH

rochacbruno:
I was thinking if we can `force` the use of this function whenever an `api.post` is made against a distribution, so we keep the same external API

rochacbruno:
Take this example https://github.com/PulpQE/pulp-smash/blob/c199dd183513aa85f2eb1291d0dfdf3d9c5d739e/pulp_smash/api.py#L440
current code
```py
def post(self, url, json=_SENTINEL, **kwargs):
        """Send an HTTP POST request."""
        if json is _SENTINEL:
            return self.request('POST', url, **kwargs)
    return self.request('POST', url, json=json, **kwargs)
```

New:

```py
def post(self, url, json=_SENTINEL, **kwargs):
        """Send an HTTP POST request."""
        if json is _SENTINEL:
            return self.request('POST', url, **kwargs)
    if url in LIST_OF_ASYNC_ENDPOINTS_INCLUDING_DISTRIBUTION:
        return handle_async_response(self.request('POST', url, json=json, **kwargs))
    else:
        return self.request('POST', url, json=json, **kwargs)
```

rochacbruno:
question:

If it returns a `202` why is not being catched by this https://github.com/PulpQE/pulp-smash/blob/c199dd183513aa85f2eb1291d0dfdf3d9c5d739e/pulp_smash/api.py#L78

?

kersommoura:
@rochacbruno, good suggestions. That is why I opened this one as a draft PR.

kersommoura:
closed in favor of https://github.com/PulpQE/pulp-smash/pull/1184

PR #1183, opened 2019-03-18
[DO NOT MERGE] Add task_handler, ensureteardown and fixes some constants.

rochacbruno:
This PR is introduces 3 changes (instead of separete PRs) because certguard/functional test case needs to require this PR. 
https://github.com/pulp/pulp-certguard/pull/9

1. added dependency injection method to switch response handler

```py
client = api.Client(cfg, api.safe_handler)

# will get using `safe_handler`
client.get(url)

# Will get using different injected handler dependency
client.using_handler(api.json_handler).get(url)
```

2. Add a task handler to return `created`, `updated` or `done_task` resources

The `task_handler` will spawn the create task, pool task until it is `done`, if error will raise, if success will return the creatd distribution entity. (if it was a `PATCH|PUT` it would return the modified entity).

```py
distribution = client.using_handler(api.task_handler).post(DISTRIBUTION_PATH, gen_distribution())
```

3. ensureteardown

unittest will not execute `tearDownClass` if error occurs in a `@classmethod` use ExitStack to fix it

```py
from pulp_smash.utils import ensure_teardownclass

class TestCase(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        with ensure_teardownclass(cls):
            result = 1 / 0  # will raise ZeroDivisionError

    @classmethod
    def tearDownClass(cls):
        print("I will always be executed even if the error is on a classmethod")
        # this would never print without the `ensure_teardownclass`

```

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -7,6 +7,7 @@ handling of href paths and HTTP 202 status codes. This module provides a
 customizable client that makes it easier to work with the API in a safe and
 concise manner.
 """
+import copy
 import warnings
 from time import sleep
 from urllib.parse import urljoin, urlparse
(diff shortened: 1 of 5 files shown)


kersommoura:
I will review this PR with more time later on, but I have a few questions.

1 - Why can not we just create another client?

```python

safe_client = api.Client(cfg, api.safe_handler)

# will get using `safe_handler`
client.get(url)

```

We should break this PR in small ones. This PR is taking care of many different parts of the code. If we need to revert this PR later.  We will not be able to.

I understand the motivation for this one, but we will have to deal with those changes one by one. Like item 3 could be a separate PR.

rochacbruno:
> 1 - Why can we not just create another client?

Because we like cool, readable, performance cached, idiomatic stuff :)

If I already have a `Client` instance and I need a new one that differs only on the handler, I can just `clone` the existing and change specific things (in this case the handler)

I am ok to keep the `old verbose` approach and remove this feature if you like.

rochacbruno:
@kersommoura I can break this PR in 2 or 3 different PRs

This is going to break https://github.com/pulp/pulp-certguard/pull/9 which is using all the things introduced in this PR

rochacbruno:
closing in favor of

https://github.com/PulpQE/pulp-smash/pull/1184
https://github.com/PulpQE/pulp-smash/pull/1185
https://github.com/PulpQE/pulp-smash/pull/1186
https://github.com/PulpQE/pulp-smash/pull/1187

PR #1189, opened 2019-03-22
Overhaul CI and apply code format

kersommoura:
This PR does:

* Drop Python 3.4 support.
* Drop linters from setup.py
* Adopt black code sytle.
* Fix flake8 issues.
* Remove pylint as linter.
* Update MakeFile.
* Update Docs.


kersommoura:
This is the first big change, there are many other hooks to add to pre-commit. For instance: 

https://pre-commit.com/hooks.html

With time we can add those, and adjust the code as needed.

PR #1190, opened 2019-03-27
Lazy load the cli.Client machine

rochacbruno:
## Problem:

Everytime we instantiate a new instance of `cli.Client` its `__init__` method
code tries to make communication with the remote host to determine if it is
a local or ssh machine to be used.

The communication with `machine` should be done only when it is necessary,
which is when the `run` command is called.

### To reproduce the problem

```json
{
  "hosts": [
    {
      "hostname": "this-host-does-not-exists-or-server-is-down",
     ...
    }
  ]
}
```

```bash
$ export PULP_SMASH_CONFIG_FILE=/path/to/above/settings.json
$ python -c "from pulp_smash import cli, config; c = cli.Client(config.get_config()); print(c)"
plumbum.machines.session.SSHCommsError: No communication channel detected. Does the remote exist?
```

**We have not called the **run** method, so the communication with server should not be done yet.**

## Solution:

Transform `machine` in a lazy `@property` then it is instantiated only
when it is first called.

Not calling `run`
```bash
$ python -c "from pulp_smash import cli, config; c = cli.Client(config.get_config()); print(c)"
<pulp_smash.cli.Client object at 0x7f5439275ef0>
```

When `run` is called
```bash
$ python -c "from pulp_smash import cli, config; c = cli.Client(config.get_config()); c.run(['ls'])"
plumbum.machines.session.SSHCommsError: No communication channel detected. Does the remote exist?
```

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -193,26 +193,30 @@ class Client:  # pylint:disable=too-few-public-methods
                 pulp_host = cfg.get_hosts("pulp cli")[0]
             else:
                 pulp_host = cfg.get_hosts("shell")[0]
+
         self.pulp_host = pulp_host
-        hostname = pulp_host.hostname
-        transport = pulp_host.roles.get("shell", {}).get("transport")
-        if transport is None:
-            transport = "local" if hostname == socket.getfqdn() else "ssh"
-        if transport == "local":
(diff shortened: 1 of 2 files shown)


ragabala:
@rochacbruno thanks for the documentation :100:

PR #1194, opened 2019-04-02
Add publication()

ipanova:


pulp_smash/pulp3/utils.py
```diff
@@ -139,6 +139,27 @@ def publish(cfg, publisher, repo, version_href=None):
...
+    """
+
+    if version_href is None:
+        data = {"repository": repo["_href"]}
+    else:
+        data = {"repository_version": version_href}
+    client = api.Client(cfg, api.json_handler)
+    call_report = client.post(path, data)
+    tasks = tuple(api.poll_spawned_tasks(cfg, call_report))
+    return client.get(tasks[-1]["created_resources"][0])
```
rochacbruno: The `task_handler` already does this for generic `task based resource creation`
https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L205-L268

The only difference right now is that it is reading `tasks[0]` assuming it has only one, we can change to read `tasks[-1]` or make it a parameter.

pulp_smash/pulp3/utils.py
```diff
@@ -139,6 +139,27 @@ def publish(cfg, publisher, repo, version_href=None):
     return client.get(tasks[-1]["created_resources"][0])
 
 
+def publication(cfg, path, repo, version_href=None, **kwargs):
```
kersommoura: Besides @rochacbruno comments.

path is not listed in the docstrings.

**kwargs not used in the code.

kersommoura:
@ipanova , can we add a simple commit message? If there is a related issue even better. Just to be consistent with what we have been doing.

ipanova:
based on the comments i am closing the pr and going to use task_hander.

PR #1196, opened 2019-04-03
Add smart_handler as the default for api.Client

rochacbruno:
This handler can decide which handler to use based
on content response.

### DIFF
--- a/pulp_smash/api.py
+++ b/pulp_smash/api.py
@@ -265,6 +265,38 @@ def task_handler(client, response):
     return done_task
 
 
+def smart_handler(client, response):
+    """Decides which handler to call based on response content.
+
+    Do the following:
+
+    1. Pass response through safe_handler to handle 202 and raise_for_status.
+    2. Return the response if it is not Pulp 3.
+    3. Return the response if it is not application/json type.
+    4. Pass response through task_handler if is JSON 202 with 'task'.
+    5. Pass response through page_handler if is JSON but not 202 with 'task'.
+    """
+    # safe_handler Will raise_for_Status, handle 202 and pool tasks
+    response = safe_handler(client, response)
+
+    try:
+        check_pulp3_restriction(client)
+    except ValueError:
+        # If pulp is not 3+ return the result of safe_handler by default
+        return response
+
+    if response.headers.get("Content-Type") != "application/json":
+        # Not a valid JSON, return pure response
+        return response
+
+    # We got JSON is that a task call report?
+    if response.status_code == 202 and "task" in response.json():
+        return task_handler(client, response)
+
+    # Its JSON, it is not a Task, default to page_handler
+    return page_handler(client, response)
+
+
 class Client:
     """A convenience object for working with an API.
 
@@ -317,6 +349,7 @@ class Client:
     * :func:`pulp_smash.api.json_handler`
     * :func:`pulp_smash.api.page_handler`
     * :func:`pulp_smash.api.task_handler`
+    * :func:`pulp_smash.api.smart_handler`
 
     As mentioned, this class has configurable request and response handling
     mechanisms. We've covered response handling mechanisms — let's move on to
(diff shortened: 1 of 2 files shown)


kersommoura:
@rochacbruno I do like this idea. To have a generic response_handler for Pulp 3 that the user does not need to know in advance what kind of status code, or a pagination will be part of the response. We need to test this to assure that we are not missing any edge case. What are about we call it generic_handler or general_handler?

rochacbruno:
@kersommoura 

I suggest something like

- `base_handler`
- `default_handler`
- `dispatch_handler` (ugly name, but very meaningful)

kersommoura:
@rochacbruno naming anything it is so hard. @bherrin3 can decide.

kersommoura:
In my mind the name has to show the intent...and the intent it is to be a general reponse_handler for Pulp 3. One that the user will pick in the cases that user does not know what response will get back from Pulp 3. But I am fine with any choice.

rochacbruno:
@kersommoura if we want to show intent, then `smart_handler` is ok! 

Like other tech examples:
- Smart Phone
- Smart Templates

We want to express the idea that

## This handler can make its own choice based on context

I vote for `smart_` or `dispatch_`

rochacbruno:
Or `auto_handler` - `automatic_handler`

bherrin3:
hmm.. I think you may already have the right name with `smart_handler`. That has intelligent to decide which is the appropriate handler given the case. I don't think it infers anything other than it uses logic to decide the appropriate handler.

Else, a method name that names the action is doing. I am not sure why this is post-pended with `_handler`. However, of the names presented, `smart_handler` is the best I have seen.

Using `base_` or `default_` assumes there is a base class (subset) that is common to all handlers you are using, which is not the case here. The appropriate handler is being chosen. Maybe the function name shouldn't have `handler` in it since it is chosing and returning a handler and not actually a handler?

I don't have a `dog in this race`. I would just choose something and comment what it does. :)

rochacbruno:
Here's a good name:
`client = api.Client(reponse_handler=api.use_this_if_you_dont_know_how_to_choose_a_handler).get(...)`

:smile:

Why so serious?

lets call it `ninja_handler`, `super_handler`, `handler_with_lasers` or use some buzzword like `awesome_handler`

rochacbruno:
Lets stick with `smart_handler` ? So we can put some test cases and merge

kersommoura:
@rochacbruno, I am reviewing this one. Good job. Nice unittests.

I updated the [pagination test](https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/using_plugin/test_pagination.py) to use the `smart_handler`, and test failed. I am evaluating this one right now.

kersommoura:
@rochacbruno, it was an error on my side. I tested a few tests for Pulp 3 using the `smart_handler`, and all passed.

pulp_smash/api.py
```diff
@@ -265,6 +265,41 @@ def task_handler(client, response):
...
+    try:
+        check_pulp3_restriction(client)
+    except ValueError:
+        # If pulp is not 3+ return the result of safe_handler by default
+        return response
+
+    ctype = response.headers.get("Content-Type")
+    is_202 = response.status_code == 202
+
+    if ctype != "application/json":
```
kersommoura:
```suggestion
    if response.headers.get("Content-Type") != "application/json":
```

pulp_smash/api.py
```diff
@@ -265,6 +265,41 @@ def task_handler(client, response):
...
+        # Not a valid JSON, return pure response
+        return response
+
+    # We got JSON is that a task call report?
```
kersommoura: Move this `is_202` closer.
```suggestion
    # We got JSON is that a task call report?
    is_202 = response.status_code == 202
```

is_202 = response.status_code == 202

pulp_smash/api.py
```diff
@@ -265,6 +265,41 @@ def task_handler(client, response):
...
+    response = safe_handler(client, response)
+
+    try:
+        check_pulp3_restriction(client)
+    except ValueError:
+        # If pulp is not 3+ return the result of safe_handler by default
+        return response
+
+    ctype = response.headers.get("Content-Type")
+    is_202 = response.status_code == 202
```
kersommoura:
```suggestion

```

pulp_smash/api.py
```diff
@@ -265,6 +265,41 @@ def task_handler(client, response):
...
+    # safe_handler Will raise_for_Status, handle 202 and pool tasks
+    response = safe_handler(client, response)
+
+    try:
+        check_pulp3_restriction(client)
+    except ValueError:
+        # If pulp is not 3+ return the result of safe_handler by default
+        return response
+
+    ctype = response.headers.get("Content-Type")
```
kersommoura:
```suggestion
```

Issue #1173, opened 2019-02-26
Pulp-smash polling timeout is too high for Travis


daviddavis:
We're running pulp-smash tests in Travis. It looks like when pulp-smash polls, it will wait up to 30 min:

https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L526

The problem is that Travis times out after 10 min. We should have pulp-smash timeout after about 5-8 min. This would provide us with better information as we would know a task is taking too long vs simply having an error appear in Travis.

Perhaps this poll number could be a config variable?

kersommoura:
@daviddavis, this time was chosen to solve a Pulp 2 issue. Let`s evaluate to have different time out for Pulp 3.

daviddavis:
@kersom, would it be possible to have a configuration setting? You could even default it to the current value.

daviddavis:
Also, I want to point out [that the issue of having a hard-coded timeout has been problematic for us before](https://github.com/PulpQE/Pulp-2-Tests/issues/6).

kersommoura:
@daviddavis, I think so. We will evaluate this, and we are currently evaluating our whole process to define the settings file.

PR #1203, opened 2019-04-25
Add optional timeout to settings.json 

kersommoura:
Add optional timeout variable as part of settings.json. This will allow one to
specify the timeout for the task system in seconds from now, if not
defined it will be set to default value of 1800s. This was the previous
used value.

Tweak the task system timeout to use seconds, and query Pulp more often.
Reduce the sleep time on `poll_task` function to 0.3s.

Adjust unittests, and remove old tests.

closes:#1173

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -104,6 +104,7 @@ JSON_CONFIG_SCHEMA = {
                     "items": {"$ref": "#/definitions/pulp 2 host"},
                 },
                 "pulp": {"$ref": "#/definitions/pulp"},
+                "general": {"$ref": "#/definitions/general"},
             },
         },
         {
@@ -117,6 +118,7 @@ JSON_CONFIG_SCHEMA = {
                     "items": {"$ref": "#/definitions/pulp 3 host"},
                 },
                 "pulp": {"$ref": "#/definitions/pulp"},
+                "general": {"$ref": "#/definitions/general"},
             },
         },
     ],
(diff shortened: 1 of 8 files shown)


pulp_smash/config.py
```diff
@@ -95,7 +95,7 @@ def _get_pulp_3_content_role():
     "anyOf": [
         {
             "additionalProperties": False,
-            "required": ["pulp", "hosts"],
+            "required": ["pulp", "general", "hosts"],
```
rochacbruno: I think this should not be required, or it will break compatibility with existing settings.

The general may be optional, if not preovided then it will use the default timeout config.

pulp_smash/config.py
```diff
@@ -104,11 +104,12 @@ def _get_pulp_3_content_role():
                     "items": {"$ref": "#/definitions/pulp 2 host"},
                 },
                 "pulp": {"$ref": "#/definitions/pulp"},
+                "general": {"$ref": "#/definitions/general"},
             },
         },
         {
             "additionalProperties": False,
-            "required": ["pulp", "hosts"],
+            "required": ["pulp", "general", "hosts"],
```
rochacbruno: same here, should not be required

pulp_smash/config.py
```diff
@@ -131,6 +133,14 @@ def _get_pulp_3_content_role():
                 "selinux enabled": {"type": "boolean"},
             },
         },
+        "general": {
+            "additionalProperties": False,
+            "required": ["timeout"],
```
rochacbruno: I think timeout is optional, if user does not want to provide it, then pulp-smash uses the default value.

pulp_smash/config.py
```diff
@@ -545,11 +556,19 @@ def load(cls, xdg_subdir=None, config_file=None):
...
             )
             loaded_config["hosts"] = loaded_config.pop("systems")
+
+        try:
+            timeout = loaded_config["general"]["timeout"]
+        except KeyError:
+            # For old versions of settings.json without the value of timeout
+            # defined.
+            timeout = 1800
+
```
rochacbruno: all the above try can be

```python
timeout = loaded_config.get("general", {}).get("timeout", 1800)
```
try to `get` a `general` dict from config, if doesn't exist return empty dict, then try to `get` the `timeout` if not defined use default `1800`

kersommoura: @rochacbruno, I will update properly.

PR #1207, opened 2019-05-14
Temporarily update task polling system for Pulp 2

kersommoura:
Temporarily update task polling system to address Pulp 2 issue.
Pulp-smash was polling at a 0.3s rate that is occasionally returning
a task "COMPLETE" but the values for the task had yet to be reset.

Adjust the task system to use the previous value of 2s for Pulp 2.

See: https://pulp.plan.io/issues/4795


kersommoura:
Tested against Pulp 2 and Pulp 3.

PR #1208, opened 2019-06-06
Update for new Pulp 3 default ports

mikedep333:
Over the course of the Pulp 3 development cycle, API and Content are now separate servers by default, and the default ports have changed.

Conditional logic was written to keep Pulp 2 config the same.


kersommoura:
@mikedep333, thanks for your PR. Can we squash those commits in just one? We follow this standard.

mikedep333:
@kersommoura squashed, thank you!

kersommoura:
We are currently testing Pulp 3 using the installation provided by the ansible-installer with nginx. Then the http interactions are happening using the port 80. 

With those changes the settings.json created will be like...

```json
{
  "general": {
    "timeout": 1800
  },
  "hosts": [
    {
      "hostname": "192.168.122.20",
      "roles": {
        "api": {
          "port": 24817,
          "scheme": "http",
          "service": "nginx",
          "verify": false
        },
        "content": {
          "port": 24816,
          "scheme": "http",
          "service": "pulp_content_app",
          "verify": false
        },
        "pulp resource manager": {},
        "pulp workers": {},
        "redis": {},
        "shell": {
          "transport": "ssh"
        }
      }
    }
  ],
  "pulp": {
    "auth": [
      "admin",
      "admin"
    ],
    "selinux enabled": false,
    "version": "3.0"
  }
}

```

Not sure if this what the defaults that we need. @rochacbruno, what do you think?

It seems that for the dev env the port 24817 is the right one, but once the ngnix is added...the 80 for http is the right one.

mikedep333:
With Pulp 3 the choice is now between gunicorn alone, or gunicorn with nginx, correct?

If so, I am in favor of updating the prompts and all for Pulp 3, as well as the default ports depending upon webserver.

rochacbruno:
@mikedep333 I like your idea, however pulp-smash is a `functional testing tool` and that means it is meant to use on `black boxed` tests. Ideally pulp-smash would not know which web-server is running behind the scenes and would just know which ports and addresses can use to connect to Pulp.

Right now this config file has too many information like which queue engine, which webserver technology etc.. that is there only to be enable to `start-stop` services and only because of that, if we could we would like to have a config file with only the address and ports, just as the **users** would know.

I vote to keep the `80` as the default for API and then the change is very easy when it is really needed.

kersommoura:
See: https://pulp.plan.io/issues/4965

PR #1209, opened 2019-06-11
Add util get_served_content_url for Pulp 3

kersommoura:
`get_served_content_url` will return the URL that content is served
given a distribution.
Adjust function `download_content_unit` to use the newly added function.


kersommoura:
@rochacbruno, that is the goal.

PR #1210, opened 2019-07-02
Changing download_policies

bmbouter:
With https://github.com/pulp/pulpcore/commit/59f3450286d0a3a88e4b9fd1edb3568ee65bf54f
the download policy by default only includes 'immediate'.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/constants.py
+++ b/pulp_smash/pulp3/constants.py
@@ -22,13 +22,9 @@ CONTENT_PATH = urljoin(BASE_PATH, "content/")
 
 DISTRIBUTION_PATH = urljoin(BASE_PATH, "distributions/")
 
-DOWNLOAD_POLICIES = ("immediate", "on_demand", "streamed")
+IMMEDIATE_DOWNLOAD_POLICIES = ("immediate",)
 
-JWT_PATH = urljoin(BASE_PATH, "jwt/")
-
-LAZY_DOWNLOAD_POLICIES = tuple(
-    [item for item in DOWNLOAD_POLICIES if item != "immediate"]
-)
+ON_DEMAND_DOWNLOAD_POLICIES = ("on_demand", "streamed")
 
 MEDIA_PATH = "/var/lib/pulp"
 


pulp_smash/pulp3/constants.py
```diff
@@ -22,13 +22,9 @@
 
 DISTRIBUTION_PATH = urljoin(BASE_PATH, "distributions/")
 
-DOWNLOAD_POLICIES = ("immediate", "on_demand", "streamed")
+DOWNLOAD_POLICIES = ("immediate", )
 
-JWT_PATH = urljoin(BASE_PATH, "jwt/")
```
asmacdo: :1st_place_medal:

kersommoura: We kept this one around since  JWT could be added back based on the discussions at the time.

pulp_smash/pulp3/constants.py
```diff
@@ -22,13 +22,9 @@
 
 DISTRIBUTION_PATH = urljoin(BASE_PATH, "distributions/")
 
-DOWNLOAD_POLICIES = ("immediate", "on_demand", "streamed")
+DOWNLOAD_POLICIES = ("immediate", )
```
kersommoura: This name can be misleading. Since it is plural. I would suggest:
```suggestion
IMMEDIATE_DOWNLOAD_POLICY = ("immediate", )
```

bmbouter: I pushed this fix, but what about just making it

`IMMEDIATE_POLICY = "immediate"` since it's single now.

pulp_smash/pulp3/constants.py
bmbouter: Since then though we removed the JWT entirely so I think we should remove this too.

bmbouter:
I pushed PRs to pulpcore and pulp_file with fixes:
https://github.com/pulp/pulpcore/pull/199
https://github.com/pulp/pulp_file/pull/255

pulp_smash/pulp3/constants.py
```diff
@@ -2,9 +2,9 @@
 """Constants for Pulp 3 tests."""
 from urllib.parse import urljoin
 
-from pulp_smash.api import (  # noqa: F401
+from pulp_smash.api import (
```
kersommoura:
```suggestion
from pulp_smash.api import (  # noqa: F401
    _P3_TASK_END_STATES as P3_TASK_END_STATES,
)
```

PR #1215, opened 2019-08-12
Problem: can't run CLI related tests on kubernetes

dkliban:
Solution: add ability to specify 'kubectl' as shell transport

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -210,6 +210,7 @@ class Client:  # pylint:disable=too-few-public-methods
 
         self._is_root_cache = None
         self._machine = None
+        self._podname = None
         logger.debug("New %s", self)
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


pulp_smash/cli.py
```diff
@@ -232,6 +233,11 @@ def machine(self):
 
             if transport == "local":
                 self._machine = plumbum.machines.local
+            elif transport == "kubectl":
+                self._machine = plumbum.machines.local
+                chain = self._machine["kubectl"]["get", "pods"] | \
+                        self._machine["grep"]["-E", "pulp-api-(\w+)-(\w+)"]
+                self._podname = chain().split(" ")[0]
```
mikedep333: Are you trying to get only the name of the pod? If so, just add "-o" to grep.

pulp_smash/cli.py
```diff
@@ -276,6 +282,9 @@ def run(self, args, sudo=False, **kwargs):
         kwargs.setdefault("retcode")
         logger.debug("Running %s cmd (sudo:%s) - %s", args, sudo, kwargs)
 
+        if self._podname:
+            args = ("kubectl", "exec", self._podname) + tuple(args)
```
mikedep333: We need -- (dash dash) at the end also.

mikedep333: We need -- (dash dash) at the end also.

So that we can run commands with arguments.

pulp_smash/cli.py
dkliban: yep, i added teh -o. thank you!

pulp_smash/cli.py
```diff
@@ -276,6 +282,9 @@ def run(self, args, sudo=False, **kwargs):
         kwargs.setdefault("retcode")
         logger.debug("Running %s cmd (sudo:%s) - %s", args, sudo, kwargs)
 
+        if self._podname:
+            args = ("kubectl", "exec", self._podname) + tuple(args) + ("--",)
```
mikedep333: The -- needs to become before the command you actually run.

PR #1217, opened 2019-08-20
[WIP] kubectl: Always use sudo in front of it rather than conditionally

mikedep333:
k3s's embedded kubectl requires this. And that's what we use on Travis CI.


kersommoura:
@mikedep333, do we have any travis PR/branch with tests running using this change?

kersommoura:
I just saw the pulpcore PRs.

PR #1223, opened 2019-10-28
Modifications for typed repositories

dralley:
re: #5625
https://pulp.plan.io/issues/5625

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -81,9 +81,46 @@ def sync(cfg, remote, repo, **kwargs):
         created sync.
     """
     client = api.Client(cfg)
-    data = {"repository": repo["pulp_href"]}
+    data = {"remote": remote["pulp_href"]}
     data.update(kwargs)
-    return client.post(urljoin(remote["pulp_href"], "sync/"), data)
+    return client.post(urljoin(repo["pulp_href"], "sync/"), data)
+
+
+def modify_repo(
+    cfg, repo, base_version=None, add_units=None, remove_units=None
+):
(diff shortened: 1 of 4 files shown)


dralley:
Work-in-progresss, don't merge

kersommoura:
@dralley, we can send a draft PR when you do want to merge yet.

https://github.blog/2019-02-14-introducing-draft-pull-requests/

pulp_smash/pulp3/utils.py
```diff
@@ -81,9 +81,44 @@ def sync(cfg, remote, repo, **kwargs):
...
     """
     client = api.Client(cfg)
-    data = {"repository": repo["pulp_href"]}
+    data = {"remote": remote["pulp_href"]}
     data.update(kwargs)
-    return client.post(urljoin(remote["pulp_href"], "sync/"), data)
+    return client.post(urljoin(repo["pulp_href"], "sync/"), data)
+
+
+def modify_repo(cfg, repo, base_version=None, add_units=None, remove_units=None):
```
kersommoura: If `add_units is None`, and `remove_units is None,` but `base_version is not None,` it will attempt to modify a repository without units to be added or removed, is it an expected behave?  How pulp will handle this?

Should add logic to restrict this?

dralley: I initially had that logic there, but I discovered that a bunch of tests do actually do exactly this (make a new version without changing anything at all).

kersommoura: @dralley, Interesting.  I was expecting Pulp to raise an exception in case just `base_version` is provided.

PR #1226, opened 2019-10-30
Add section about release process documentation

kersommoura:
Add section about release process in the documentation.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -1,0 +1,77 @@
+Release Process
+===============
+
+Location: :doc:`/index` → :doc:`/release`
+
+The two goals of the release process are to create a tagged commit which
+bumps the version file and describes changes since the last release, and to
+generate packages which can be uploaded to PyPI.
+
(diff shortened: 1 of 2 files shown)


docs/release.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    cat VERSION 
+
+4. From the root of Pulp Smash dir run the script to create a new release.
+
+.. code-block:: sh
+
+    git status
+    ./scripts/release.sh 'string_new_version'
+
```
rochacbruno: @kersommoura worth mentioning that you should not have a virtualenv activated when running this script?

kersommoura: Let me verify this again.

kersommoura: There was no problem running the script with a python virtual env activated. 

https://github.com/PulpQE/pulp-smash/blob/1ee2d133a1080254a85fe0cf89b898fd8b9de897/scripts/release.sh#L75

PR #1228, opened 2019-12-18
Option for local fixture

fabricio-aguiar:
https://pulp.plan.io/issues/5873
ref #5873

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -19,6 +19,7 @@ from packaging.version import Version
 from xdg import BaseDirectory
 
 from pulp_smash import exceptions
+from pulp_smash.constants import PULP_FIXTURES_BASE_URL
 
 
 def _get_pulp_2_api_role():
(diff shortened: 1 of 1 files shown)


pulp_smash/config.py
```diff
@@ -489,6 +490,13 @@ def get_base_url(self, pulp_host=None, role="api"):
             pass
         return urlunsplit((scheme, netloc, "", "", ""))
 
+    def get_fixtures_url(self):
+        """Return fixtures URL."""
+        local_fixture = self.custom.get("local_fixture")
+        if local_fixture:
+            return urljoin(self.get_base_url(role="api"), "/static/fixtures/")
```
daviddavis: Can you have this default to `https://repos.fedorapeople.org/pulp/pulp/fixtures/`?

daviddavis: Is it possible for local_fixture to be the url address? Something like:

```python
def get_fixtures_url(self):
   local_fixture = self.custom.get("local_fixture")
   if local_fixture:
       return local_fixture
   else:
       return PULP_FIXTURES_BASE_URL
```

PR #1229, opened 2020-01-02
Use requests directly to download content

mdellweg:
This stops the pulp-smash client to warn about the wrong hostname and
prevents sending the admin credentials to the content server.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -123,7 +123,7 @@ def modify_repo(
     return client.post(urljoin(repo["pulp_href"], "modify/"), params)
 
 
-def download_content_unit(cfg, distribution, unit_path, **kwargs):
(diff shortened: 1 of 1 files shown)


mdellweg:
Alternatively, another config object without `auth` and the content server could be used.
This was the easy way to get rid of warnings like:
```
RuntimeWarning: This client should be used to communicate with localhost, but a request is being made to pulp3-source-debian10.example.com. The request will be made, but beware that information intended for localhost (such as authentication tokens) may now be sent to pulp3-source-debian10.example.com. Here's the list of options being sent with this request: {'verify': False, 'auth': ('admin', 'password'), 'url': <...>}
```

mdellweg:
Just realized, that there is a much better wrapper already in smash: `utils.get_http`

mdellweg:
@rochacbruno ping

PR #1231, opened 2020-02-28
Handling setuptools error

fabricio-aguiar:
Reference: https://github.com/pypa/setuptools/issues/1694#issuecomment-466010982
[noissue]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,9 @@
+[build-system]
+requires = [
+    "setuptools >= 40.8.0",
+    "wheel >= 0.29.0",
+]
+build-backend = 'setuptools.build_meta'
+
 [tool.black]
 line-length = 79


pyproject.toml
```diff
@@ -1,2 +1,10 @@
+[build-system]
+requires = [
+    "setuptools >= 40.8.0",
+    "setuptools_scm >= 2.0.0, <4",
```
rochacbruno: Why do we need `setuptools_scm` here? this project takes its version from https://github.com/pulp/pulp-smash/blob/master/VERSION instead of from scm tags.

fabricio-aguiar: actually I just copied the example solution from the comment: https://github.com/tox-dev/tox/blob/master/pyproject.toml#L1
I'll remove that

PR #1234, opened 2020-03-23
Generate bindings config

fabricio-aguiar:
https://pulp.plan.io/issues/6339
ref #6339

### DIFF
--- a/pulp_smash/config.py
+++ b/pulp_smash/config.py
@@ -18,6 +18,8 @@ import jsonschema
 from packaging.version import Version
 from xdg import BaseDirectory
 
+from pulpcore.client.pulpcore import Configuration
+
 from pulp_smash import exceptions
 from pulp_smash.constants import PULP_FIXTURES_BASE_URL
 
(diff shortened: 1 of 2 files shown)


pulp_smash/config.py
```diff
@@ -512,6 +512,14 @@ def get_content_host_base_url(self):
             role="content" in pulp_host.roles and "content" or "api",
         )
 
+    def get_bindings_config(self):
```
bmbouter: This looks great! Should it also set `configuration.safe_chars_for_path_param = "/"`? Is that required for 100% of clients?

dkliban: it should do that for sure.

fabricio-aguiar: I also thought about it, but `safe_chars_for_path_param` is not set on `__init__`:
```python
class Configuration(object):
    """NOTE: This class is auto generated by OpenAPI Generator

    Ref: https://openapi-generator.tech
    Do not edit the class manually.

    :param host: Base url
    :param api_key: Dict to store API key(s)
    :param api_key_prefix: Dict to store API prefix (e.g. Bearer)
    :param username: Username for HTTP basic authentication
    :param password: Password for HTTP basic authentication
    """

    def __init__(self, host="http://localhost:24817",
                 api_key=None, api_key_prefix=None,
                 username="", password=""):
```

bmbouter: I think setting the attribute afterwards (not with `__init__`) is correct because that is how we do it already [here for example](https://github.com/pulp/pulp_ansible/blob/ba4b021f48b1f64a982bd8bb997b7dd31df84ded/pulp_ansible/tests/functional/utils.py#L37).

@dkliban what do you think?

PR #1236, opened 2020-04-15
Add docker transport to Cli

mdellweg:


mikedep333:
You know, I was thinking this during the podman PR: They purposely have the same syntax.

mdellweg:
All i know for sure is, that assigning the tty (`-t`) is not working. Probably podman is less picky there.

PR #1241, opened 2020-06-15
Moving bindings utils from plugins to pulp-smash

fao89:
Addressing: https://github.com/pulp/pulp_rpm/pull/1722#pullrequestreview-422964847

### DIFF
--- a/pulp_smash/pulp3/bindings.py
+++ b/pulp_smash/pulp3/bindings.py
@@ -1,0 +1,67 @@
+from unittest import TestCase
+from time import sleep
+
+from pulpcore.client.pulpcore import ApiClient, TasksApi
+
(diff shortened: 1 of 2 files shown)


pulp_smash/pulp3/bindings.py
```diff
@@ -0,0 +1,57 @@
...
+    completed = ["completed", "failed", "canceled"]
+    task = tasks.read(task_href)
+    while task.state not in completed:
+        sleep(SLEEP_TIME)
+        task = tasks.read(task_href)
+
+    if task.state == "completed":
+        return task.created_resources
+
+    return task.to_dict()
```
mdellweg, line 38: I would prefer if failed task threw an exception. It would make for easier tests and the function would have a single return type.

PR #1243, opened 2020-10-15
Adds utils.get_pulp_setting() for retrieving pulp 3 settings

dkliban:
[noissue]

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -149,3 +149,26 @@ def ensure_teardownclass(testcase):
         stack.callback(testcase.tearDownClass)
         yield stack
         stack.pop_all()
+
+
+def get_pulp_setting(cli_client, setting_name):
+    """Retrieves the value of a Pulp setting from system under test."""
+    if cli_client.transport in ["docker", "podman", "kubectl"]:
+        manager = "pulpcore-manager"
+    else:
+        ps_output = cli_client.run(("ps", "ax")).stdout.splitlines()
+        for line in ps_output:
+            if "pulpcore" in line:
+                if "bin/python" in line.split()[4]:
+                    bin_dir = line.split()[4].rsplit("/", maxsplit=1)[0]
+        manager = "{}/pulpcore-manager".format(bin_dir)
+    command = (
+        "from django.conf import settings; print(repr(getattr(settings,"
+        "'{}', None)))".format(setting_name)
+    )
+    value = (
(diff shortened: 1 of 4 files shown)


pulp_smash/utils.py
```diff
@@ -149,3 +149,33 @@ def setUpClass(cls):
         stack.callback(testcase.tearDownClass)
         yield stack
         stack.pop_all()
+
+
+def get_pulp_setting(cli_client, setting_name, python_type="str"):
```
mdellweg, line 154: What about calling something like this?
`pulpcore-manager shell -c "from django.conf import settings; print(repr(settings.INSTALLED_PULP_PLUGINS))"`

dkliban: That's a good idea. I'll change it to `pulpcore-manager shell -c "from django.conf import settings; print(repr(getattr(settings, setting_name, None)))"`

dkliban: I am trying this on an older container (3.6.2) and the output has some extra text in it. 

```
pulp: pulpcore.app.serializers.base:WARNING: Please provide either 'view_name' or 'view_name_pattern' for DetailRelatedField on _call_with_frames_removed.
'pulpcore.app.models.storage.FileSystem'
```

mdellweg: Hmm, the warning should be sent to stderr, right? You should be able to filter that.

pulp_smash/utils.py
```diff
@@ -149,3 +149,25 @@ def setUpClass(cls):
...
+        manager = "{}/pulpcore-manager".format(bin_dir)
+    command = "from django.conf import settings; print(repr(getattr(settings, '{}', None)))".format(
+        setting_name
+    )
+    value = (
+        cli_client.run((manager, "shell", "-c", command))
+        .stdout.rstrip("\n")
+        .replace("'", "")
+    )
+    return value
```
mdellweg, line 173:
```suggestion
    return python_type(value)
```

mdellweg (approved):
This looks much better.

Is my comment getting the `python_type` thing correctly?

pulp_smash/utils.py
dkliban: I did not mean for python_type kwarg to remain. It was needed for the previous implementation with dynaconf.

mdellweg: So this will always return a `repr` string of the config option.
Maybe using json to actually get the proper type would be an option?

PR #1244, opened 2020-11-24
Use json to retrieve pulp settings

mdellweg:
[noissue]

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -6,6 +6,7 @@ but the reverse should not be done.
 """
 import contextlib
 import hashlib
+import json
 import uuid
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


pulp_smash/utils.py
```diff
@@ -163,12 +164,9 @@ def get_pulp_setting(cli_client, setting_name):
...
-        "'{}', None)))".format(setting_name)
-    )
-    value = (
-        cli_client.run((manager, "shell", "-c", command))
-        .stdout.rstrip("\n")
-        .replace("'", "")
+        "import json; from django.conf import settings;"
+        "print(json.dumps(getattr(settings,'{}', None)))".format(setting_name)
     )
+    json_value = cli_client.run((manager, "shell", "-c", command))
```
goosemania, line 170: Works like a charm, just read the stdout and that would be it.

`json_value = cli_client.run((manager, "shell", "-c", command)).stdout.rstrip("\n")`

goosemania (approved):

goosemania:
Thank you!

PR #1247, opened 2020-11-30
Change the output of the monitor_task utility

dralley:
There are a lot of use cases for needing parts of the task output that
are not created resources. It's easy enough to get the created
resources if you do want them that we should just return the raw task
output.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/bindings.py
+++ b/pulp_smash/pulp3/bindings.py
@@ -1,8 +1,7 @@
 from unittest import TestCase
 from time import sleep
 
-from pulpcore.client.pulpcore import ApiClient, TasksApi
-
+from pulpcore.client.pulpcore import ApiClient, OrphansApi, TaskGroupsApi, TasksApi, TaskGroupsApi
 from pulp_smash.api import _get_sleep_time
 from pulp_smash.config import get_config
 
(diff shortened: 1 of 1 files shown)


fao89 (approved):

fao89:
it probably will break some pulp_rpm tests

dralley:
It'll break several across multiple projects.  Not going to merge this immediately

pulp_smash/pulp3/bindings.py
```diff
@@ -59,7 +68,34 @@ def monitor_task(task_href):
         sleep(SLEEP_TIME)
         task = tasks.read(task_href)
 
-    if task.state == "completed":
-        return task.created_resources
+    if task.state != "completed":
+        raise PulpTaskError(task=task)
```
mdellweg, line 72: While you are at it, should we add a timeout?

dralley: I don't think we can set a default timeout since this function is used in tons of places in very different situations, but I'm not against adding the feature itself.  Do we have a specific use case?

dralley: If we come up with one, let's just add it separately.  Want to go ahead and get this merged so I can close out a dozen tabs :)

mdellweg: I was thinking of a hanging task. It would be nicer if the CI didn't exit with "Your time is up".

dralley: @mdellweg The problem is, it's hard to find a good generic value for how long to wait, so it'd need to be pretty high by default, or be off by default.  If it's off by default, we'd need to go through quite some effort to use the feature.  If it's on by default, then all long-running tasks (like the performance tests) need to explicitly disable it, which seems kind of annoying.

dralley: It's probably still useful I just don't want to worry about it right now.  It could require many changes across repos again.

PR #1251, opened 2021-01-19
Small fixups

dralley:


pulp_smash/utils.py
```diff
@@ -75,7 +75,7 @@ def get_sha256_checksum(url):
     return _CHECKSUM_CACHE[url]
 
 
-def http_get(url, **kwargs):
+def http_get(url, verify=False, **kwargs):
```
dralley, line 78: Since this method doesn't accept ```cfg``` as an argument, it doesn't use the smash config settings, so when it's used with Pulp URLs and doesn't like the SSL, problems happen.

pulp_smash/pulp3/bindings.py
```diff
@@ -36,7 +36,7 @@ class PulpTaskError(Exception):
 
     def __init__(self, task):
         """Provide task info to exception."""
-        description = task.error["description"]
+        description = task.to_dict()["error"]["description"]
```
mdellweg, line 39: I'm not opposed to this change, but is it needed?

pulp_smash/utils.py
mdellweg: Can we make this default to None and take the value from pulp-smash settings "content" role if unset?
This function is meant to be used in tests, and i doubt that any test can make an educated guess about the state of SSL.

pulp_smash/pulp3/bindings.py
dralley: It *looks* like it shouldn't be necessary, but for some reason it was.  I actually made this change back in December in response to some weird corner case where it wasn't working as expected, but I've forgotten what it was exactly.

I'm not opposed to leaving it out until we re-discover the reason though.

pulp_smash/utils.py
dralley: I'd feel a little weird about doing that, because the content role is specific to Pulp's configuration, but ```http_get()``` is primarily used on non-Pulp resources, like downloading content from external repos to compare the file checksum against Pulp.  That's probably why it never used the ```cfg``` to begin with.

pulp_smash/pulp3/bindings.py
mdellweg: both would be fine with me.

dralley: Let's just merge as-is then.

mdellweg (approved):

PR #1253, opened 2021-04-19
Fix monitor_task

mdellweg:
Do not depend on description to be available.

[noissue]


dralley:
@mdellweg What's the status of this PR, is it still needed?

mdellweg:
It's not needed anymore to make any current tests pass.

PR #1254, opened 2021-05-17
Add pulp-cli as dependency

fao89:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
         "pulpcore-client",
         "pyxdg",
         "requests",
-        "pulpcore-client",
+        "pulp-cli",
     ],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@
         "pulpcore-client",
         "pyxdg",
         "requests",
-        "pulpcore-client",
```
fao89, line 48: `pulpcore-client` was duplicated

setup.py
```diff
@@ -45,7 +45,7 @@
...
+        "pulp-cli",
```
mdellweg, line 48: I don't see, what this is trying to solve. Are any of the pulp-smash using tests depending on the cli?

fao89: it is forcing click 8.0 and breaking pulp-cli tests

PR #1255, opened 2021-05-18
Pin 'click' to <8.0.0 due to compatibility issues.

ggainey:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     ],
     packages=find_packages(include=["pulp_smash", "pulp_smash.*"]),
     install_requires=[
-        "click",
+        "click<8.0.0",
         "jsonschema",
         "packaging",
         "plumbum",


setup.py
```diff
@@ -38,7 +38,7 @@
     ],
     packages=find_packages(include=["pulp_smash", "pulp_smash.*"]),
     install_requires=[
-        "click",
+        "click<8.0.0",
```
bmbouter, line 41: The worry with this is that [pulpcore is using the newer version](https://github.com/pulp/pulpcore/blob/master/requirements.txt#L6) so this will cause it to receive a version incompatible when run there. I think the solution is to port pulp-smash to work with >8.0 and pin like pulpcore does.

bmbouter: Here's how RQ fixed it FWIW https://github.com/rq/rq/pull/1465/files

dralley: Yeah I actually thought this was the plan.  Which is my bad, I mis-read IRC.

ggainey: We're fixing this not only the wrong way, but in the wrong place - closing this PR :)

ggainey: Short-term, reopening to pin pulp-smash to 8.0.

dralley (approved):

PR #1258, opened 2021-07-08
Adds unit fetch that returns entire response obj

bmbouter:
In some cases we need not just the contnet, but the entire response
object.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/utils.py
+++ b/pulp_smash/pulp3/utils.py
@@ -4,6 +4,8 @@ from collections import defaultdict
 import warnings
 from urllib.parse import urljoin, urlsplit
 
+import requests
+
 from packaging.version import Version
 
 from pulp_smash import api, config, utils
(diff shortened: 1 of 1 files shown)


pulp_smash/pulp3/utils.py
```diff
@@ -131,6 +133,33 @@ def download_content_unit(_cfg, distribution, unit_path, **kwargs):
     return utils.http_get(unit_url, **kwargs)
 
 
+def download_content_unit_return_requests_response(_cfg, distribution, unit_path, **kwargs):
```
dralley, line 136: I feel like we should decompose this into two functions, one that produces a pulp url for a content unit, and one that makes the request.

Then the url logic can be shared between these two download functions, and it'll be easier to phase the other function out (since once the url logic is removed it's just a thin wrapper around `http_get`)

dralley (approved):

PR #1259, opened 2021-07-22
Add a utility for downloading directories on the host system using wget

dralley:
[noissue]


mdellweg (approved):
I like!

Crazy ideas:
- Can we provide a check function (or decorator) that raises a skip exception (or can be used in the skip decorator)?
- Would it be possible to use curl if wget was not found? (recursive may be a problem)

dralley:
> Would it be possible to use curl if wget was not found? (recursive may be a problem)

I would prefer not to - too much potential for weirdness and different behavior for unclear reasons.  We should just make `wget` available (and I believe it already is but if not it can be)

> Can we provide a check function (or decorator) that raises a skip exception (or can be used in the skip decorator)?

We could but I'm not sure it's necessarily worth it.  Do you think it would make sense for this function to just raise `unittests.SkipTest()` directly?

mdellweg:
> > Would it be possible to use curl if wget was not found? (recursive may be a problem)
> 
> I would prefer not to - too much potential for weirdness and different behavior for unclear reasons. We should just make `wget` available (and I believe it already is but if not it can be)

I kind of suspected that...
> 
> > Can we provide a check function (or decorator) that raises a skip exception (or can be used in the skip decorator)?
> 
> We could but I'm not sure it's necessarily worth it. Do you think it would make sense for this function to just raise `unittests.SkipTest()` directly?

Maybe yes. That sounds fine.

PR #1262, opened 2021-12-07
Makes pulp_smash a pytest plugin

bmbouter:
This provides a common place for any pytest fixtures to be added.

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -1,0 +1,236 @@
+import asyncio
+import threading
+import socket
+import ssl
+
+from dataclasses import dataclass
+from typing import Any
+
+from aiohttp import web
+
+import pytest
+import trustme
+
+from pulp_smash.api import _get_sleep_time
+from pulp_smash.config import get_config
+from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
+from pulp_smash.pulp3.fixture_utils import add_recording_route
+
+from pulpcore.client.pulpcore import ApiClient, TasksApi
+
+
+cfg = get_config()
+SLEEP_TIME = _get_sleep_time(cfg)
+
+
+class ThreadedAiohttpServer(threading.Thread):
+    def __init__(self, shutdown_event, app, host, port, ssl_ctx):
+        super().__init__()
+        self.shutdown_event = shutdown_event
+        self.app = app
+        self.host = host
+        self.port = port
+        self.ssl_ctx = ssl_ctx
+
+    def run(self):
+        loop = asyncio.new_event_loop()
+        runner = web.AppRunner(self.app)
+        asyncio.set_event_loop(loop)
+        loop.run_until_complete(runner.setup())
+        site = web.TCPSite(runner, host=self.host, port=self.port, ssl_context=self.ssl_ctx)
+        loop.run_until_complete(site.start())
+        while True:
+            loop.run_until_complete(asyncio.sleep(1))
+            if self.shutdown_event.is_set():
+                break
+
+
+@dataclass
+class ThreadedAiohttpServerData:
+    host: str
+    port: int
+    shutdown_event: Any
+    thread: Any
+    ssl_ctx: Any
+    requests_record: Any
+
+    def make_url(self, path):
+        if path[0] != "/":
+            raise ValueError("The `path` argument should start with a '/'")
+
+        if self.ssl_ctx is None:
+            protocol_handler = "http://"
+        else:
+            protocol_handler = "https://"
+
+        return f"{protocol_handler}{self.host}:{self.port}{path}"
+
+
+## Webserver Fixtures
+
+
+@pytest.fixture
+def unused_port():
+    def _unused_port():
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind(("127.0.0.1", 0))
+            return s.getsockname()[1]
+
+    return _unused_port
+
+
+@pytest.fixture
+def gen_threaded_aiohttp_server(unused_port):
+    fixture_servers_data = []
+
+    def _gen_threaded_aiohttp_server(app, ssl_ctx, call_record):
+        host = cfg.aiohttp_fixtures_origin
+        port = unused_port()
+        shutdown_event = threading.Event()
+        fixture_server = ThreadedAiohttpServer(shutdown_event, app, host, port, ssl_ctx)
+        fixture_server.daemon = True
+        fixture_server.start()
+        fixture_server_data = ThreadedAiohttpServerData(
+            host=host,
+            port=port,
+            shutdown_event=shutdown_event,
+            thread=fixture_server,
+            requests_record=call_record,
+            ssl_ctx=ssl_ctx,
+        )
+        fixture_servers_data.append(fixture_server_data)
+        return fixture_server_data
+
+    yield _gen_threaded_aiohttp_server
+
+    for fixture_server_data in fixture_servers_data:
+        fixture_server_data.shutdown_event.set()
+
+    for fixture_server_data in fixture_servers_data:
+        fixture_server_data.thread.join()
+
+
+@pytest.fixture
+def gen_fixture_server(gen_threaded_aiohttp_server):
+    def _gen_fixture_server(fixtures_root, ssl_ctx):
+        app = web.Application()
+        call_record = add_recording_route(app, fixtures_root)
+        return gen_threaded_aiohttp_server(app, ssl_ctx, call_record)
+
+    yield _gen_fixture_server
+
+
+@pytest.fixture(scope="session")
+def pulpcore_client():
+    configuration = cfg.get_bindings_config()
+    return ApiClient(configuration)
+
+
+@pytest.fixture(scope="session")
+def tasks_api_client(pulpcore_client):
+    return TasksApi(pulpcore_client)
+
+
+## Orphan Handling Fixtures
+
+
+@pytest.fixture
+def delete_orphans_pre():
+    delete_orphans()
+    yield
+
+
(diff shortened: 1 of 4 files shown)


pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    return ApiClient(configuration)
+
+
+@pytest.fixture
+def tasks_api_client(pulpcore_client):
+    return TasksApi(pulpcore_client)
+
+
+@pytest.fixture
+def async_monitor_task(tasks_api_client):
```
mdellweg, line 23: I still fail to see the value in this being a "fixture".

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+@pytest.fixture
+def pulpcore_client():
+    cfg = get_config()
+    configuration = cfg.get_bindings_config()
+    return ApiClient(configuration)
+
+
+@pytest.fixture
+def tasks_api_client(pulpcore_client):
+    return TasksApi(pulpcore_client)
```
mdellweg, line 19: I pretty much like this. Can we mark it as scope="session"?

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: I agree. I'm not finished with this one yet. The issue I'm struggling with is that if this becomes a normal function then I can't access the chain of fixtures it depends on tasks_api_client -> pulpcore client fixtures. I'm going to look at this more today.

mdellweg: Oh right. There is a point.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture
+def delete_orphans_post():
+    yield
+    delete_orphans()
+
+
+@pytest.fixture
+def delete_orphans_pre_post(delete_orphans_pre, delete_orphans_post):
+    yield
+    delete_orphans()
```
mdellweg, line 55: That is once too often now. ;)

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: I think this was a very good idea.

bmbouter:
Alright I think these achieve everything I'm happy with so I'm going to undraft it and leave a few comments for reviewers.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,209 @@
...
+            port=port,
+            shutdown_event=shutdown_event,
+            thread=fixture_server,
+            requests_record=call_record,
+        )
+
+    yield _gen_threaded_aiohttp_server
+
+    for shutdown_event in shutdown_events:
+        shutdown_event.set()
```
mdellweg, line 95: Can we join the thread here?

bmbouter: I pushed a revision that joins().

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+@pytest.fixture
+def delete_orphans_post():
+    yield
+    delete_orphans()
+
+
+@pytest.fixture
+def delete_orphans_pre_post(delete_orphans_pre, delete_orphans_post):
+    yield
```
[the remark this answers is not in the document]
mdellweg: sounds good. We could depend other fixtures on this one to get the order right.
(cleanup is in reverse dependency order.)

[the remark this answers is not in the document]
bmbouter: Agreed. I made this change and the runtime dropped by a few more seconds for the 7 tests I've been running. It also removed one unecessary fixture in the other PR too.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def run(self):
+        loop = asyncio.new_event_loop()
+        runner = web.AppRunner(self.app)
+        asyncio.set_event_loop(loop)
+        loop.run_until_complete(runner.setup())
+        site = web.TCPSite(runner, host=self.host, port=self.port, ssl_context=self.ssl_ctx)
+        loop.run_until_complete(site.start())
+        while True:
+            loop.run_until_complete(asyncio.sleep(1))
```
mdellweg, line 43: This is replacing the run_forever here?

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    configuration = cfg.get_bindings_config()
+    return ApiClient(configuration)
+
+
+@pytest.fixture(scope="session")
+def tasks_api_client(pulpcore_client):
+    return TasksApi(pulpcore_client)
+
+
+## Oprhan Handling Fixtures
```
mdellweg, line 134:
```suggestion
## Orphan Handling Fixtures
```

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+        cfg.aiohttp_fixtures_origin,
+    )
+
+
+## Client Side TLS Fixtures
+
+
+@pytest.fixture
+def client_tls_certificate_authority():
+    return trustme.CA()
```
mdellweg, line 168: Is this creating a new CA?
If yes, can we scope it to session?
If not, how is this different from tls_certificate_authority?

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    def _gen_object_with_cleanup(api_client, data):
+        new_obj = api_client.create(data)
+        obj_hrefs.append((api_client, new_obj.pulp_href))
+        return new_obj
+
+    yield _gen_object_with_cleanup
+
+    delete_task_hrefs = []
+    for api_client, pulp_href in obj_hrefs:
+        task_url = api_client.delete(pulp_href).task
```
mdellweg, line 232: I'm not sure we can rely on all models to only delete in tasks.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+        loop.run_until_complete(runner.setup())
+        site = web.TCPSite(runner, host=self.host, port=self.port, ssl_context=self.ssl_ctx)
+        loop.run_until_complete(site.start())
+        while True:
+            loop.run_until_complete(asyncio.sleep(1))
+            if self.shutdown_event.is_set():
+                break
+
+
+@dataclass
```
mdellweg, line 48: TIL

mdellweg (commented):
I don't think, we need the conftest.py in this repo. But probably does not hurt also.

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: Yes it is. It's checking for the shutdown event and otherwise goes back to serving fixtures data.

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: It is creating a CA with trustme. The tls certificate is the certificate produced with the CA. We can make this session scoped I think.

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: I agree, but we can use this for all of those types of models. I was thinking the name should reflect this better, but I didn't come up with anything.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    ssl_ctx.verify_mode = ssl.CERT_REQUIRED
+    tls_certificate.configure_cert(ssl_ctx)
+    return ssl_ctx
+
+
+## Object Cleanup fixtures
+
+
+@pytest.fixture
+def gen_object_with_cleanup():
```
bmbouter, line 220: Maybe `gen_objects_with_cleanup_async_delete` ?

mdellweg: I would do autodetection whether there is waiting on tasks needed.

pulp_smash/pulp3/pytest_plugin/__init__.py
mdellweg: Agreed to wait for this to become a problem.
(autodetect is still my preferred solution)

mdellweg (approved):

dkliban (approved):

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -0,0 +1,236 @@
+import asyncio
+import threading
+import socket
+import ssl
+
+from dataclasses import dataclass
```
dralley, line 6: This breaks compatibility with Python 3.6

bmbouter: Hmmm, maybe I need to rework this so it doesn't break old release branches....

bmbouter: This got addressed https://github.com/pulp/pulp-smash/commit/a04c4ee6626876c385010def5afb488910fce883

PR #1266, opened 2022-01-17
Adds pytest and pytest-xdist as a dependency

bmbouter:
These are not used directly in pulp-smash, but pytest-xdist is going to
be used by everyone who uses pulp-smash soon, so let's make it available
everywhere by adding it here.

[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,9 +45,10 @@ setup(
         "packaging",
         "plumbum",
         "pulpcore-client",
+        "pytest",
+        "pytest-xdist",
         "pyxdg",
         "requests",
-        "pulpcore-client",
         "trustme",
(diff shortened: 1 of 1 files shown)


fao89 (approved):

setup.py
```diff
@@ -45,9 +45,10 @@
         "packaging",
         "plumbum",
         "pulpcore-client",
+        "pytest",
+        "pytest-xdist",
         "pyxdg",
         "requests",
-        "pulpcore-client",
```
fao89, line 50: actually, I think it is needed

bmbouter: It was a duplicate, it's still at L#47

PR #1270, opened 2022-01-25
Adds rerooting auto-support

bmbouter:
This discovers the API_ROOT for the server under test and then builds
URLs in pulp-smash built off of that instead of the hardcoded
`/pulp/api/v3/`.

[noissue]


fao89 (approved):
LGTM

mdellweg (approved):

PR #1272, opened 2022-01-31
Adds task following support for pytest fixtures

bmbouter:
The `gen_object_with_cleanup` was not compatible with objects that
required the monitoring of a task for completion. This PR monitors the
task and then handles the created_resource.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -360,7 +360,23 @@ def gen_object_with_cleanup():
 
     def _gen_object_with_cleanup(api_client, data):
         new_obj = api_client.create(data)
-        obj_hrefs.append((api_client, new_obj.pulp_href))
+        try:
+            obj_hrefs.append((api_client, new_obj.pulp_href))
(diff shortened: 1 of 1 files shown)


pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -360,7 +360,28 @@ def gen_object_with_cleanup():
...
-        obj_hrefs.append((api_client, new_obj.pulp_href))
+        try:
+            obj_hrefs.append((api_client, new_obj.pulp_href))
+        except AttributeError:
+            # This is a task and the real object href comes from monitoring it
+            import pydevd_pycharm
+
+            pydevd_pycharm.settrace(
+                "localhost", port=29437, stdoutToServer=True, stderrToServer=True
+            )
```
fao89, line 371: was this supposed to be here?

bmbouter: Nope, and it need a bit more work too. I'm moving back to draft. Thanks for commenting on this.

dkliban (approved):

PR #1273, opened 2022-02-03
Adds nightly pytest marker

bmbouter:
[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -32,6 +32,10 @@ def pytest_configure(config):
         "markers",
         "serial: marks tests as required to run serially without any other tests also running",
     )
+    config.addinivalue_line(
+        "markers",
+        "nightly: marks tests as intended to run during the nightly CI run",
+    )
 
 
 class ThreadedAiohttpServer(threading.Thread):


fao89 (approved):
This LGTM, but I have a pytest question,

when running pytest not nightly, wouldn't it run parallel tests?
or not parallel would run nightly tests?

bmbouter:
> This LGTM, but I have a pytest question,
> 
> when running pytest not nightly, wouldn't it run parallel tests? or not parallel would run nightly tests?

I've been asking myself similar questions. Here's my current thinking at least. The nightly test run will run "all tests" like it does now. All non-nightly test runs will disinclude the nightly tagged tests by specifying `-m "not nightly"`. I'm putting this into a plugin_template PR right now actually, I'd like to know what you think is best.

bmbouter:
@fao89 wdyt about https://github.com/pulp/plugin_template/pull/567

fao89:
> > This LGTM, but I have a pytest question,
> > when running pytest not nightly, wouldn't it run parallel tests? or not parallel would run nightly tests?
> 
> I've been asking myself similar questions. Here's my current thinking at least. The nightly test run will run "all tests" like it does now. All non-nightly test runs will disinclude the nightly tagged tests by specifying `-m "not nightly"`. I'm putting this into a plugin_template PR right now actually, I'd like to know what you think is best.

I don't have an opinion on that, I'm just afraid of collision of pytest markers, I believe `-m "not nightly"`will trigger parallel tests, but I'm not sure if this will happen

PR #1275, opened 2022-02-10
Use a proxy.py pre-release with a bugfix we need

bmbouter:
We should test this and use proxy.py==2.4.1 when it's released.

[noissue]


ggainey (approved):
lg2m - let's see how it goes!

PR #1276, opened 2022-02-11
Add fixtures to start and stop pulp services

goosemania:
re https://github.com/pulp/pulpcore/pull/1434

[noissue]


pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -252,6 +266,57 @@ def tasks_api_client(pulpcore_client):
...
+def status_api_client(pulpcore_client):
+    return StatusApi(pulpcore_client)
+
+
+@pytest.fixture
+def stop_and_check_services(status_api_client):
+    """Stop services and wait up to 10 seconds to check if services have stopped."""
+
+    def _stop_and_check_services(pulp_services=None):
+        svc_mgr.stop(pulp_services or PULP_SERVICES)
```
mdellweg, line 279: I'm wondering, if the waiting should be done in the `smash_cli.ServiceManager`.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -13,14 +15,26 @@
...
 
 cfg = get_config()
+cli = smash_cli.Client(cfg)
 SLEEP_TIME = _get_sleep_time(cfg)
+PULP_HOST = cfg.hosts[0]
+PULP_SERVICES = (
+    "pulpcore-content",
+    "pulpcore-api",
+    "new-pulpcore-worker@1",
+    "new-pulpcore-worker@2",
```
bmbouter, line 34: I think we need to also have the regular worker names too in cases where the system under test is not the single container.

pulp_smash/pulp3/pytest_plugin/__init__.py
dralley: I agree, it should probably be implemented as `def stop(services, blocking=True)`

goosemania: I'm not sure about that. 
`smash_cli.ServiceManager` is generic enough to stop and start any service, e.g. posgresql, and not necessarily pulp.
Here all the logic for checking and waiting is specifically for pulp.
I think either we keep it as is in fixtures, or we need to provide some callback to service manager for waiting and checking according to the services it is working with. I'd vote for keeping as is.

Thougths? @mdellweg @dralley @dkliban

dkliban: I would leave it as is.

mdellweg: That's a good point. Keep it this way then.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -285,6 +294,57 @@ def task_schedules_api_client(pulpcore_client):
...
+@pytest.fixture
+def status_api_client(pulpcore_client):
+    return StatusApi(pulpcore_client)
+
+
+@pytest.fixture
+def stop_and_check_services(status_api_client):
+    """Stop services and wait up to 10 seconds to check if services have stopped."""
+
+    def _stop_and_check_services(pulp_services=None):
```
mdellweg, line 306: How are the pulp_services determined, if i have a dev-box with 8 workers? (At least with systemd we have a pulpcore service that will (re-)start all pulp processes.)
Maybe we would need such a thing in the container setup first. So this should not hold the PR back.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -14,14 +16,21 @@
...
 from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
 from pulp_smash.pulp3.fixture_utils import add_recording_route
 
-from pulpcore.client.pulpcore import ApiClient, TasksApi
+from pulpcore.client.pulpcore import ApiClient, StatusApi, TasksApi
+from pulpcore.client.pulpcore.exceptions import ApiException
 
 
 cfg = get_config()
+cli = smash_cli.Client(cfg)
```
mdellweg, line 28: Maybe blind, but i don't see this used. Also it collides with the import on line 16.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -14,14 +16,21 @@
...
 SLEEP_TIME = _get_sleep_time(cfg)
+PULP_HOST = cfg.hosts[0]
+PULP_SERVICES = ("pulpcore-content", "pulpcore-api", "pulpcore-worker@1", "pulpcore-worker@2")
+
+svc_mgr = smash_cli.ServiceManager(cfg, PULP_HOST)
```
mdellweg, line 33: This could be a fixture in the same way as `cli_client`.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -285,6 +294,57 @@ def task_schedules_api_client(pulpcore_client):
         return TaskSchedulesApi(pulpcore_client)
 
 
+@pytest.fixture
+def status_api_client(pulpcore_client):
+    return StatusApi(pulpcore_client)
+
+
+@pytest.fixture
+def stop_and_check_services(status_api_client):
```
mdellweg, line 303: I would add these in the "section" called "infrastructure fixtures".

mdellweg (changes requested):
The `cli` name clash is the only required change between some suggestions.

pulp_smash/pulp3/pytest_plugin/__init__.py
goosemania: copy-paste from the previous place of residence, I'll remove it

pulp_smash/pulp3/pytest_plugin/__init__.py
goosemania: Currently they are hard-coded. I agree there is a room for improvement, outside of this PR.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -257,6 +262,57 @@ def cli_client(pulp_cfg):
     return cli.Client(pulp_cfg)
 
 
+@pytest.fixture(scope="session")
+def svc_mgr(pulp_cfg):
+    return cli.ServiceManager(pulp_cfg, PULP_HOST)
+
+
+@pytest.fixture
```
dralley, line 270: I don't want to block this PR, but I'm not sure I understand the benefit of having them be pytest fixtures?  What does the fixture decoration provide us here

dralley (approved):

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -285,6 +341,11 @@ def task_schedules_api_client(pulpcore_client):
         return TaskSchedulesApi(pulpcore_client)
 
 
+@pytest.fixture
```
mdellweg, line 344:
```suggestion
@pytest.fixture(scope="session")
```

pulp_smash/pulp3/pytest_plugin/__init__.py
dralley: +1

fao89:
as the pulpcore PR got merged, I think we should merge this PR

pulp_smash/pulp3/pytest_plugin/__init__.py
goosemania: I can address it in a follow-up PR if needed.

fao89: @goosemania you are free now, I can do a follow-up PR if needed

mdellweg: Yes, and no. Yes, you only create (and tear down) the fixture once per session. But what is "the fixture"? In this case it is an instance of `StatusApi` preloaded with the `pulpcore_client` containing hostname, username and password. This object then has methods that can be called by tests and they will perform http requests on the spot.
Assuming these objects behave nicely in threads with our parallel testing strategy, scoping them by the session is the way for ultimate reusing.

mdellweg: In the case of the service restart, the fixture to be cached is the function that will restart the services. And it can be run multiple times. Restarting services however is not meant for parallel tests.

PR #1283, opened 2022-03-16
Add add_to_cleanup fixture

mdellweg:
This fixture will collect references to pulp objects to be deleted after
a test. It will attempt deletion in reversed order and await all tasks
in parallel.

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -422,7 +422,9 @@ else:
 
 
 @pytest.fixture
-def delete_orphans_pre():
+def delete_orphans_pre(request):
+    if request.node.get_closest_marker("parallel") is not None:
+        raise pytest.UsageError("This test is not suitable to be marked parallel.")
     delete_orphans()
     yield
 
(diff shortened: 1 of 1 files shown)


mdellweg:
Rationale: In pulp container we create Distributions and namespaces by push operations, so we cannot use `gen_object_with_cleanup` there. Also we do not have `unittest.TestCase.teardown()` available anymore.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -403,7 +403,9 @@ def __exit__(self, exc_type, exc_value, traceback):
 
 
 @pytest.fixture
-def delete_orphans_pre():
+def delete_orphans_pre(request):
+    if request.node.get_closest_marker("parallel") is not None:
+        raise pytest.UsageError("This test is no suitable to be marked parallel.")
```
bmbouter, line 408: Nice!

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -506,13 +508,35 @@ def ssl_ctx_req_client_auth(
...
 
 @pytest.fixture
-def gen_object_with_cleanup():
-    obj_hrefs = []
+def add_to_cleanup():
+    """Fixture to allow pulp objects to be deleted after the test."""
+    obj_refs = []
 
+    def _add_to_cleanup(api_client, pulp_href):
+        obj_refs.insert(0, (api_client, pulp_href))
```
bmbouter, line 516: Python lists are not optimized to insert at the beginning, but the collections.deque is optimized to insert from both ends. I think we should use that.

bmbouter (changes requested):
I'm requesting one data structure change please.

pulp_smash/pulp3/pytest_plugin/__init__.py
mdellweg: What about calling `obj_refs.reverse()` before cleanup? Or using a reverse iterator (does python have such a thing?)?

pulp_smash/pulp3/pytest_plugin/__init__.py
mdellweg:
```suggestion
        raise pytest.UsageError("This test is not suitable to be marked parallel.")
```

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: Then we have to reverse them? I think this is what `collections.deque` was made for actually so I think we should use it.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -525,13 +527,35 @@ def ssl_ctx_req_client_auth(
...
+    """Fixture to allow pulp objects to be deleted after the test."""
+    obj_refs = []
 
+    def _add_to_cleanup(api_client, pulp_href):
+        obj_refs.append((api_client, pulp_href))
+
+    yield _add_to_cleanup
+
+    delete_task_hrefs = []
+    for api_client, pulp_href in reversed(obj_refs):
```
bmbouter, line 540: Can you put a little commend on why processing in reverse order is important?

bmbouter (approved):
Thanks!

PR #1285, opened 2022-04-11
Adds the --pulp-no-leftover-objs pytest arg

bmbouter:
[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -19,8 +19,17 @@ from pulp_smash.config import get_config
 from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
 from pulp_smash.pulp3.fixture_utils import add_recording_route
 
-from pulpcore.client.pulpcore import ApiClient, StatusApi, TasksApi
+from pulpcore.client.pulpcore import (
+    ApiClient,
+    ContentguardsApi,
+    DistributionsApi,
+    PublicationsApi,
+    RepositoriesApi,
+    StatusApi,
+    TasksApi,
+)
 from pulpcore.client.pulpcore.exceptions import ApiException
+from pulpcore.tests.functional.api.using_plugin.utils import gen_pulpcore_client
 
 
 cfg = get_config()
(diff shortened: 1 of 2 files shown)


bmbouter:
With this a plugin could define something like this pulp_file example in their conftest.py

```
def pytest_check_for_leftover_pulp_objects(config):
    file_client = gen_file_client()

    types_to_check = [
        RepositoriesFileApi(file_client),
        # ContentFilesApi(file_client),  # It's ok to leave behind orphans
        RemotesFileApi(file_client),
        PublicationsFileApi(file_client),
        DistributionsFileApi(file_client),
    ]
    for type_to_check in types_to_check:
        if type_to_check.list().count > 0:
            raise Exception(f"This test left over a {type_to_check}.")
```

bmbouter:
So this is neat and all, but I realized that pulpcore could check the generic list endpoints of the following object types which would not require plugins to add any code! I'm going to revise this to do that instead. At least I learned a lot about how to write a pytest plugin.

pulp_smash/pulp3/pytest_plugin/newhooks.py
```diff
@@ -0,0 +1,4 @@
+def pytest_check_for_leftover_pulp_objects(config):
```
mdellweg, line 1: Is this filename a common one, or taken from an example? Should we name it `smashhooks` or `pulphooks`?

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -29,7 +30,52 @@
...
+    if item.config.getoption("--pulp-no-leftover-objs"):
+        item.config.hook.pytest_check_for_leftover_pulp_objects(config=item.config)
+
+
+## pytest configuration
+
+
 def pytest_configure(config):
+    if config.pluginmanager.hasplugin("xdist") and config.getoption("-n"):
+        raise Exception("The --pulp-no-leftover-objs cannot be used with -n from xdist")
```
mdellweg, line 77: This condition seems incomplete to me.

bmbouter:
@mdellweg what name did you want for the argument? I forgot the suggestion from the video call.

bmbouter:
[Here's an example](https://github.com/pulp/pulp_file/pull/681) of the hook being used by pulp_file.

bmbouter:
@mdellweg this is ready for re-review and I'll rename the arg when you tell me what sounds best to you.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -29,7 +38,67 @@
...
+## Pulp No Leftover Objs Feature
+
+
+def pytest_addoption(parser):
+    group = parser.getgroup("pulp-smash")
+    group.addoption(
+        "--pulp-no-leftover-objs",
+        action="store_true",
+        dest="pulp_no_leftover_objs",
+        help="Enable this to have Pulp plugins check for objects leftover by tests.",
```
mdellweg, line 50:
```suggestion
        "--pulp-no-leftovers",
        action="store_true",
        dest="pulp_no_leftovers",
        help="Enable this to have Pulp plugins check for objects leftover by tests.",
```

I thought calling them objects is redundant. And saving few keystrokes will hopefully help people to run it more often.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -29,7 +38,67 @@
...
+
+    types_to_check = [
+        ContentguardsApi(pulpcore_client),
+        DistributionsApi(pulpcore_client),
+        PublicationsApi(pulpcore_client),
+        RepositoriesApi(pulpcore_client),
+    ]
+    for type_to_check in types_to_check:
+        if type_to_check.list().count > 0:
+            raise Exception(f"This test left over a {type_to_check}.")
```
mdellweg, line 71: is `repr(type_to_check)` readable? Also maybe call it `model_to_check`.

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: I changed this in my next push.

pulp_smash/pulp3/pytest_plugin/__init__.py
bmbouter: I switched this to model_to_check in my next push.

bmbouter: I think it's pretty readable it produces a statement like: `This test left over a <class 'pulpcore.client.pulpcore.api.distributions_api.DistributionsApi'>.`

mdellweg (approved):

PR #1287, opened 2022-04-12
Do not run podman with sudo

### LINKED ISSUE #1286: Registry client commands should not be executed with sudo
At the moment, all registry client commands are executed with `sudo` permissions:
https://github.com/pulp/pulp-smash/blob/7f8d1de5fb5fba221f3de98e001bc4990009eb42/pulp_smash/cli.py#L1028

This should not be needed for `podman`. But, some commands may try to access files that require `sudo` permissions. As a result, we would probably need to specify, e.g., `--authfile` parameter in some cases, like so:
```
podman login -u admin -p password --authfile=$HOME/containers/auth.json pulp3-source-fedora35.localhost.example.com
```
```
podman pull pulp3-source-fedora35.localhost.example.com/test:asfd --authfile=$HOME/containers/auth.json 
```

lubosmj:
closes #1286

### DIFF
--- a/pulp_smash/cli.py
+++ b/pulp_smash/cli.py
@@ -311,7 +311,9 @@ class Client:  # pylint:disable=too-few-public-methods
         elif self.transport == "podman":
             args = ("podman", "exec", "-i", self._podname) + tuple(args)
 
-        if sudo and args[0] != "sudo" and not self.is_superuser:
+        # docker still requires some commands to be executed as sudo; running docker without
+        # sudo in a rootless mode is supported as of 19.03.14
+        if sudo and args[0] != "sudo" and not self.is_superuser or self.transport == "docker":
             args = ("sudo",) + tuple(args)
 
         code, stdout, stderr = self.machine[args[0]].run(args[1:], **kwargs)
@@ -1025,7 +1027,7 @@ class RegistryClient:
             args[0] = urlunsplit(urlsplit(args[0])._replace(scheme="")).strip("//")
 
         cmd = (self.name, command) + tuple(args)
-        result = self._client.run(cmd, sudo=True)
+        result = self._client.run(cmd)
         try:
             # most of client responses are JSONable
             return json.loads(result.stdout)


lubosmj:
This PR removes the need for executing commands with `sudo`. As demonstrated in https://github.com/pulp/pulp_container/pull/705, all functional tests can be run without using `sudo` along with `podman`.

pulp_smash/cli.py
```diff
@@ -311,7 +311,9 @@ def run(self, args, sudo=False, **kwargs):
         elif self.transport == "podman":
             args = ("podman", "exec", "-i", self._podname) + tuple(args)
 
-        if sudo and args[0] != "sudo" and not self.is_superuser:
+        # docker still requires some commands to be executed as sudo; running docker without
+        # sudo in a rootless mode is supported as of 19.03.14
+        if sudo and args[0] != "sudo" and not self.is_superuser or self.transport == "docker":
```
mdellweg, line 316: Is image pulling and pushing really supposed to need root?
We are not attempting to run any containers. I would rather assume we don't need this.

lubosmj: I have never been able to pull or push content without `sudo` by default. This is the error that I usually see:
```
(pulp) [vagrant@pulp3-source-fedora35 ~]$ docker pull busybox
Using default tag: latest
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/images/create?fromImage=busybox&tag=latest": dial unix /var/run/docker.sock: connect: permission denied
```

It is possible to manage the `docker` client as a non-user. But, with some additional configuration: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user. Is it worth it?

mdellweg: Probably not.

mdellweg (approved):

pulp_smash/cli.py
ipanova: It is not possible to sign and push signatures with docker, hence we don't need to solve  the questions related to gpp keyring,home etc. We can keep running docker with sudo. I would not separate `sudo in a rootless mode is supported as of 19.03.14` since this information is not confirmed, as you've pointed out.
You just need to update your signing tests that will need to skipped for  cases when they are run from under docker.
https://github.com/pulp/pulp_container/pull/633/files#diff-d43c113124ff60410fbd72c9e5c1a7dbdda49ea39f02d8626286527242db013dR53

PR #1291, opened 2022-04-19
Removes pulpcore fixtures

bmbouter:
The pulpcore fixtures are moving to the pulpcore repo so they can be
versioned with pulpcore.

Required PR: https://github.com/pulp/pulpcore/pull/2562

[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -16,7 +16,7 @@ from yarl import URL
 from pulp_smash import cli
 from pulp_smash.api import _get_sleep_time
 from pulp_smash.config import get_config
-from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
+from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.fixture_utils import add_recording_route
 
 from pulpcore.client import pulpcore as pulpcore_bindings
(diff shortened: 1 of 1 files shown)


mdellweg:
I don't think, we can remove them any time soon (the old versioning problem). At least we need to make sure all plugins known to us do not use those fixtures anymore on any branch that is still supposed to be released.
I still think we should strive for the deletion, but there may be a lot of places we need to look first.
The good news is, that nothing stops us from adding the fixtures to pulpcore, as pytest will just use the ones it finds first.

bmbouter:
> I don't think, we can remove them any time soon (the old versioning problem). At least we need to make sure all plugins known to us do not use those fixtures anymore on any branch that is still supposed to be released. I still think we should strive for the deletion, but there may be a lot of places we need to look first. The good news is, that nothing stops us from adding the fixtures to pulpcore, as pytest will just use the ones it finds first.

I agree with the problem statement, but I think we should go ahead and handle that because we'll backport the changes on whatever plugins + branches need it to also switch to using pulpcore. In practice I don't think it would be more than 10 PRs which seems very doable to finally get this resolved. Would you be ok with that? I've marked this as draft because I don't want to do anything without a shared plan.

mdellweg:
How about, we add the fixtures to pulpcore and start using them at least on pulp_file. Then we delete them here and see what needs fixing with a working example in the quiver.

bmbouter:
@mdellweg can you review this even at draft with the understanding I won't undraft and merge until we have pulpcore:main and pulp_file:main both using the new fixtures with https://github.com/pulp/pulpcore/pull/2562 and https://github.com/pulp/pulp_file/pull/687

bmbouter:
After this is merged I will work to go fix any breaking backport branches on any plugins asap.

mdellweg (approved):

bmbouter:
Here goes nothing!

PR #1293, opened 2022-05-10
Add --nightly flag to commandline

mdellweg:
[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -28,16 +28,20 @@ PULP_HOST = cfg.hosts[0]
 PULP_SERVICES = ("pulpcore-content", "pulpcore-api", "pulpcore-worker@1", "pulpcore-worker@2")
 
 
-## Pulp No Leftover Objs Feature
-
-
 def pytest_addoption(parser):
     group = parser.getgroup("pulp-smash")
     group.addoption(
         "--pulp-no-leftovers",
         action="store_true",
         dest="pulp_no_leftovers",
+        default=False,
         help="Enable this to have Pulp plugins check for objects leftover by tests.",
+    )
+    group.addoption(
+        "--nightly",
+        action="store_true",
+        default=False,
+        help="Enable to run nightly test.",
     )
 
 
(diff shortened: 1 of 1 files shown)


pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -28,17 +28,21 @@
...
     group = parser.getgroup("pulp-smash")
     group.addoption(
         "--pulp-no-leftovers",
         action="store_true",
         dest="pulp_no_leftovers",
+        default=False,
         help="Enable this to have Pulp plugins check for objects leftover by tests.",
     )
+    group.addoption(
+        "--nightly",
```
bmbouter, line 41: I guess I'm unclear on what the benefit is from this PR. Running all the tests is what you don't specify filtering through markers.

mdellweg: "nightly" is our way of saying "this test will run slow and use a lot of resources". So the benefit is that a contributor running pytest in his dev env will run all the usual tests and get a result in a reasonable amount of time. They can still run all tests by specifying `--nightly`. This is literally the example given in the pytest docs, just they call it "slow".

mdellweg: To me it is rather important how to run tests as a dev. Making running the tests locally a bad experience, i think slows down development at the cost of running tests more often in the CI. And the increased load in the CI has all sorts of bad impact on the project. What i want to say is "the default set of tests" to me is what is exercised by simply calling `pytest` without parameters.

pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -57,6 +61,18 @@ def pytest_runtest_teardown(item):
         item.config.hook.pytest_check_for_leftover_pulp_objects(config=item.config)
 
 
+def pytest_collection_modifyitems(config, items):
+    # Skip nightly tests by default
+    # https://docs.pytest.org/en/7.1.x/example/simple.html#control-skipping-of-tests-according-to-command-line-option
+    if config.getoption("--nightly"):
```
bmbouter, line 67:
```suggestion
    if not config.getoption("--nightly"):
```

Consider having the `not` here make this branch do things instead of the `return` without the `not`.

mdellweg: I took the example from here: https://docs.pytest.org/en/6.2.x/example/simple.html#control-skipping-of-tests-according-to-command-line-option

bmbouter:
I hand tested this locally and it worked nicely!

```
devel/pulp_file/pulp_file/tests/functional/api/test_sync.py::test_sync_file_protocol_handler PASSED                                                                                                                                                                                                                    [ 20%]
devel/pulp_file/pulp_file/tests/functional/api/test_sync.py::test_mirrored_sync SKIPPED (need --nightly option to run)                                                                                                                                                                                                 [ 40%]
devel/pulp_file/pulp_file/tests/functional/api/test_sync.py::test_invalid_url PASSED                                                                                                                                                                                                                                   [ 60%]
devel/pulp_file/pulp_file/tests/functional/api/test_sync.py::test_invalid_file PASSED                                                                                                                                                                                                                                  [ 80%]
devel/pulp_file/pulp_file/tests/functional/api/test_sync.py::test_duplicate_file_sync PASSED                                                                                                                                                                                                                           [100%]
```

bmbouter (approved):
This is great!

PR #1296, opened 2022-05-12
Update version to 1!0.15.0

ipanova:
[noissue]

We need to update the version to 1!0.15.0
On pypi we have versions up to 1!0.14.0 including, so we whether update version here or need to withdraw some pypi releases.


pedro-psb:
Reason: we don't maintain/use pulp-smash anymore.

PR #1297, opened 2022-05-12
Fixes how pulpcore-manager command is discovered

dkliban:
[noissue]

### DIFF
--- a/pulp_smash/utils.py
+++ b/pulp_smash/utils.py
@@ -158,9 +158,8 @@ def execute_pulpcore_python(cli_client, command):
         ps_output = cli_client.run(("ps", "ax")).stdout.splitlines()
         bin_dir = "/usr/local/bin"
         for line in ps_output:
-            if "pulpcore" in line:
-                if "bin/python" in line.split()[4]:
-                    bin_dir = line.split()[4].rsplit("/", maxsplit=1)[0]
+            if "pulpcore" in line and "bin/python" in line:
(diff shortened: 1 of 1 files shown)


bmbouter (approved):
Ha excellent

pulp_smash/utils.py
```diff
@@ -159,8 +159,8 @@ def execute_pulpcore_python(cli_client, command):
         bin_dir = "/usr/local/bin"
         for line in ps_output:
             if "pulpcore" in line:
-                if "bin/python" in line.split()[4]:
-                    bin_dir = line.split()[4].rsplit("/", maxsplit=1)[0]
+                if "bin/python" in line.split()[5]:
+                    bin_dir = line.split()[5].rsplit("/", maxsplit=1)[0]
```
mdellweg, line 163: Assuming this has worked for someone before, the format of ps must have chaged. Can we use something more stable? Maybe `ps ax -o command`?

mdellweg (changes requested):

pulp_smash/utils.py
dkliban: This command was never run inside the container before. It was only run inside our developer VMs. This change continues to work inside a VM and also inside the centos based container.

dkliban:
@mdellweg I am merging this as it is now so I can stop requiring this PR everwhere. Let me know if you want me to change it though.

PR #1299, opened 2022-06-03
`gen_object_with_cleanup` needs kwargs support

bmbouter:
[noissue]


dkliban (approved):

bmbouter:
After more investigation we learned that the bindings need to stay that way, so if you want to use `gen_object_with_cleanup(file_content_api_client, ... )` you would need this.

PR #1300, opened 2022-06-08
Add support for multiple created_resources

bmbouter:
When creating objects that need cleanup later, you may end up with
multiple created_resource objects in the task. This adds support for
that situation, and it will find the right type instead of raising an
Exception.

[noissue]

### DIFF
--- a/pulp_smash/pulp3/pytest_plugin/__init__.py
+++ b/pulp_smash/pulp3/pytest_plugin/__init__.py
@@ -500,16 +500,17 @@ def gen_object_with_cleanup(add_to_cleanup):
             # This is a task and the real object href comes from monitoring it
             task_data = monitor_task(new_obj.task)
 
-            num_created_resources = len(task_data.created_resources)
-            if num_created_resources != 1:
-                msg = (
-                    f"gen_object_with_cleanup can only handle tasks with exactly 1 "
-                    f"created_resource. this task has f{num_created_resources}"
-                )
-                raise NotImplementedError(msg)
+            for created_resource in task_data.created_resources:
+                try:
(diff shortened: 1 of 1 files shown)


pulp_smash/pulp3/pytest_plugin/__init__.py
```diff
@@ -500,16 +500,17 @@ def _gen_object_with_cleanup(api_client, *args, **kwargs):
...
-            new_obj = api_client.read(task_data.created_resources[0])
-            add_to_cleanup(api_client, new_obj.pulp_href)
+            for created_resource in task_data.created_resources:
+                try:
+                    new_obj = api_client.read(created_resource)
+                except Exception:
+                    pass  # This isn't the right created_resource for this api_client
+                else:
+                    add_to_cleanup(api_client, new_obj.pulp_href)
+                    return new_obj
```
dkliban, line 510: This will return after processing one created resource. Is that intentional?

bmbouter: It is, but there may be cases where we need more later? The issue with returning N is that the interface right now is only setup to return one object, not N, and it's already in broad use.

dkliban: In what use case do you expect it to return 2 resources? Uploading a file to a repository? new content and new repo version?

bmbouter: Yeah it's exactly that case, when creating a new file content unit and also specifying a repo at creation time it also creates the new repo version. That's why I need this actually.

dkliban (approved):

PR #1303, opened 2022-07-13
Replace HTTPFound with HTTPNotFound

dkliban:
[noissue]


bmbouter (approved):
Thanks @dkliban Thanks @mdellweg !

PR #1306, opened 2022-10-04
Add more logic for gen_object_with_cleanup

gerrod3:
[noissue]


mdellweg:
I don't see what this is accomplishing. Also i would not wanto to make any more changes to pulp-smash.
Can we get this PR finished first? https://github.com/pulp/pulpcore/pull/3386

mdellweg:
Is this telling us that when a remote is referenced by a repository, and we delete the repository first we need to wait for the task before we can delete the remote?
In that case, i think we need to add the remote as a shared_resource to the repository delete task instead.

gerrod3:
> I don't see what this is accomplishing. Also i would not wanto to make any more changes to pulp-smash. Can we get this PR finished first? [pulp/pulpcore#3386](https://github.com/pulp/pulpcore/pull/3386)

This is specifically for using domains in `gen_object_with_cleanup`. Domains can not be deleted without all objects within being deleted first. `add_object_for_cleanup` tries to delete all the objects in parallel (dispatched in reverse order), but this means that the domain delete can be started before the other object's delete tasks are finished causing an error on cleanup.

gerrod3:
Closing as I changed domain delete to be async so this is no longer needed.

PR #1307, opened 2023-02-14
Replace the function scope with the class scope

lubosmj:
[noissue]


mdellweg (approved):

lubosmj:
In pulpcore, we have used the class scope for the fixtures in question. The PR is here to prevent discrepancies between registered fixtures.

PR #1308, opened 2023-04-13
Set the tls-verify option based on the scheme for podman

lubosmj:
[noissue]

This is required for tests to run in oci_env out of the box because we do not use the https scheme by default.


pulp_smash/cli.py
```diff
@@ -965,6 +965,18 @@ def __init__(self, cfg, raise_if_unsupported=None, pulp_host=None):
...
+        else:
+            tls_verify = "--tls-verify=true"
+
+        self.pull = lambda *args: self._dispatch_command("pull", *args, tls_verify)
+        self.push = lambda *args: self._dispatch_command("push", *args, tls_verify)
+        self.manifest_push = lambda *args: self._dispatch_command(
+            "manifest", "push", *args, tls_verify
+        )
+        self.login = lambda *args: self._dispatch_command("login", *args, tls_verify)
+
```
ggainey, line 979: What's the reason behind doing this for only a subset of the commands?

ggainey (approved):

